PR #2, opened 2017-10-07
Convolutions and pooling on CPU

dfdx:
I continue to decompose Knet.jl into reusable pieces that may be utilized by other projects. This PR is almost literal copy of the corresponding files in Knet minus dependencies on `KnetArray` plus a couple of renamings:

 * `conv4(w, x)` -> `conv2d(x, w)`, since it looks like a more standard name and argument order
 * `conv4x`, etc. -> `conv2d_grad_x`, etc, just to make naming more obvious
 * `libknet8` -> `libnnlib`, because it's almost 4am at my location and I can't come up with a better name :)

I also tried to replace C++ implementations of `im2col` and `col2im` with pure Julia, but the indexing differences are just crazy. Maybe later, maybe never. 

This PR doesn't include tests since the intended way is to cross-check it with CUDNN.jl implementation (e.g. see JuliaGPU/CUDNN.jl#8). 

cc: @denizyuret


dfdx:
Ready for review!

MikeInnes:
This is great! I'm a wary of the C++ dependency though; we can't really expect all users (e.g. on Windows) to have `make` or a compiler.

Would it be possible to use the implementations in [ImageFiltering.jl](https://github.com/JuliaImages/ImageFiltering.jl), even if we have to wrap them a bit for the API?

dfdx:
ImageFiltering.jl is quite different and lacks a number of features such as gradient for convolutions and the entire max pooling. Also it seems like the leading approach for convolutions in CNN is to use im2col trick (e.g. see "Implementation as Matrix Multiplication" section of Andrej Karpathy's [course](http://cs231n.github.io/convolutional-networks/)), while ImageFiltering.jl seems to rely more on FFT and other techniques. 

The most straightforward way to make it platform independent is to re-implement `im2col` and pooling in pure Julia. But it's time consuming: last time I spent ~1.5h to figure out index matching between C++ and Julia in `im2col` with no success. Now I simply don't have more time to invest into it. 

Alternatively, we could improve a build script to work for Windows as well. I believe something like MinGW should do the trick, but I don't have a Windows machine to test it.

denizyuret:
I plan to find a way to compile libknet on windows as well.  This may mean
moving away from Make and doing compiler calls from Julia, which would also
work for the conv problem.  I am open to other suggestions though.

SimonDanisch:
Btw, you can quite easily get gcc on windows with WinRPM: https://github.com/Voxel8/Clipper.jl/blob/master/deps/build.jl

Also, could we get some more julia implementations out of Mocha? Seems like there is at least an im2col implementation: https://github.com/pluskid/Mocha.jl/blob/master/benchmarks/native-im2col/im2col-bm.jl

denizyuret:
I have ported Mocha's im2col to Knet:
https://github.com/denizyuret/Knet.jl/blob/master/src/conv.cpp

deniz

dfdx:
@denizyuret yeah, that's exactly the implementation I copied to this PR. The problem is that it's in C++ and thus introduces an additional dependency which isn't easy to get on Windows.

dfdx:
> Also, could we get some more julia implementations out of Mocha? Seems like there is at least an im2col implementation: https://github.com/pluskid/Mocha.jl/blob/master/benchmarks/native-im2col/im2col-bm.jl

Pooling is also [there](https://github.com/pluskid/Mocha.jl/blob/master/benchmarks/native-pooling/pooling-bm.jl), but according to their benchmarks, both - convolutions and pooling are 1.8-9x slower in Julia than in C++. Need to re-test it, but if my network runs 4h instead of 2h just because of Julia-only implementation, I don't really buy it. 

WinRPM looks good. Is it stable enough to include as a dependency?

SimonDanisch:
should be! the servers have problems every couple of months, but it's usually resolved after ~one day

dfdx:
Just checked Mocha benchmarks on Lenovo T470s. Results for julia vs. C++ convolutions are nearly the same: 

```
julia> r1 = @benchmark im2col_jl()
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     254.258 μs (0.00% GC)
  median time:      256.514 μs (0.00% GC)
  mean time:        262.358 μs (0.00% GC)
  maximum time:     829.546 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> r2 = @benchmark im2col_c()
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     251.192 μs (0.00% GC)
  median time:      255.233 μs (0.00% GC)
  mean time:        264.944 μs (0.00% GC)
  maximum time:     555.211 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
However, results for pooling are very different: 

```
julia> r1 = @benchmark pooling_julia()
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     83.799 ms (0.00% GC)
  median time:      84.300 ms (0.00% GC)
  mean time:        85.386 ms (0.00% GC)
  maximum time:     91.201 ms (0.00% GC)
  --------------
  samples:          59
  evals/sample:     1

julia> r2 = @benchmark pooling_native()
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.232 ms (0.00% GC)
  median time:      1.233 ms (0.00% GC)
  mean time:        1.264 ms (0.00% GC)
  maximum time:     1.879 ms (0.00% GC)
  --------------
  samples:          3951
  evals/sample:     1
```
Also, benchmarks in Mocha miss convolution and pooling gradient implementations.

SimonDanisch:
Well, there are a couple of things in the C version which should improve performance and should be easy to port - e.g. using `@threads` - I would be surprised if we can't get it to the same speed!

dfdx:
Putting `Threads.@threads` before the first loop as `#pragma omp parallel for` in C++ version makes results even worse:

```
julia> r1 = @benchmark pooling_julia()
BenchmarkTools.Trial: 
  memory estimate:  144 bytes
  allocs estimate:  1
  --------------
  minimum time:     107.716 ms (0.00% GC)
  median time:      108.959 ms (0.00% GC)
  mean time:        109.040 ms (0.00% GC)
  maximum time:     111.117 ms (0.00% GC)
  --------------
  samples:          46
  evals/sample:     1
```
------
For some reason I can't reproduce initial result for C++ version too, getting this instead: 

```
r2 = @benchmark pooling_native()
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     19.257 ms (0.00% GC)
  median time:      21.428 ms (0.00% GC)
  mean time:        27.128 ms (0.00% GC)
  maximum time:     54.887 ms (0.00% GC)
  --------------
  samples:          185
  evals/sample:     1
```
Which is still 4x faster than the best for Julia.

MikeInnes:
On my machine, Tim's implementations are a solid 3x faster than Knet's on a reasonably sized image.

dfdx:
> On my machine, Tim's implementations are a solid 3x faster than Knet's on a reasonably sized image.

Can you share the code?

SimonDanisch:
For me, @threads reduces the runtime from 75ms to 18ms...
```Julia
function pooling(input::Array{T}, output::Array, mask::Array, kernel, pad, stride) where T
  width, height, channels, num = size(input)
  pooled_width  = Int(ceil(float(width +2*pad[1]-kernel[1]) / stride[1]))+1
  pooled_height = Int(ceil(float(height+2*pad[2]-kernel[2]) / stride[2]))+1

  Threads.@threads for n = 1:num
    @inbounds for c = 1:channels
      for ph = 1:pooled_height
        @simd for pw = 1:pooled_width
          hstart = (ph-1)*stride[2] - pad[2] + 1
          wstart = (pw-1)*stride[1] - pad[1] + 1
          hend = min(hstart + kernel[2] - 1, height)
          wend = min(wstart + kernel[1] - 1, width)
          hstart = max(1, hstart)
          wstart = max(1, wstart)

          maxval = T(-Inf)
          maxw = 0
          maxh = 0
          for w = wstart:wend
            for h = hstart:hend
              val = input[w,h,c,n]
              if val > maxval
                maxval = val
                maxw = w
                maxh = h
              end
            end
          end
          output[pw,ph,c,n] = maxval
          mask[pw,ph,c,n] = (maxh-1) * width + maxw-1
        end
      end
    end
  end
end
```

SimonDanisch:
another `@simd` at `if val > maxval` puts it down to 14 (so it's now 14ms vs 9ms)

SimonDanisch:
this is 10ms
```julia
function inner_pooling(n, width, height, channels, pooled_width, pooled_height, input, output, mask, kernel, pad, stride)
    @inbounds for c = 1:channels
      for ph = 1:pooled_height
        @simd for pw = 1:pooled_width
          hstart = (ph-1)*stride[2] - pad[2] + 1
          wstart = (pw-1)*stride[1] - pad[1] + 1
          hend = min(hstart + kernel[2] - 1, height)
          wend = min(wstart + kernel[1] - 1, width)
          hstart = max(1, hstart)
          wstart = max(1, wstart)

          maxval = -Inf
          maxw = 0
          maxh = 0
          for w = wstart:wend
            @simd for h = hstart:hend
              val = input[w, h, c, n]
              bigger = val > maxval
                maxval = ifelse(bigger, val, maxval)
                maxw = ifelse(bigger, w, maxw)
                maxh = ifelse(bigger, h, maxh)
            end
          end
          output[pw,ph,c,n] = maxval
          mask[pw,ph,c,n] = (maxh-1) * width + maxw-1
        end
      end
    end
end

function pooling(input::Array{T}, output::Array, mask::Array, kernel, pad, stride) where T
    width, height, channels, num = size(input)
    pooled_width  = Int(ceil(float(width +2*pad[1]-kernel[1]) / stride[1]))+1
    pooled_height = Int(ceil(float(height+2*pad[2]-kernel[2]) / stride[2]))+1

    Threads.@threads for n = 1:num
        inner_pooling(n, width, height, channels, pooled_width, pooled_height, input, output, mask, kernel, pad, stride)
    end
end
```

SimonDanisch:
And the GPU version: 0,026ms
```julia
using CLArrays
using GPUArrays: linear_index
function inner_pooling(state, width, height, channels, pooled_width, pooled_height, input, output, mask, k, pad, stride)
    n = linear_index(state)
    @inbounds for c = 1:channels
        for ph = 1:pooled_height
            for pw = 1:pooled_width
                hstart = (ph-1)*stride[2] - pad[2] + 1
                wstart = (pw-1)*stride[1] - pad[1] + 1
                hend = min(hstart + k[2] - 1, height)
                wend = min(wstart + k[1] - 1, width)
                hstart = max(Int(1), hstart)
                wstart = max(Int(1), wstart)
                maxval = -Inf
                maxw = 0
                maxh = 0
                for w = wstart:wend
                    for h = hstart:hend
                        val = input[w, h, c, n]
                        bigger = val > maxval
                        maxval = ifelse(bigger, val, maxval)
                        maxw = ifelse(bigger, w, maxw)
                        maxh = ifelse(bigger, h, maxh)
                    end
                end
                output[pw,ph,c,n] = maxval
                mask[pw,ph,c,n] = (maxh-1) * width + maxw-1
            end
        end
    end
end

function pooling(input::CLArray, output::CLArray, mask::CLArray, kernel, pad, stride)
    width, height, channels, num = size(input)
    pooled_width  = Int(ceil(float(width +2*pad[1]-kernel[1]) / stride[1]))+1
    pooled_height = Int(ceil(float(height+2*pad[2]-kernel[2]) / stride[2]))+1
    args = (width, height, channels, pooled_width, pooled_height, input, output, mask, kernel, pad, stride)
    gpu_call(inner_pooling, input, args, num)
end
cl_arrs = CLArray.((input, output1, mask1))
pooling(CLArray(input), CLArray(output1), CLArray(mask1), kernel, pad, stride)
@btime pooling($(cl_arrs...), $(cl_arrs[2]), $(cl_arrs[3]), $kernel, $pad, $stride)
```

SimonDanisch:
Woops, too good to be true, forgot to synchronize! it's 42ms which makes more sense considering it's a gamer gpu with Float64!

MikeInnes:
See #9 for an ImageFiltering.jl-based version of straight convolutions, at any rate.

dfdx:
> See #9 for an ImageFiltering.jl-based version of straight convolutions, at any rate.

Looks good! I'm getting ~2x improvement on my tests. How hard it would be to add strides and padding? I guess it would invalidate some if ImageFiltering's  algorithm and may slow down convolution significantly.

MikeInnes:
Padding at least can be implemented as a separate `pad` function (which can create a kind of view). That will work out of the box with ImageFiltering and we can dispatch appropriately to CUDNN. Same for reflecting kernels eventually.

Looking over ImageFiltering I think we can also use it for pooling (although we might eventually want a custom implementation to make the gradient calculation faster).

Striding I'm not sure about. Perhaps worth an issue on ImageFiltering to get Tim's thoughts.

dfdx:
I've just pushed a build script for Windows using WinRPM, @MikeInnes could you please activate AppVeyor for this repository so I could debug it on Windows?

dfdx:
I forgot I can [turn it for my fork](https://ci.appveyor.com/project/dfdx/nnlib-jl). Nevertheless, would be nice to have AppVeyor enabled for FluxML/NNlib as well to see its  build status. 

Currently this PR builds and passes tests on Julia 0.6 in Linux, macos and Windows (failed status is because of Julia nightlies). Does it look good enough for all users? Of course, we can  try other (e.g. pure-Julia) implementations later without making our users wait.

MikeInnes:
It's not ideal, but it's probably the lesser evil. @iblis17, I know you're on a less common system, any chance you could check this works for you?

What do you think of the API I outlined in #9? It may be that the C++ code can't support it in full generality, but even so it'd be nice to make things as Julian as possible for the future. Of course it's also good to get going and finalise the API later.

REQUIRE
```diff
@@ -1,2 +1,3 @@
 julia 0.6
-Requires
+WinRPM
```
iblis17:
```diff
diff --git a/REQUIRE b/REQUIRE
index d376198..98e2c02 100644
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
-WinRPM
-Requires
\ No newline at end of file
+@windows WinRPM
+Requires
```

----

I got this

```julia
julia> Pkg.build("NNlib")           
ERROR: WinRPM is not an installed package                                
Stacktrace:                         
 [1] build!(::Array{String,1}, ::Set{Any}, ::String) at ./pkg/entry.jl:623                                                                         
 [2] build!(::Array{String,1}, ::Set{Any}, ::String) at ./pkg/entry.jl:624                                                                         
 [3] build!(::Array{String,1}, ::Dict{Any,Any}, ::Set{Any}) at ./pkg/entry.jl:635                                                                  
 [4] build(::Array{String,1}) at ./pkg/entry.jl:650                      
 [5] (::Base.Pkg.Dir.##4#7{Array{Any,1},Base.Pkg.Entry.#build,Tuple{Array{String,1}}})() at ./pkg/dir.jl:36                                        
 [6] cd(::Base.Pkg.Dir.##4#7{Array{Any,1},Base.Pkg.Entry.#build,Tuple{Array{String,1}}}, ::String) at ./file.jl:70                                 
 [7] #cd#1(::Array{Any,1}, ::Function, ::Function, ::Array{String,1}, ::Vararg{Array{String,1},N} where N) at ./pkg/dir.jl:36                      
 [8] build(::String, ::Vararg{String,N} where N) at ./pkg/pkg.jl:254
```

deps/build.jl
```diff
@@ -0,0 +1,37 @@
+#! /usr/bin/env julia
+
+path = joinpath(dirname(@__FILE__), "..", "src")
+
+if is_windows()
+
```
iblis17: well, some trailing space in this `if` block.

REQUIRE
MikeInnes: You probably just need to do `Pkg.resolve()`. Although this should probably be `@windows WinRPM` @dfdx.

deps/build.jl
```diff
@@ -0,0 +1,37 @@
...
+        error("no windows c++ compiler (cl.exe) found, and WinRPM with g++ is failing as well.")
+    end
+    
+end
+
+if is_unix()
+    cd(path) do
+        # Note: on Mac OS X, g++ is aliased to clang++.
+        run(`g++ -c -fPIC -std=c++11 conv.cpp`)
+        run(`g++ -shared -o conv.so conv.o`)
```
iblis17: The default compiler on FreeBSD is clang.
And `cc/c++` is a soft link to it. (But i don't sure `cc/c++` are available on other platform for not

```diff
diff --git a/deps/build.jl b/deps/build.jl
index 995783e..cca00f8 100644
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -31,7 +31,7 @@ end               
 if is_unix()                       
     cd(path) do                    
         # Note: on Mac OS X, g++ is aliased to clang++.                 
-        run(`g++ -c -fPIC -std=c++11 conv.cpp`)
-        run(`g++ -shared -o conv.so conv.o`)
+        run(`c++ -c -fPIC -std=c++11 conv.cpp`)
+        run(`c++ -shared -o conv.so conv.o`)
     end                            
 end
```

src/Makefile
```diff
@@ -0,0 +1,20 @@
...
+libnnlib.so: $(OBJ)
+	$(CXX) --shared $(CFLAGS) $^ -o $@
+
+%.o: %.cpp
+	$(CXX) -c $(CFLAGS) $< -o $@
+
+clean:
+	-rm *.o *.cu libknet8.so
+
+.PRECIOUS: %.cu
```
iblis17: hmm, `Pkg.build("NNlib")` won't invoke this Makefile, right?

iblis17:
and... `Pkg.test("NNlib")` passed for me.

iblis17:
> What do you think of the API I outlined in #9? It may be that the C++ code can't support it in full generality, but even so it'd be nice to make things as Julian as possible for the future. Of course it's also good to get going and finalise the API later.

sorry about that I currently no more idea about the API. I seldom use convolution in my personal research :p

REQUIRE
dfdx: Ah, right. Added it.

deps/build.jl
dfdx: `c++` works for me on Ubuntu 16.04, will  update.

src/Makefile
dfdx: Yep, leftover from the previous version. Removed it.

dfdx:
@iblis17 Could you please run the latest version of this branch and tell if it build fine this time?

iblis17:
@dfdx everything works fine for me now :)

src/NNlib.jl
```diff
@@ -2,23 +2,11 @@ module NNlib
...
-    softsign,
-    conv2d,
-    conv2d_grad_w,
-    conv2d_grad_x,
-    pool,
-    pool_grad
+export σ, relu, leakyrelu, elu, swish, selu, softplus, softsign,
+  softmax, conv2d, pool
+
+const libnnlib = Libdl.find_library(is_windows() ? "conv.dll" : "conv.so", [@__DIR__])
```
SimonDanisch: You should use `Libdl.dlext` :)

MikeInnes: In this case the build script actually does .so even on OS X. That could possibly be cleaner though

MikeInnes:
Ok, let's go with this for now, and we can always experiment with the API later. I'm a bit wary of how complex this is on Windows, but if we get complaints we can possibly just host binaries somewhere.

Thanks again @dfdx!

ChrisRackauckas:
I only noticed this merged because I did `Pkg.update()` and WinRPM failed, and then I tried to show this library to someone and it failed to build on their Mac because they didn't have the XCode command line tools installed. Not having these problems was the main thing Flux.jl had over everything else, so I hope this is very temporary.

iblis17:
@ChrisRackauckas well, it's performance war (?)
Please checkout this also
https://github.com/FluxML/NNlib.jl/pull/9

dfdx:
> I only noticed this merged because I did Pkg.update() and WinRPM failed

What exactly went wrong? I believe WinRPM should work fine on Windows (where else it could :D), and it should be enough to compile C++ part. If you can provide more details, we may come up with a better approach (e.g. host compiled binaries somewhere just like Python/Conda does for Windows). 

> it failed to build on their Mac because they didn't have the XCode command line tools installed

So we must install it or MacPorts automatically during the build. Can somebody with macos comment on this? How hard it would be to make it working? 

Alternatively, we can again host pre-compiled binaries somewhere. I can create an FTP server for this, but we need to decide on who can build them (Travis has significantly reduced their support for macos recently, build take hours there now). 

> Not having these problems was the main thing Flux.jl had over everything else

An empty package doesn't have any problems at all, but it's not of much practical use. Rather than trying to reduce functionality we must think how to make it work for as much users as possible.

MikeInnes:
That's unfortunate. It's not exactly clear to me how to fix things on mac, and even with homebrew or something it doesn't look like we can test this on our CI.

Given that these binaries are relatively unlikely to change (unless we go ahead and rewrite them in Julia, which makes thing simple), I think it's ok to make releases by hand. We don't need anything advanced here, just for someone to build the dll on each system and put it up somewhere.

dfdx:
Sounds good to me. I can provide hosting and binary for Linux x64, as well as PR to download this stuff. We also need binaries for: 

 * Linux x86
 * Windows
 * macos
 * FreeBSD

Anyone willing to build these binaries and send to me?

SimonDanisch:
This sounds like a job for BinaryBuilder.jl! It works pretty smoothly by now :)

MikeInnes:
I think we can probably just do OS X and Windows – the more obscure OSs are proportianately more likely to have a decent c++ compiler.

I'm currently stuck on how to tweak this build script to build a 32bit DLL on 64bit windows, anyone know anything about this? Adding `-m32` gives a lot of link errors, I think I need to install additional WinRPM binaries but I'm not sure what.

ChrisRackauckas:
>An empty package doesn't have any problems at all, but it's not of much practical use. 

That's just silly. Come on.

>What exactly went wrong? I believe WinRPM should work fine on Windows (where else it could :D), and it should be enough to compile C++ part. If you can provide more details, we may come up with a better approach (e.g. host compiled binaries somewhere just like Python/Conda does for Windows).

Looks like I'm not getting that build issue anymore, so it might've been a temporary WinRPM snag not downloading the binary. It does that sometimes.

dfdx:
> That's just silly. Come on.

Why? Literally any code you add can make the package unusable for some folks - even if it's in pure Julia, your code may use 3rd party libraries with binary dependencies, or require amount of RAM larger than the user has (e.g. on Raspberry Pi), or not work without internet connection / behind a firewall, or require graphical environment, or whatever else. For this package, even with C++ part, we have only 2 requirements: 

1. One of the popular operating systems.
2. Internet connection. 

Which sound pretty fair.

dfdx:
> This sounds like a job for BinaryBuilder.jl! It works pretty smoothly by now :)

Looks promising, although from a quick look I can't figure out the whole development cycle. Am I right that build run in Docker containers on Travis and automatically uploaded to some storage? Does this require a separate repo (something like NNlibBuilder) or can be part of this package?

ChrisRackauckas:
>Why?

You related NNlib (and by relation, Flux.jl) without convolutions to an empty package to try and defend your point. What we're actually talking about is a one feature difference causing a large change in compatibility. Your analogy is objectively silly so don't even try to defend it and move on. 

>For this package, even with C++ part, we have only 2 requirements:

It's well known that binary dependencies will cause more issues. Your two requirements are already objectively false...

1. Tests fail on Windows 32-bit because of binary builder right now? Well, according to stats from Tony, about 30% of Julia downloads from Julialang were 32-bit Windows (last year). Not popular enough for you? (I was shocked to hear it too, but hey, not everyone is in a developed Western country with the latest tech)
2. It doesn't run on Mac without having XCode and XCode command line tools installed to run the compilation. This actually tripped up people in my lab because, well, they do science and not software engineering so this is actually a big barrier for people coming from MATLAB where it "just works".
3. WinRPM fails all the time. It fails on the nightlies all the time. 

These are the issues that cause many Julia packages with binary deps to be shaky on platforms that aren't Linux. Flux.jl wasn't one of them. Now it is, that's just true. Whether that's acceptable or not is a subjective choice. I don't think adding one feature which isn't used in all NNs is worth that cost. It especially shows itself when you try to teach a workshop on machine learning in Julia. The "install the package" part goes from "just `Pkg.add`", to installing XCode tools, debugging people's WinRPM versions, etc. I know that most devs use Linux, but this is a serious issue when trying to train newcomers that cannot just be laughed at. Seriously, just try to get a bunch of Mac/Windows MATLAB users to install it without guidence on the first try and you'll see what I mean.

But hopefully it at least goes away when there's a native Julia version.

dfdx:
> These are the issues that cause many Julia packages with binary deps to be shaky on platforms that aren't Linux

So help us make this package stable on all supported platforms. Shouting out how bad this PR is doesn't help at all, describing what exactly goes wrong does. In particular: 

1. Can you elaborate on WinRPM fails? Why does it happen? If it's caused by active development, we can just pin it to a specific version in requirements, or work out some other, more stable workflow. 
2. Can somebody with Mac from your lab (or wherever else) compile the binary and send it to us? We can use it instead of compiling until we get binaries generated automatically (e.g. via BinaryBuilder.jl). 
3. The same question about 32-bit Windows (although virtual machine may work in this case).

And just to explain why to bother with this "one feature difference" - the whole deep learning is driven by computer vision tasks (outside of CV, much more shallow nets are normally utilized), most tutorials use ConvNets as one of the first examples (e.g. [1](https://www.tensorflow.org/tutorials/), [2](http://pytorch.org/tutorials/beginner/deep_learning_60min_blitz.html), [3](https://mxnet.incubator.apache.org/tutorials/)), and they have been explicitly asked for in Flux.jl [several](https://discourse.julialang.org/t/flux-3-now-with-100-more-julia/5824/2) [times](https://discourse.julialang.org/t/flux-3-now-with-100-more-julia/5824/12). So maybe this PR temporarily breaks code for some users (which we didn't know when merging, e.g. build for 32-bit Windows [worked fine](https://ci.appveyor.com/project/dfdx/nnlib-jl) on Julia 0.6), but it also lets many others to get involved.

ChrisRackauckas:
>So help us make this package stable on all supported platforms. 

I already detailed a method that would work for all platforms: get rid of the binary dep. This didn't need to be in the whole stack as a requirement. It should've just been a separate package with the binary dep that implements the convolution and is compatible with Flux's interface. Then the user could opt into that at any time, and even after a native Julia one is implemented it could still be around for testing/benchmarking against. But instead it was put in as the only binary requirement when it's an isolated function and has isolated tests, meaning it's essentially already it's own package, just tagged onto an existing package and giving it installation issues. Mitigating those issues still doesn't fix the core issue that it didn't make sense to add the dep here in the first place and proper development practices would've made this a non-issue.

dfdx:
Thank you for your helpful response and answering my questions. With them this package will be so much better. 

I could describe in detail what complexities putting convolutions into a separate package would bring, and how messy dependencies would be. But I doubt my words will ever have any effect, so sorry, but I have to get back to work.

iblis17:
Inspire by this:
https://github.com/andreasnoack/JuliaBlas.jl/blob/master/src/gemmFull.jl#L95-L454
We still can have pure Julia implementation with performance critical part baked by LLVM IR.

PR #7, opened 2017-10-13
act: implement exponential linear unit

iblis17:


MikeInnes:
This is great, thanks. Small tweak – could you change the `1` literals to `one(x)`? (The `0` in the comparison is fine as is, I think).

iblis17:
done 👌

iblis17:
🤔 , no idea why GH do not update yet

MikeInnes:
Github was just having a slow day yesterday. Thanks for the patch!

PR #8, opened 2017-10-18
act: Swish, a self-gated activation function

iblis17:
Ref: [Swish: a Self-Gated Activation Function](https://arxiv.org/pdf/1710.05941.pdf)


MikeInnes:
Awesome. If nothing else we can say we get activation functions faster than anyone else :)

PR #10, opened 2017-11-05
Add selu, softsign and softplus functions, add docs and tests

Drvi:
Hi,

Here I've added the selu, softsign and softplus activation functions, added some tests and added a "unicode plot" view of the function to the docstrings (inspired by JuliaMLs LossFunctions.jl).

I'm a julia beginner so I assume that I've not done things in the best way possible and I'd be grateful for some feedback.

**What is missing:**
I didn't check that the function definitions play nicely with julias GPU support and automatic differentiation. There are more activation functions that could be added, most of them with learnable parameters (I'm not sure how to design such functions so that they work smoothly with e.g. Flux models).

In LossFunctions.jl, they provide some additional functionality for each loss like checking the functions mathematical properties (e.g. continuity) and their explicit first and second derivatives and so on. Something similar is also possible for activation functions as well (an interesting example would be the recommended dropout value, which is typically 0, but for selu its authors recommend a value corresponding to the minimum of the selu function, although I'm not sure if there are more examples like that).


src/activation.jl
```diff
@@ -3,43 +3,274 @@
...
+        │    .__.--/''`      |                   │
+        │-/''`               |                   │
+     -1 │                    |                   │
+        └────────────────────────────────────────┘
+        -3                                       3
+                            x
+```
 """
-leakyrelu(x, a = oftype(x, 0.01)) = max(a*x, x)
+leakyrelu(x::T, a::T) where T = max(a * x, x)
```
MikeInnes: Can you explain these changes, here and for `elu`?

MikeInnes:
Ok, I think I understand that with the changes you were just getting Float32 etc to propagate properly. I've made that a bit simpler.

Otherwise this looks good! The tests are useful and the plots are a great idea.

Drvi:
Thanks for polishing this up! For some reason I struggled with the docstrings (and the `selu` is still rendering weird in vs code).

Should I add the derivatives? E.g.

```julia
∂σ(x) = σ(x) * (one(x) * σ(x))
```

Also, in lossfunctions.jl they provide some additional information for each function. 
They are using types for that but maybe something simple like:

```julia
ismonotonic(a) = ismonotonic(Val{a})
ismonotonic(::Type{Val{σ}}) = true

# julia> ismonotonic(σ)
# true
```
would suffice (not sure if there is demand for that). Again, I'm a julia beginner, but I'd like to help:)

MikeInnes:
We don't need to define gradients explicitly, ForwardDiff will help with that :)

I wouldn't be against things like `ismonotonic` as such, but it'd be better to add things where there's a clear use case. (FYI – you can write that as `ismonotonic(::typeof(σ)) = true`).

What kind of areas are you interested in helping in? Perhaps I can give some pointers. Within NNlib specifically, one really useful thing would be to port the current [C++ convolution kernels](https://github.com/FluxML/NNlib.jl/blob/38ba0254e3b1103e6d76546d4e82cd4c5fba378e/src/conv.cpp) into Julia.

Drvi:
I'll have a look at the convolution kernels. I hoped I could start this weekend but I barely had the time to look at the code. Hopefully, I'll soon have something to show. 

Here are some interesting articles about different approaches to convolutions:

https://arxiv.org/pdf/1709.03395.pdf
https://arxiv.org/pdf/1704.04428.pdf

From their benchmarks it seems that im2row is faster that im2col for their C implementation, but porting im2col probably makes more sense since julia is column-major. The kn2col (kn2row) should be less memory demanding and is probably worth considering.

I also found https://arxiv.org/pdf/1509.09308.pdf but I haven't read it carefully yet.

MikeInnes:
Good stuff. It would be great to experiment with these (although it's also fine to have an initial working version that isn't optimal).

Also check out Imagefiltering.jl. I'm not sure if it's possible to make full use of it, but if we can it's very fast.

PR #12, opened 2017-11-17
Added tanh activation function

Vaibhavdixit02:
I guess this would be an acceptable addition?


iblis17:
well, there is a `Base.tanh`

Vaibhavdixit02:
Oh yes, should I go close it then?

MikeInnes:
Yeah, we can just use `Base.tanh` for this. But thanks for the effort nonetheless!

PR #13, opened 2017-11-30
Pre-built binaries

MikeInnes:
Testing CI here.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,15 +3,12 @@
 path = joinpath(dirname(@__FILE__), "..", "src")
 
 if is_apple()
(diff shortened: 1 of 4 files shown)


deps/build.jl
```diff
@@ -2,7 +2,13 @@
 
 path = joinpath(dirname(@__FILE__), "..", "src")
 
-if is_windows()
+version = "0.1.0"
+
+if is_apple()
+  download("https://www.dropbox.com/s/syt6q6qv6ehc4p0/nnlib-$version.dylib?dl=1",
```
iblis17: how about hosting the binary via GitHub release page?

MikeInnes: I'm working on getting a real host for these, for now these should hold up.

MikeInnes:
This doesn't work on win32, but that's the same with the current build script anyway.

PR #16, opened 2017-12-03
Stack overflow fix

### LINKED ISSUE #15: elu causes stack overflow
```julia
Dense(2,2,elu)(randn(2))
```

Drvi:
This should fix #15.
I've added some tests about handling ForwardDiff.Dual as input.

Please review:-)

Thanks


MikeInnes:
This is useful, thanks. The original reason for adding `AbstractFloat` was that integers sometimes cause type stability issues (e.g. `max(x, a*x)` can return int or float), so we need to be careful about that.

I think the best option is to change `AbstractFloat` to `Real` and cast to float explicitly where it might be an issue. e.g.

```julia
leakyrelu(x::Reak, a = oftype(x/1, 0.01)) = max(a*x, x/1)
```

Can you give that approach a go?

MikeInnes:
I had a go at the simpler version (56bb47335f71d84edd0283475d21a68298aa53a6) and kept your tests in 56bb47335f71d84edd0283475d21a68298aa53a6. Thanks!

PR #17, opened 2017-12-10
Replace old binary artifact with the new one during Pkg.build

dfdx:
Otherwise build fails on Linux when run more than once.


MikeInnes:
Thanks, merged in 32f94cd985b4d1c8146739a19c20316638ca9c26

PR #22, opened 2018-01-12
Adding logsoftmax

boathit:
Adding `logsoftmax` function by following the convention of `softmax`.

### DIFF
--- a/src/logsoftmax.jl
+++ b/src/logsoftmax.jl
@@ -1,22 +1,22 @@
 using Base.Threads
 
 function logsoftmax!(out::AbstractVecOrMat, xs::AbstractVecOrMat)
-    @threads for j = 1:size(xs, 2)
-        @inbounds begin
-            xi_max = xs[1, j]
-            for i = 1:size(out, 1)
-                xi_max = max(xi_max, xs[i, j])
-            end
-            s = zero(eltype(out))
-            for i = 1:size(out, 1)
-                s += exp(xs[i, j] - xi_max)
-            end
-            for i = 1:size(out, 1)
-                out[i, j] = xs[i, j] - log(s) - xi_max
-            end
-        end
+  @threads for j = 1:size(xs, 2)
+    @inbounds begin
+      xi_max = xs[1, j]
+      for i = 1:size(out, 1)
+        xi_max = max(xi_max, xs[i, j])
+      end
+      s = zero(eltype(out))
+      for i = 1:size(out, 1)
+        s += exp(xs[i, j] - xi_max)
+      end
+      for i = 1:size(out, 1)
+        out[i, j] = xs[i, j] - log(s) - xi_max
+      end
     end
-    out
+  end
+  out
 end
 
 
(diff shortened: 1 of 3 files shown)


MikeInnes:
Awesome, thanks! Can you add a test for this? You can perhaps compare it to regular `softmax` to make sure it's correct.

boathit:
Sure. Actually, I have compared it with the PyTorch implementation `torch.nn.functional.log_softmax()`  and the outputs are consistent. I will add several test cases in `test.jl`.

MikeInnes:
Great, thanks!

Just FYI, this won't be usable with Flux until you add gradient definitions there too; happy to give pointers there if you need them.

boathit:
Hi @MikeInnes, I indeed added the gradient definition, see the line `∇logsoftmax(Δ, xs) = ∇softmax(Δ ./ softmax(xs), xs)`

MikeInnes:
Sure, that's great; but Flux won't know about the gradient unless you tell it, so you need a definition [something like this](https://github.com/FluxML/Flux.jl/blob/b0b8c9dbd1303e6363a881553175bb7d1ebdceb7/src/tracker/lib.jl#L137) on the Flux side.

boathit:
Oh, I see. Perhaps I can make another PR for Flux. Besides, even though the current `logsoftmax()` works with CuArrays but I am not sure whether CUDNN has more efficient GPU version.

MikeInnes:
Yeah, this will be very slow for CuArrays. For softmax we [add CUDNN overloads](https://github.com/JuliaGPU/CuArrays.jl/blob/00276e3c234188d728a94c12d88674839de642dd/src/dnn/nnlib.jl#L10-L18) in CuArrays itself, and the same thing could be done for `logsoftmax`.

boathit:
Hi @MikeInnes, as spoken a [PR #153](https://github.com/FluxML/Flux.jl/pull/153) has been made to Flux.jl to register backward function for `logsoftmax`.

MikeInnes:
:+1: Thanks!

src/logsoftmax.jl
```diff
@@ -0,0 +1,34 @@
...
+      xi_max = xs[1, j]
+      for i = 1:size(out, 1)
+        xi_max = max(xi_max, xs[i, j])
+      end
+      s = zero(eltype(out))
+      for i = 1:size(out, 1)
+        s += exp(xs[i, j] - xi_max)
+      end
+      for i = 1:size(out, 1)
+        out[i, j] = xs[i, j] - log(s) - xi_max
```
vchuravy: Do you remember why you subtract `xi_max` here?
IIRC the definition of logsoftmax normally is

```
x_max = maximum(xs)
s = log(sum(exp.(xs - x_max)))
out = xs .- s
```


```

Comment (boathit) [reply src/logsoftmax.jl]: I think we are doing the same thing as pytorch. If you replace `s` with `log(sum(exp.(xs - x_max))) + x_max` you will get 
```
out = xs .- log(sum(exp.(xs - x_max))) - x_max
```

PR #26, opened 2018-01-30
add logsigmoid

boathit:
Adding `logsigmoid` which will be useful for defining the Binary Cross Entropy loss [#143](https://github.com/FluxML/Flux.jl/issues/143)


MikeInnes:
Cool, thanks!

PR #27, opened 2018-01-31
Added Conv3D layer

tejank10:
**Changelog**
1. Redefined `im2col_dims` to support n-dimensions.
2. Added `Conv3D` layer
3. Changed function naming in `conv.cpp` and `conv.jl` in accordance with 2d and 3d layers.


MikeInnes:
Great work, and I'm happy with this as it stands.

The main thing right now is that we need to build a new binary for each platform. We're not really set up for this, so it'd be useful to integrate [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl) somehow.

Also, would you be able to hook up the CUDNN kernels for this, [like so](https://github.com/JuliaGPU/CuArrays.jl/blob/70f6689de8c0ef5f750bddab642cd1b36f5d4ade/src/dnn/nnlib.jl#L32)?

tejank10:
Any clue on integrating with [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl)?

MikeInnes:
I don't know any more than you do right now, just having looked at the readme myself. But I'll take a look at it soon if you don't get to it first.

tejank10:
Cool, I'll be happy to do that!

ChrisRackauckas:
I could help with BinaryBuilder/BinaryProvider a bit, but would like to see it as pure Julia code.

tejank10:
I coded the pooling layers to Julia. For `avgpool3d`, `gradtest` passes for smaller matrices (of size < 4x4x4). But it is failing for sizes above that. `avgpool3d` code is modelled after `avgpool2d` which is passing the `gradtest`. Can anyone point me to any possible bug that could be causing this?

MikeInnes:
Was the C++ version of `avgpool3d` working properly? Sounds like your best bet is to check the Julia version against that on a range of inputs.

tejank10:
Updated my branch with the latest changes in Flux. Now, I am facing this error for many functions:

`  MethodError: no method matching pool3d(::TrackedArray{…,Array{Float64,5}}; window=2, padding=0, mode=0)`
as these function accept `Array` instead of `TrackedArray`

tejank10:
Wrote convolutions and pooling as pure Julia code

CarloLucibello:
great! how is performance compared to c++ implementations?

tejank10:
I evaluated the time taken for performing the gradtests of the convolution and pool layers. I ran the code multiple times. The C++ version took 10.387s on average whereas that of Julia took 10.67s.

CarloLucibello:
awesome!

tejank10:
@MikeInnes Please review.

MikeInnes:
Really awesome work!

If you're up for an extra challenge, it might be fun to try and make a `conv` function that can work on an arbitrary number of dimensions -- so we don't need separate versions for 2d, 3d, 4d, etc.

tejank10:
Thanks, Mike! I'll start work on that soon. Thinking of a recursive approach.

datnamer:
Is there still a special behind the scenes conv3d that gets dispatched by the nd function? Or is the generic implementation fast enough now? Or neither?

MikeInnes:
The `conv` function currently just provides the API we want; it dispatches to specialised kernels for the implementation.

PR #31, opened 2018-03-04
N dimensional conv & pool

tejank10:
convolutions and pooling for N-dimensional image

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -61,29 +61,29 @@ function ∇conv_data!(dx::AbstractArray{T,3}, dy::AbstractArray{T,3},
     return dx
 end
 
-conv!(y::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
-      pad = 0, stride = 1) where T =
-  conv2d!(y, x, w, padding = pad, stride = stride)
+#conv!(y::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
+#      pad = 0, stride = 1) where T =
+#  conv2d!(y, x, w, padding = pad, stride = stride)
 
-∇conv_filter!(dw::AbstractArray{T,4}, dy::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
-              pad = 0, stride = 1) where T =
-  conv2d_grad_w!(dw, x, w, dy, padding = pad, stride = stride)
+#∇conv_filter!(dw::AbstractArray{T,4}, dy::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
+#              pad = 0, stride = 1) where T =
+#  conv2d_grad_w!(dw, x, w, dy, padding = pad, stride = stride)
 
-∇conv_data!(dx::AbstractArray{T,4}, dy::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
-            pad = 0, stride = 1) where T =
-  conv2d_grad_x!(dx, x, w, dy, padding = pad, stride = stride)
+#∇conv_data!(dx::AbstractArray{T,4}, dy::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
+#            pad = 0, stride = 1) where T =
+#  conv2d_grad_x!(dx, x, w, dy, padding = pad, stride = stride)
 
-conv!(y::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
-      pad = 0, stride = 1) where T =
-  conv3d!(y, x, w, padding = pad, stride = stride)
+#conv!(y::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
+#      pad = 0, stride = 1) where T =
+#  conv3d!(y, x, w, padding = pad, stride = stride)
 
-∇conv_filter!(dw::AbstractArray{T,5}, dy::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
-              pad = 0, stride = 1) where T =
-  conv3d_grad_w!(dw, x, w, dy, padding = pad, stride = stride)
+#∇conv_filter!(dw::AbstractArray{T,5}, dy::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
+#              pad = 0, stride = 1) where T =
+#  conv3d_grad_w!(dw, x, w, dy, padding = pad, stride = stride)
 
-∇conv_data!(dx::AbstractArray{T,5}, dy::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
-            pad = 0, stride = 1) where T =
-  conv3d_grad_x!(dx, x, w, dy, padding = pad, stride = stride)
+#∇conv_data!(dx::AbstractArray{T,5}, dy::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5};
+#            pad = 0, stride = 1) where T =
+#  conv3d_grad_x!(dx, x, w, dy, padding = pad, stride = stride)
 
 conv!(y::A, x::A, w::A; pad = 0, stride = 1) where A<:AbstractArray =
   conv_nd!(y, x, w, padding = pad, stride = stride)
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -108,22 +108,22 @@ end
 
 # Interface
 
-maxpool(x::AbstractArray, k; pad = map(_->0,k), stride = k) =
+maxpool(x::AbstractArray, k, pad = map(_->0,k); stride = k) =
   maxpool!(similar(x, pdims(size(x), k, pad, stride)),
-           x, k, pad = pad, stride = stride)
+           x, k, pad, stride = stride)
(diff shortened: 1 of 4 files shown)


tejank10:
@MikeInnes Please review

dfdx:
Any benchmarks comparing it with C++ version? 

Also I'm not sure about making padding a required argument, can't we just use something like `pad=tuple(zeros(Int, N)...)` to provide default value for N-dimensional array?

tejank10:
That's a good idea :). I have implemented that now.

dfdx:
I'd test ND convolution in Julia against existing 2D and 3D convolutions in C++. In fact, there should have been benchmarks for 2D/3D convolutions between Julia and C++ implementations, but I lost track of NNlib for a while and missed when it has been reimplemented in pure Julia. Previously [we used](https://github.com/FluxML/NNlib.jl/pull/2) benchmarks from [Mocha.jl](https://github.com/pluskid/Mocha.jl/blob/master/benchmarks/native-im2col/im2col-bm.jl), it should be easy to adapt them for new implementations. 

If C++ wins, it's not necessarily a show stopper, but we need to know how much we lose with pure-Julia and generic implementation. Also note that CUDNN - perhaps the primary provider for fast convolutions on GPU - only supports 2D and 3D convolutions, so we need to position generic ND convolutions correctly to not confuse deep learning people.

tejank10:
Thanks a lot for your input, it helped me a lot in benchmarking conv. 
Here are the results:

`Julia conv2d TrialEstimate(529.985 μs)`  
`C conv2d TrialEstimate(535.165 μs)`  
`Julia conv3d TrialEstimate(41.316 ms)`  
`C conv3d TrialEstimate(26.435 ms)`
`Julia conv_nd 2dTrialEstimate(761.929 ms)`  
`Julia conv_nd 3dTrialEstimate(51.210 s)`  
`Conv2d Julia vs Conv2d C: TrialRatio(99.03%)`
`Conv3d Julia vs Conv3d C: TrialRatio(156.29%)`  
`Conv_nd 2d Julia vs Conv2d  C: TrialRatio(142372.71%)`  
`Conv_nd 3d Julia vs Conv3d C: TrialRatio(193717.42%)`

Well, the results turn out to be poor for generic conv. I guess the main culprit of this is the extra loop we have run for finding the indices.

dfdx:
Hm, 200x slowdown looks a bit too high. Most likely there's just a couple very slow operations that make the whole thing perform poorly. Can you [profile](https://docs.julialang.org/en/latest/manual/profile/) it? It should be pretty easy to identify a bottleneck. 

Also I believe GitHub still doesn't deliver notifications for inline comments, so I'll copy it here: in `conv.jl` on line 161 (and in all similar places) you use `pad=tuple(zeros(Int, N))`, which returns a tuple with a single element of type `Array`, e.g. `([0,0,0],)`. Instead, you can use `pad=tuple(zeros(Int, N)...,)` (note dots before the trailing comma) to unpack array into a tuple and get something like `(0, 0, 0)`.

tejank10:
Thanks for the profiling link, and sorry for the late reply. 
I have profiled the `im2col_nd!`. 
[Here](https://github.com/FluxML/NNlib.jl/files/1821425/prof.txt) are the results. Below are some lines from the file pertaining directly to `im2col_nd!`:
`      1    ./broadcast.jl:192; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      13   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:153; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      97   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:154; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      19   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:155; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      2    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:158; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      5    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:161; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      295  .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:165; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      1    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:167; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      213  .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:168; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      7    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:169; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      20   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:171; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      55   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:173; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      4    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:176; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      2    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:177; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      10   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:178; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      10   .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:179; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      695  .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:181; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`
`      3    .../tejank10/Documents/NNlib.jl/src/impl/conv.jl:182; im2col_nd!(::Array{Float64,3}, ::Array{Float64,2}, ::Tuple{Int64,Int...`

Regarding the default arguments: Thanks for pointing them out, my bad! I'll rectify that soon.

tejank10:
Optimized the code on line 181 to obtain nearly 4.8x speedup over the previous version. Here are the benchmarks:
`Julia conv2d TrialEstimate(462.809 μs)`
`C conv2d TrialEstimate(366.904 μs)`
`Julia conv3d TrialEstimate(32.801 ms)`
`C conv3d TrialEstimate(23.515 ms)`
`Julia conv_nd 2dTrialEstimate(127.972 ms)`
`Julia conv_nd 3dTrialEstimate(9.534 s)`
`Conv2d Julia vs Conv2d C: TrialRatio(126.14%)`
`Conv3d Julia vs Conv3d C: TrialRatio(139.49%)`
`Conv_nd 2d Julia vs Conv2d  C: TrialRatio(34878.79%)`
`Conv_nd 3d Julia vs Conv3d C: TrialRatio(40543.79%)`

tejank10:
More optimizations, 1.4x speedup
```
Julia conv2d TrialEstimate(465.402 μs)
C conv2d TrialEstimate(361.856 μs)
Julia conv3d TrialEstimate(36.782 ms)
C conv3d TrialEstimate(23.582 ms)
Julia conv_nd 2dTrialEstimate(80.823 ms)
Julia conv_nd 3dTrialEstimate(6.802 s)
Conv2d Julia vs Conv2d C: TrialRatio(128.62%)
Conv3d Julia vs Conv3d C: TrialRatio(155.97%)
Conv_nd 2d Julia vs Conv2d  C: TrialRatio(22335.55%)
Conv_nd 3d Julia vs Conv3d C: TrialRatio(28845.23%)
```

dfdx:
There are quite few samples in profiling, let's run it in a loop for, say, 1000 iterations. It should also decrease time for inference since currently it looks like a major runtime cost. 

Yet we are still dozens of times  slower than specialized 2D and 3D convolutions, so it makes sense to dispatch to them in normal cases and use full ND convolutions for arrays with higher dimensions. @MikeInnes does it sound like a reasonable approach for NNlib/Flux?

tejank10:
[Here](https://github.com/FluxML/NNlib.jl/files/1835013/prof.txt) is the profiling result. 
Totally agreed with you on the second point. :+1: .

staticfloat:
@tejank10 Not to add more work for you here, but it would be really useful to add a `dilation` factor here, to skip input pixels and enable dilated (or atrous) convolution.

I'm imagining we could have a parameter called `dilation` that is a tuple of integers similarly to `pad`, such that if I had a 2d convolution and I added `dilation = (1, 1)` I would get something like this:

image

jekbradbury:
There are merge markers left in the most recent commit.

staticfloat:
This is going to need to be completely reworked now that #94 has landed.  It shouldn't be too difficult, but may take some work.  For now, I'm going to close this.  Thanks for all your work @tejank10!

PR #33, opened 2018-03-17
Add maxout

americast:
Adds a generic form of `maxout`.  ([Link to paper](https://arxiv.org/pdf/1302.4389.pdf)).

It decreases the number of elements in a particular axis of a matrix, by dividing the axis into `n` chunks and comparing the corresponding elements. The maxima are retained. This implementation behaves merely like an activation layer, decreasing the number of elements in an axis by one of its proper divisors.

An alternate implementation of `maxout` may be found at https://github.com/FluxML/Flux.jl/pull/209.

### DIFF
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -3,7 +3,7 @@ module NNlib
 using Requires
 
 export σ, sigmoid, relu, leakyrelu, elu, swish, selu, softplus, softsign, logσ, logsigmoid,
-  softmax, logsoftmax, maxpool, meanpool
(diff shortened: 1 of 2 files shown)


MikeInnes:
Looks good – however, it looks like this can be a normal function rather than a struct, with appropriate keyword arguments and such. Also, how's GPU compatibility?

ToucheSir:
Since this PR was first opened, NNlib has become more focused on providing reusable functions rather than layer structs. Also, Flux now has a [Maxout layer](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Maxout).

PR #36, opened 2018-04-16
Scalar pad and stride for maxpool and meanpool

tejank10:
Scalar pad and stride as suggested in issue [#235](https://github.com/FluxML/Flux.jl/issues/235) for pooling


MikeInnes:
:+1: Thanks!

PR #38, opened 2018-04-19
Use let in @fix #37

### LINKED ISSUE #37: @fix and fused broadcasts
Continued from https://github.com/JuliaGPU/CuArrays.jl/issues/71

The issue
```
> x = CuArray([2.]);
> @fix x.*log.(x)
ERROR: Broadcast output type Any is not concrete
```

Using `CUDAnative.log` explicitly is fine:
```
> x.*CUDAnative.log.(x)
1-element CuArray{Float64,1}:
 1.38629
```

If I expand `@fix` I get
```
> @macroexpand @fix x.*log.(x)
quote 
    ##702 = (NNlib._cufunc)(log, (NNlib.cudata).((x,))...)
    x .* ##702.(x)
end
```
Which evaluates to
```
> f = CUDAnative.log;
> x.*f.(x)
ERROR: Broadcast output type Any is not concrete
```
But using only `f` works.
```
> f.(x)
1-element CuArray{Float64,1}:
 0.693147
```

It seems to me there is some issue with fused broadcasts and saving an intrinsic in a variable? Could we change the `@fix` macro to evaluate `(NNlib._cufunc)(log, (NNlib.cudata).((x,))...)` when running the macro, and return `:(x .* CUDAnative.log.(x))` instead?

gustafsson:
Allows @fix to be used in more complex expressions.

### DIFF
--- a/src/cubroadcast.jl
+++ b/src/cubroadcast.jl
@@ -25,5 +25,10 @@ macro fix(ex)
     push!(fs, :($f_ = $_cufunc($f, $cudata.(($(xs...),))...)))
     :($f_.($(args...)))
   end
-  :($(fs...); $ex) |> esc
+
+  quote
+    let $(fs...)
+      $ex
+    end
+  end |> esc
 end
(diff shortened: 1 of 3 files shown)


MikeInnes:
Great, thanks!

PR #40, opened 2018-05-09
Dilation support for Conv

tejank10:
Added dilation support for convolution layers (2d and 3d). @staticfloat


staticfloat:
Awesome!  I'm a little unfamiliar with where this all lives; do 1D convolutions live here as well, or is that a separate location?  (I ask because I want to do WaveNet-style convolutions, which are 1-d dilated convolutions on audio data)

tejank10:
Yes, 1D convolutions are here. Dilation is available for that as well.

MikeInnes:
Awesome. I'll let @staticfloat have the final word on this; if he's happy we can merge.

MikeInnes:
Thanks @tejank10!

Issue #41, opened 2018-05-10
Separable Convolutions


staticfloat:
Separable Convolutions (https://arxiv.org/abs/1610.02357) are really, really nice to have, and a must for fast/realtime image processing.  If nobody else does this, I will likely implement an im2col and view-based implementation of these later this summer, but I figured we should track this as they're rapidly becoming a mainstay in efficient (e.g. mobile) computer vision systems.

jekbradbury:
I think that is currently WIP here https://github.com/FluxML/model-zoo/pull/41 and should perhaps be moved to NNlib?

avik-pal:
The implementation in the pull request is quite slow and I am currently working on the im2col version of it. So I will try to get it done by this week possibly and send a PR as soon as possible if that is fine.

datnamer:
Can this be closed per #42 ?

MikeInnes:
Yup, would have thought so.

PR #42, opened 2018-05-17
Depthwise Convolutions

avik-pal:
Adds support for Depthwise Convolutions. #41

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -186,7 +186,7 @@ meanpool_cpu!(y::AbstractArray{<:Real,5}, x::AbstractArray{<:Real,5}, k::Dims{3}
 
 # Deprecated 0.3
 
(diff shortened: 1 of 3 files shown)


avik-pal:
@MikeInnes here are some [benchmarks](https://gist.github.com/avik-pal/4c75bd00565993985de86ffce36ee9ba). I will add the gradients and tests in furthur commits.

avik-pal:
@MikeInnes I have added all the test cases. It would be great if you could review it. As for the test failure, I guess it is due to segmentation fault in the conv2d test case.

MikeInnes:
Looks good to me. Maybe try rebasing; I think @tejank10's fixes should make the failures go away.

MikeInnes:
:+1: Thanks!

PR #45, opened 2018-05-23
Fixed NaN error in dilation

tejank10:
@staticfloat please review

### DIFF
--- a/src/impl/conv.jl
+++ b/src/impl/conv.jl
@@ -44,7 +44,7 @@ end
 
 function col2im_2d!{T}(col::AbstractArray{T,2}, img::AbstractArray{T,3}, width::Int, height::Int,
   channels::Int, kernel_w::Int, kernel_h::Int, pad_w::Int, pad_h::Int, stride_w::Int,
-  stride_h::Int, dil_h::Int, dil_w::Int, mode::Int)
+  stride_h::Int, dil_w::Int, dil_h::Int, mode::Int)
 
   height_col = div(height + 2pad_h - (kernel_h - 1) * dil_h - 1, stride_h) + 1
   width_col = div(width + 2pad_w - (kernel_w - 1) * dil_w - 1, stride_w) + 1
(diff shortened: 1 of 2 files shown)


staticfloat:
Confirmed that this suppresses the symptoms I was seeing.  Let's go ahead and add this workload to test suite (since the previous test didn't seem to catch this), remove the commented out `dilation_dims()` calls, and call it good.

As to _why_ this was happening, I think this was due to us passing in the incorrect dimensions to the `gemm()` calls, which caused OpenBLAS to read from the wrong memory locations.

staticfloat:
Uh oh, we need to fix gradients as well:

```julia
# 2-length convolution with 1 channel in and 1 channel out
w = reshape(Float64[1, 1], (2, 1, 1))

# 128-length input vector with 1 channel and 1 batch.
x = reshape(Float64[1:128...], (128, 1, 1))

# Do the convolution (this doesn't have any NaN's)
y = NNlib.conv(x, w; dilation = 2)
@test !any(isnan.(y))

# Generate random upstream derivative
dy = randn(size(y))

# Calculate backprop for w
dws = []
for idx in 1:1000
    dw = NNlib.∇conv_filter(dy, x, w; dilation=2)
    push!(dws, dw)
end

# Calculate backprop for x
dxs = []
for idx in 1:1000
    dx = NNlib.∇conv_data(dy, x, w; dilation=2)
    push!(dxs, dx)
end

@show any([any(isnan.(dw)) for dw in dws])
@show any([any(isnan.(dx)) for dx in dxs])
```

In my experiments, `dx` gets `NaN`'s sometimes, `dw` has not, but that could just be because my `w` is very small, so whatever memory region the BLAS routines are reaching into to get uninitialized memory might not be getting read from if `w` and `x` are too small.

tejank10:
`im2col_dims` was returning extra memory size. Also, I had goofed up the order of dimensions in `dilation` parameter of `col2im2d!`

MikeInnes:
Thanks! I'm getting segfaults on 0.7, so hopefully this will address those.

PR #46, opened 2018-05-29
Add hard-saturating (clipped) sigmoid and tanh activations

freeboson:
Added `hard_tanh`, `hardσ = hard_sigmoid`, and `hardσ_keras = hardσ_keras`. These are linearized versions of those functions. The Keras variant of `hardσ`, `hardσ_keras` is just to match `keras.activations.hard_sigmoid`.

### DIFF
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -2,8 +2,9 @@ module NNlib
 
 using Requires
 
-export σ, sigmoid, relu, leakyrelu, elu, swish, selu, softplus, softsign, logσ, logsigmoid,
-  softmax, logsoftmax, maxpool, meanpool
+export σ, sigmoid, hardσ, hard_sigmoid, hardσ_keras, hard_sigmoid_keras,
+  relu, leakyrelu, elu, swish, selu, softplus, softsign, logσ, logsigmoid,
+  softmax, logsoftmax, maxpool, meanpool, hard_tanh
 
 include("numeric.jl")
 include("activation.jl")
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -39,6 +39,34 @@ end
 
 const logsigmoid = logσ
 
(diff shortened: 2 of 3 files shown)


MikeInnes:
Neat – but does hard_sigmoid really need to be a macro?

freeboson:
How dare you?! :cry: 

I actually was making it a macro so that you could force the type to be the same as the cutoff, instead of using the type of the input. I think the only difference with using a macro now instead of just a closure is having constants for the cutoff and slope in the lowered code vs a `getfield`. This surprisingly makes slightly different native code! I'm happy to change it to:

```julia
function hard_sigmoid_gen(cutoff::Real)
    m = 1/(2*cutoff)
    x -> ifelse(x ≤ -cutoff, zero(x/1), ifelse(x ≥ cutoff, one(x/1), x*oftype(x/1, m) + oftype(x/1, 0.5)))
end
hard_sigmoid = hard_sigmoid_gen(2.0)
```

but is this really how you want to live your life?

freeboson:
What are you going to do with all your extra free time?

```julia
julia> x = randn(1000000);

julia> @btime hard_sigmoid.(x);
  1.601 ms (26 allocations: 7.63 MiB)

julia> @btime h.(x);
  1.610 ms (26 allocations: 7.63 MiB)
```

freeboson:
@MikeInnes do you want me to just make it a function?

staticfloat:
@freeboson I'm sorry that this has languished for so long; it seems to have fallen through the cracks.  If you are still interested in providing this, please recreate your changes on top of the latest `NNlib#master`, and redefine this as a function, instead of a macro.  To play nicely with types, you can use things like `oftype()` or even `typeof()` to get the type of an input parameter at runtime (the compiler will fold that away when it specializes the activation function on the input datatypes).

I'm going to close this, feel free to open a new PR if you're still interested.

PR #47, opened 2018-05-31
Julia 0.7

MikeInnes:
Bit unfortunate that we lose the CuArrays and ForwardDiff integration tests here. Solution might be a new repo for GPU integration tests across the Flux ecosystem.


MikeInnes:
@tejank10 the last deprecations come from the max pooling backward functions, which I think you wrote – could you take a look at those?

PR #53, opened 2018-07-05
Reflect mode parameter from NNlib

ayush1999:
Added `mode` parameter in the external functions (src/conv.jl), (option to choose cross-correlation or convolution), which were using `mode = 0` by default.
I will create a corresponding PR in Flux too, to adapt to the changes made in this PR.

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -24,10 +24,15 @@ padtuple(x::Tuple,p::Integer) = map(_->p, head(head(x)))
 padtuple(x::Tuple,p::Tuple) = p
 padtuple(x::AbstractArray,p) = padtuple(size(x),p)
 
-function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
+function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, flipkernel=true) where A<:AbstractArray
   pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+  if flipkernel
   conv!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)),
-        x, w, pad = pad_, stride = stride_, dilation = dilation)
+        x, w, pad = pad_, stride = stride_, dilation = dilation, flipkernel=flipkernel)
+  else
+  crosscor!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)),
+      x, w, pad = pad_, stride = stride_, dilation = dilation, flipkernel=flipkernel)
+  end
 end
 
 ∇conv_data(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray =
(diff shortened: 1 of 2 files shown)


MikeInnes:
I think we need to also expose this parameter from the `conv!` family of functions.

Can you also create a `crosscor!` function, similar to `conv!` that just calls `conv!` with `flipkernel`. I think we should document that rather than the flipkernel argument.

We'll need to implement this over both branches. If you target master, I can take care of merging into julia-0.6.

src/conv.jl
```diff
@@ -62,8 +68,12 @@ function ∇conv_data!(dx::AbstractArray{T,3}, dy::AbstractArray{T,3},
 end
 
 conv!(y::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
-      pad = 0, stride = 1, dilation = 1) where T =
-  conv2d!(y, x, w, padding = pad, stride = stride, dilation = dilation)
+      pad = 0, stride = 1, dilation = 1, flipkernel=true) where T =
+  conv2d!(y, x, w, padding = pad, stride = stride, dilation = dilation, flipkernel=flipkernel)
+
+crosscor!(y::AbstractArray{T,4}, x::AbstractArray{T,4}, w::AbstractArray{T,4};
```
MikeInnes: Cross correlation should have `flipkernel=false`, or at least be different from `conv!`, no? I also think it'd be better to keep this generic to `AbstractArray`s and just forward whatever the arguments are to `conv!`.

src/conv.jl
```diff
@@ -24,10 +24,15 @@ padtuple(x::Tuple,p::Integer) = map(_->p, head(head(x)))
...
 
-function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
+function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, flipkernel=true) where A<:AbstractArray
   pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+  if flipkernel
   conv!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)),
-        x, w, pad = pad_, stride = stride_, dilation = dilation)
+        x, w, pad = pad_, stride = stride_, dilation = dilation, flipkernel=flipkernel)
+  else
+  crosscor!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)),
```
MikeInnes: It doesn't make sense to do an `if/else` here; you pass `flipkernel` through so removing it would behave the same way.

It would however be good to have an out-of-place `crosscor` function that calls `crosscor!`.

src/conv.jl
```diff
@@ -91,7 +101,7 @@ function dcdims(x::NTuple{4,Int}, w::NTuple{4,Int}, pad, stride)
   ((x[1] + 2 * pad[1] - w[1])÷stride[1] + 1,(x[2] + 2 * pad[2] - w[2])÷stride[2] + 1,w[3]*w[4],x[4])
 end
 
-function depthwiseconv(x::A, w::A; pad = 0, stride = 1, flipkernel = true) where A<:AbstractArray
+function depthwiseconv(x::A, w::A; pad = 0, stride = 1, flipkernel = false) where A<:AbstractArray
```
MikeInnes: Why is `flipkernel` false by default for depthwise conv, but true for `conv`?

MikeInnes:
Bump -- would be good to have this in for #67. Perhaps @avik-pal can take a look at finishing it up.

ayush1999:
@avik-pal Ping me if you need this finished. (I'll probably open a new fresh PR)

staticfloat:
Superseded by #94

PR #54, opened 2018-07-07
2D Transpose Convolutions

tejank10:
Added 2D transpose convolutions and tests

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -42,7 +42,7 @@ function wdims(x::NTuple{N}, y::NTuple{N}, pad, stride, dilation) where N
     elseif i == N-1
       x[i]
     else # i == N
-      y[i]
+      y[i-1]
     end
   end
 end
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -39,12 +39,12 @@ padtuple(x::Tuple,p::Integer) = map(_->p, head(head(x)))
 padtuple(x::Tuple,p::Tuple) = p
 padtuple(x::AbstractArray,p) = padtuple(size(x),p)
 
-function conv(x::A, w::A, y_dims=nothing; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
+function conv(x::A, w::A; size=nothing, pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
   pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
-  if y_dims == nothing
-    y_dims = cdims(size(x), dilation_dims(w, dilation), pad_, stride_)
+  if size === nothing
+    size = cdims(Base.size(x), dilation_dims(w, dilation), pad_, stride_)
   end
-  conv!(similar(x, y_dims), x, w, pad = pad_, stride = stride_, dilation = dilation)
+  conv!(similar(x, size), x, w, pad = pad_, stride = stride_, dilation = dilation)
 end
 
 function crosscor(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
@@ -53,12 +53,12 @@ function crosscor(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:Abstra
         x, w, pad = pad_, stride = stride_, dilation = dilation)
 end
 
-function ∇conv_data(dy::A, w::A, x_dims=nothing; pad = 0, stride = 1, dilation = 1, flipkernel = 0) where A<:AbstractArray
+function ∇conv_data(dy::A, w::A; size=nothing, pad = 0, stride = 1, dilation = 1, flipkernel = 0) where A<:AbstractArray
   pad_, stride_, dilation_ = padtuple(dy, pad), padtuple(dy, stride), padtuple(dy, dilation)
-  if x_dims == nothing
-    x_dims = ctdims(size(dy), size(w), pad_, stride_, dilation_)
+  if size === nothing
+    size = ctdims(Base.size(dy), Base.size(w), pad_, stride_, dilation_)
   end
-  x = similar(dy, x_dims)    
+  x = similar(dy, size) 
   ∇conv_data!(zero(x), dy, x, w, pad = pad_, stride = stride_, dilation = dilation_, flipkernel=flipkernel)
 end
 
@@ -239,3 +239,14 @@ meanpool_cpu!(y::AbstractArray{<:Real,5}, x::AbstractArray{<:Real,5}, k::Dims{3}
               k::Dims{3}; pad = (0,0), stride = k) =
   meanpool3d_grad!(dx, dy, y, x,
                    window = k, padding = pad, stride = stride)
+
+# Deprecated
+
+export conv2d, maxpool2d, avgpool2d
+# 0.3
+@deprecate conv2d(x, w; kw...) NNlib.conv(x, w; kw...)
+@deprecate maxpool2d(x::AbstractArray{<:Real,4}, k::Integer) maxpool(x, (k,k))
+@deprecate meanpool2d(x::AbstractArray{<:Real,4}, k::Integer) meanpool(x, (k,k))
+
+# 0.4.2
+@deprecate ∇conv_data(dy::A, x::A, w::A; kw...) where A<:AbstractArray ∇conv_data(dy, w; size=size(x), kw...)
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -17,6 +17,21 @@ function cdims(x::NTuple{N}, w::NTuple{N}, pad, stride) where N
   end
 end
 
+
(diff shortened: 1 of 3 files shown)


tejank10:
That sounds cool. I was working on writing the gradient hook up for it. I felt that we should refactor the conv interface. 
Currently, `∇conv_data` and `conv` interface do not accept same number of arguements. `∇conv_data` takes the gradients (`dy`) as one of the inputs. So, `∇conv_data` ends up taking 3 arguements. I feel that this is more suited for ∇conv_data to act as gradient of `conv` rather than a forward pass of transposed convolution.

When this setup of  `∇conv_data` is used for transposed conv, it throws an error in the backward pass.
````
ERROR: Gradient is not a tuple of length 3
```
because the `back_` function requires the length of gradient tuple to be the same as the number of arguements to the `∇conv_data`.

tejank10:
This should fix the failing gradtest

tejank10:
This has been now resolved for v1.0.

MikeInnes:
It'd be good if we could take the opportunity to make the interface a bit more consistent. If I understand correctly, `∇conv_data` and `∇conv_filter` are currently taking an unnecessary extra argument, so it'd be better to just remove those everywhere, rather than adding the argument to `conv`. We will probably have to add deprecations for the current forms though.

tejank10:
I realized `∇conv_data` and `∇conv_filter` extra arguments are required. Using `cdims` or `ctdims` cannot give us exact dimensions every time because of integer division.
For eg: During convolution if `size(input) = (10,10,1,1)` and `size(kernel)=(3,3,1,1)` and `stride=2` then `size(output)=(4,4,1,1)`.
If `output` is passed through `∇conv_data` hoping it to use `ctdims` to infer the dimensions of `input` using dimensions of `output` and `kernel`, it would produce `(9,9,1,1)`

src/conv.jl
```diff
@@ -36,8 +53,14 @@ function crosscor(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:Abstra
         x, w, pad = pad_, stride = stride_, dilation = dilation)
 end
 
-∇conv_data(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1, flipkernel = 0) where A<:AbstractArray =
-  ∇conv_data!(zero(x), dy, x, w; pad = pad, stride = stride, dilation = dilation, flipkernel=flipkernel)
+function ∇conv_data(dy::A, w::A, x_dims=nothing; pad = 0, stride = 1, dilation = 1, flipkernel = 0) where A<:AbstractArray
```
MikeInnes: Couple small things
* Can you make this a keyword argument `size`
* Can you add a simple three-arg wrapper with a deprecation warning
* You should compare `x_dims === nothing` so that type inference can remove the check

tejank10: When `size` method is being used on `conv`, it gives `MethodError` if this argument is named `size`. How about using `dims` instead?

src/conv.jl
```diff
@@ -17,17 +17,34 @@ function cdims(x::NTuple{N}, w::NTuple{N}, pad, stride) where N
...
 
-function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
+function conv(x::A, w::A; size=nothing, pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
   pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
-  conv!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)),
-        x, w, pad = pad_, stride = stride_, dilation = dilation)
+  if size === nothing
+    size = cdims(Base.size(x), dilation_dims(w, dilation), pad_, stride_)
+  end
+  conv!(similar(x, size), x, w, pad = pad_, stride = stride_, dilation = dilation)
```
MikeInnes: Does the size argument actually make sense for `conv`? I don't know if there's a similar ambiguity in the sizes as compared with the transpose.

src/conv.jl
```diff
@@ -36,11 +53,17 @@ function crosscor(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:Abstra
...
+    size = ctdims(Base.size(dy), Base.size(w), pad_, stride_, dilation_)
+  end
+  dx = zeros(eltype(dy), size) 
+  ∇conv_data!(dx, dy, w, pad = pad_, stride = stride_, dilation = dilation_, flipkernel=flipkernel)
+end
 
-∇conv_filter(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1, flipkernel=0) where A<:AbstractArray =
-  ∇conv_filter!(zero(w), dy, x, w; pad = pad, stride = stride, dilation = dilation, flipkernel=flipkernel)
+∇conv_filter(dy::A, x::A, size::Tuple; pad = 0, stride = 1, dilation = 1, flipkernel=0) where A<:AbstractArray = 
+  ∇conv_filter!(zeros(eltype(dy),size), dy, x; pad = pad, stride = stride, dilation = dilation, flipkernel=flipkernel)
```
MikeInnes: This should use `similar`

src/conv.jl
```diff
@@ -36,11 +53,17 @@ function crosscor(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:Abstra
...
+  if size === nothing
+    size = ctdims(Base.size(dy), Base.size(w), pad_, stride_, dilation_)
+  end
+  dx = zeros(eltype(dy), size) 
+  ∇conv_data!(dx, dy, w, pad = pad_, stride = stride_, dilation = dilation_, flipkernel=flipkernel)
+end
 
-∇conv_filter(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1, flipkernel=0) where A<:AbstractArray =
-  ∇conv_filter!(zero(w), dy, x, w; pad = pad, stride = stride, dilation = dilation, flipkernel=flipkernel)
+∇conv_filter(dy::A, x::A, size::Tuple; pad = 0, stride = 1, dilation = 1, flipkernel=0) where A<:AbstractArray =
```
MikeInnes: Are you intentionally doing data and filter in different ways? Why not use a size kwarg for both?

src/conv.jl
tejank10: `conv` acts as the gradient function of input for conv transpose during the backward pass. Hence, just like `conv_data` there exists an ambiguity here as well.

src/conv.jl
tejank10: Yes, it is intentional. Because for `conv_data` it serves a dual purpose.When `size=nothing` it performs a `conv_transpose`, else it is a `conv_grad`. 
`conv_filter` has only one purpose to find the gradient of the filter. Hence, we require the value of `size` to be passed.

src/conv.jl
MikeInnes: Ok that's fine, just checking

src/conv.jl
MikeInnes: But they are not actually different operations, right? conv transpose is just the gradient with a particular inferred size. (Lmk if my understanding is off here.)

So we could in principle just write a size-inference for conv_filter; but if you don't want to do that for now it'd be fine to just do size as a kwarg without a value (which will error if it's not provided).

src/conv.jl
tejank10: To clarify, you mean to use `zero(similar(dy, size))`, right?

MikeInnes: Just similar should be fine, if you're about to overwrite it anyway.

tejank10:
`nan` test for `grad_conv_filter3d` is failing in this update, will fix it soon

src/conv.jl
tejank10: Currently we have `similar` in use in this branch, which is failing at `nan` test for `∇conv_filter`. However, if `zero` is used instead of `similar` then the tests pass.

vchuravy:
bump! this would be great to have

staticfloat:
I didn't see any big missing pieces, but I am worried that we might not have sufficient test coverage.  I am working on getting codecov or something hooked up so that we can be sure that what we're merging covers as many corner cases as possible.

staticfloat:
@tejank10 can you rebase this on top of the latest `master`?  I have merged Codecov support so we can make sure that our test cases are properly stressing each codepath of the convolutions now.

pbarker:
bump 🙂 is any help needed to get this out the door?

staticfloat:
@tejank10 Great, thanks.  Can you synthesize a few more tests to exercise the codepaths that are missing (as evidenced by the code coverage).

I'm particularly interested that we hit the first couple of branches in `ctdims()` and `wdims()` and that we hit the `size == nothing` branch in `crosscor()`, `∇conv_data()`, and `∇conv_filter()`.  If tests for those branches can be added, I think this is ready to merge.

PR #56, opened 2018-07-08
Faster ReLU, Softmax, Conv and Pool using NNPACK

americast:
Facilitates faster Softmax using [NNPACK](https://github.com/Maratyszcza/NNPACK). This requires the presence of `libnnpack.so`, a shared object file of NNPACK. The shared object file is to be placed manually at `/src`, but work is in progress to automatically take care of that using [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl).

Currently, all tests are passing locally.


src/NNPACK.jl
```diff
@@ -0,0 +1,31 @@
...
+function softmax!(out::AbstractVecOrMat{T}, xs::AbstractVecOrMat{T}) where T<:AbstractFloat
+  input = Cfloat.(xs)
+  output = similar(input)
+  ccall((:nnp_softmax_output,"libnnpack"),Void,(Csize_t, Csize_t, Ptr{Cfloat}, Ptr{Cfloat}, Ptr{Void}), Csize_t(size(xs, 2)), Csize_t(size(xs, 1)), input, output, ptp)
+  out .= output
+  return out
+end
+
+function relu(x)
+  if typeof(x) <: ForwardDiff.Dual || typeof(x) <: AbstractFloat
```
MikeInnes: This is what dispatch is for

src/NNPACK.jl
```diff
@@ -0,0 +1,31 @@
+ccall((:nnp_initialize,"libnnpack"),Void,(),)
+ptp = ccall((:pthreadpool_create, :libnnpack), Ptr{Void}, (Csize_t,), 0)
+
+function softmax!(out::AbstractVecOrMat{T}, xs::AbstractVecOrMat{T}) where T<:AbstractFloat
```
MikeInnes: Make this more direct – `VecOrMat{Float32}`. I doubt converting to and from is worth it, but if it turns out to be we can use a different overload for that.

src/NNPACK.jl
```diff
@@ -0,0 +1,31 @@
+ccall((:nnp_initialize,"libnnpack"),Void,(),)
```
MikeInnes: You should do this inside an `__init__` function so it's safe to precompile etc

src/NNlib.jl
```diff
@@ -5,6 +5,8 @@ using Requires
 export σ, sigmoid, relu, leakyrelu, elu, swish, selu, softplus, softsign, logσ, logsigmoid,
   softmax, logsoftmax, maxpool, meanpool
 
+nnpack_available=(is_linux() || is_apple()) && isfile("libnnpack.so")
```
MikeInnes: Just inline this variable if it's only used once. Also, this should look in the `deps` directory

americast:
Added `conv`. For now, supports only `Float32`, with `mode` set as 1. Support for stride and dilation can be brought about.

PR #57, opened 2018-07-09
Fix Depthwise Convolutions

avik-pal:
I have updated the gradient definition for the filter. The tests were not able to catch this case. So I have updated the tests to handle all the cases.


MikeInnes:
:+1: Thanks!

PR #58, opened 2018-07-30
Pull request/0219b3ae

pevnak:


pevnak:
I wanted to fix softplus, but totally messed up things. Will try to do it one more time.

PR #59, opened 2018-07-30
softplus

pevnak:
There is an overflow in softplus, so I have fixed it. I might done the pull-request incorrectly. But if someone (Mike) can cut&paste, including test, it would be excellent.


MikeInnes:
Easy enough to trim away the bad commits. Thanks @pevnak!

Issue #39, opened 2018-04-25
conv with SubArray


jbrea:
Is there are reason to force [x and w have the same type in conv](https://github.com/FluxML/NNlib.jl/blob/master/src/conv.jl#L27)? I would like to use conv with x being a view of another array, i.e. a SubArray. One solution could be to remove the type constraint for x on conv, conv!, ∇conv_filter!, conv2d!, conv2d_grad_w!, ∇conv_filter etc.

MikeInnes:
Just historical accident as this was ported from C code. If you want to remove the restrictions and make sure things still work OK, you're more than welcome to.

PR #63, opened 2018-09-05
allow type of x and w to differ in conv

jbrea:
Trying to address #39

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -24,16 +24,16 @@ padtuple(x::Tuple,p::Integer) = map(_->p, head(head(x)))
 padtuple(x::Tuple,p::Tuple) = p
 padtuple(x::AbstractArray,p) = padtuple(size(x),p)
 
-function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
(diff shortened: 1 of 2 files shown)


MikeInnes:
Thanks! Given that the type parameters are now not used in most cases, these can be written as `A::AbstractArray, B::AbstractArray`; hope you don't mind making that quick change.

jbrea:
I simplified some instances, as you suggested. Do you prefer to do this replacement everywhere?
`maxpool!(y::A, x::A, k; kw...) where A<:AbstractArray` looks nicely succinct.

JobJob:
IMO makes sense to make the same change everywhere relevant in conv.jl. Restricting args to be the exact same type sounds like it was just a historical accident, and `similar` for many AbstractArray types returns a different type (often a straight up Array) - which is relevant e.g. here: https://github.com/FluxML/NNlib.jl/blob/085adb7ca8ad57a11515d0396b6b120ab929216f/src/conv.jl#L155-L157

MikeInnes:
Agreed. I'll merge this once tests pass, and I'm happy to take more patches to clean this up.

PR #65, opened 2018-09-15
changed calls in conv.jl to make the convolutional neural net routine…

ityonemo:
…s type-blind.


staticfloat:
Superseded by #94.

Thanks for your contribution, @ityonemo!

PR #67, opened 2018-09-19
Add NNPACK support

avik-pal:
1. Only supports Linux platforms for now.
2. ~~Convolutions are not functional~~
3. ~~BenchmarkTools freezes when trying to benchmark with `NNPACK_CPU_THREADS` > 1~~

Speed Improvements:
```julia
x = rand(100,100,10,100)

#NNlib maxpool
@time maxpool(x,(2,2)) # 0.32 s

x = Float32.(x)
#NNPACK maxpool with NNPACK_CPU_THREADS = 4
@time maxpool(x,(2,2)) # 0.0055 s
```

~~NOTE: Travis test fails are due to unsupported hardware~~


avik-pal:
@MikeInnes 
Convolution is fixed. But the results are for cross correlation. So #53 needs to be merged.
Also I think we should change the default to cross correlation to maintain consistency

MikeInnes:
For now can we just flip the weight kernel? Should be quite cheap to do.

On (1) can we add a check that throws an error?

On (2) perhaps we can test NNlib on one of the GPU CI machines. Ideally we should not load NNPACK support if the hardware is not supported though.

avik-pal:
I will add a check for that.
Is there an easy way to check the hardware, I mean similar to `is_linux()` in case of os?

MikeInnes:
nnpack should have such a check itself, I think. You can load the library but query that before defining the actual `conv` wrappers.

avik-pal:
Is the check added in the previous commit fine?

avik-pal:
I have flipped the weights for now so that it gives the results for convolutions

deps/build.jl
```diff
@@ -0,0 +1,36 @@
+using BinaryProvider # requires BinaryProvider 0.3.0 or later
```
MikeInnes: Shouldn't this be expressed by version bounds in REQUIRE, rather than as a comment?

src/nnpack/libnnpack.jl
```diff
@@ -0,0 +1,143 @@
...
+    @check ccall((:nnp_softmax_output, libnnpack), nnp_status, (Csize_t, Csize_t, Ptr{Cfloat}, Ptr{Cfloat}, pthreadpool_t), batch_size, channels, input, output, threadpool)
+end
+
+function nnp_softmax_output(x::AbstractVecOrMat{Float32}, y::AbstractVecOrMat{Float32}; threadpool = nothing)
+    threadpool = threadpool === nothing ? pthreadpool_create() : threadpool
+    nnp_softmax_output(ndims(x) == 2 ? size(x, 2) : 1, size(x, 1), x, y, threadpool)
+    y
+end
+
+#FIXME: Output of fully connected not consistent with `kernel * input`
```
MikeInnes: At a guess this might be because NNPACK is row major

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,71 @@
...
+relu(x::AA1; inplace::Bool = true, nthreads = NNPACK_CPU_THREADS) =
+    nnp_relu_output(x, inplace ? x : similar(x), threadpool = pthreadpool_create(nthreads))
+
+leakyrelu(x::AA1, a = oftype(x/1, 0.01); inplace::Bool = true, nthreads = NNPACK_CPU_THREADS) =
+    nnp_relu_output(x, inplace ? x : similar(x), negative_slope = a, threadpool = pthreadpool_create(nthreads))
+
+# NOTE: softmax is slower than the nnlib version. So might comment it out
+
+softmax!(x::AbstractVecOrMat{Float32}; inplace::Bool = true, nthreads = NNPACK_CPU_THREADS) =
+    nnp_softmax_output(x, inplace ? x : similar(x), threadpool = pthreadpool_create(nthreads))
```
MikeInnes: Wouldn't it be better/faster to create a shared threadpool at init?

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,71 @@
+const AA{N} = AbstractArray{Float32,N}
+const AA1 = Union{AA{2}, AA{3}, AA{4}, AA{5}}
+
+relu(x::AA1; inplace::Bool = true, nthreads = NNPACK_CPU_THREADS) =
```
MikeInnes: I don't think we should overload NNlib's scalar activation functions like this.

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,26 @@
...
+if !isfile(depsjl_path)
+    error("NNPACK not installed properly, run Pkg.build(\"NNlib\"), restart Julia and try again")
+end
+include(depsjl_path)
+
+const nnlib_interface_path = joinpath(dirname(@__FILE__), "nnlib.jl")
+@init begin
+    check_deps()
+    try
+        nnp_initialize()
```
MikeInnes: Can we explicitly check the [status code](https://github.com/Maratyszcza/NNPACK/blob/1e005b0c2777f39972a4ac15bea03e0e315a3d92/include/nnpack.h#L80) rather than having a generic try/catch; otherwise load errors when including the wrappers will get hidden.

src/nnpack/nnlib.jl
avik-pal: So should I leave out the activation functions?

avik-pal:
Will get these fixed soon

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,71 @@
...
+end
+
+maxpool!(y::AA{4}, x::AA{4}, k; pad = map(_->0,k), stride = k, threadpool = nothing) =
+    nnp_max_pooling_output(x, y, k, padding = expand(Val{length(k)}, pad), stride = expand(Val{length(k)}, stride), threadpool = threadpool)
+
+function conv(x::AA{4}, w::AA{4}; pad = 0, stride = 1, dilation = 1, algo = 0, nthreads = NNPACK_CPU_THREADS)
+    dilation == 1 || dilation == (1, 1) || error("NNPACK does not support dilation > 1")
+    pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+    ((size(x, 1) - size(w, 1) + 2 * pad_[1]) % stride_[1] == 0 && (size(x, 2) - size(w, 2) + 2 * pad_[2]) % stride_[2] == 0) || error("Choose the stride, pad and kernel size properly")
+    w = w[end:-1:1, end:-1:1, :, :] # Needs to be fixed once the flipkernel and crosscor PR is merged
```
MikeInnes: Just a small thing, but FYI you can use `reverse` for this

avik-pal:
Why can't travis get `BinaryProvider` even though I have listed it in the `REQUIRE` file?

MikeInnes:
Not sure. @staticfloat any ideas?

staticfloat:
It's because you have a `Project.toml` and `Manifest.toml`; if you want to keep those, then you need to add `BinaryProvider` to those by running `julia --project` within the source directory and doing a `pkg> add BinaryProvider`.  Otherwise, you could just delete them for now and it will fallback to the `REQUIRE` file.

MikeInnes:
I though that at first but the project file is only on master, not on this branch – unless that still affects things somehow?

staticfloat:
Travis is building the result of merging this branch into the target branch.  You can tell from [this little bit](https://travis-ci.org/FluxML/NNlib.jl/jobs/441542888#L403-L411):

```
$ git fetch origin +refs/pull/67/merge:
```

staticfloat:
Avik, I'm impressed at how "clean" this is.

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,27 @@
...
+    status = nnp_initialize()
+    if status == nnp_status_unsupported_hardware
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["JULIA_NUM_THREADS"])
+    catch
+        @warn "`JULIA_NUM_THREADS` not set. So taking the NNPACK default `4`"
```
SimonDanisch: You could use `Sys.CPU_THREADS` to get a better default - if that's intended!

avik-pal: Thanks, did not know about it. Its definitely a more reasonable default

avik-pal:
Actually NNPACK works best if we choose the num threads in the range 4-8. Choosing anything far out of this range tremendously slows down convolutions. Using Sys.CPU_THREADS can give a value well outside this range for some cases(tested on antarctic).

pshashk:
It is sometimes useful to force some softmax outputs to be zero. For example, in sequence processing architectures like the transformer in order to avoid model attending to future information, some logits can be set to -Inf before passing to softmax. However, this masking pattern doesn't work in NNPACK.

master
```
x = [1,2,3,-Inf32]

julia> softmax(x)
4-element Array{Float32,1}:
 0.09003057
 0.24472848
 0.66524094
 0.0 
```

this branch
```
julia> softmax(x)
4-element Array{Float32,1}:
 NaN
 NaN
 NaN
 NaN
```

Will there be any way to switch between old and new version after this PR is merged? Or perhaps there is a better way to implement softmax masking without infinities?

pshashk:
Actually, "diagonal" softmax masking in the transformer architecture is not a very good example, because it can be efficiently achieved by: `UpperTriangular(exp.(logits) ./ cumsum(exp.(logits), dims = 1)')`. But still there are some applications in classification and ranking scenarios where external information is provided to avoid backprop over some subsets of sotfmax outputs.

avik-pal:
>Will there be any way to switch between old and new version after this PR is merged?

Even I was thinking of some way to do this. This will be needed especially in cases where the `kernel`, `padding` and `stride` do not add up properly. Since NNPACK doesnot support this I simply added an error statement for now. But some of the pretrained models in Metalhead like the Squeezenet model will end up throwing an error if we use the NNPACK functions.

>Or perhaps there is a better way to implement softmax masking without infinities?

Your example in the next comment should solve this. But the way to do this using NNPACK would be to set `x = [1,2,3,-1e100]`
The output would be:
```julia
4-element Array{Float64,1}:
0.09003057317038046
0.24472847105479764
0.6652409557748218 
0.0
```

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,86 @@
...
+#     threadpool = nthreads != NNPACK_CPU_THREADS ? pthreadpool_create(nthreads) : shared_threadpool
+#     nnp_relu_output(x, inplace ? x : similar(x), threadpool = threadpool)
+# end
+#
+# function leakyrelu(x::AA1, a = oftype(x/1, 0.01); inplace::Bool = true, nthreads = NNPACK_CPU_THREADS)
+#     threadpool = nthreads != NNPACK_CPU_THREADS ? pthreadpool_create(nthreads) : shared_threadpool
+#     nnp_relu_output(x, inplace ? x : similar(x), negative_slope = a, threadpool = threadpool)
+# end
+
+function softmax!(x::AbstractVecOrMat{Float32}; inplace::Bool = true, nthreads = NNPACK_CPU_THREADS)
```
MikeInnes: The `inplace` argument should be replaced with `softmax` and `softmax!`, like the original API

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,27 @@
...
+    if status == nnp_status_unsupported_hardware
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["JULIA_NUM_THREADS"])
+    catch
+        global NNPACK_CPU_THREADS = 4
+        @warn "JULIA_NUM_THREADS not set. So taking the NNPACK default $(NNPACK_CPU_THREADS)"
```
MikeInnes: I don't think we need to have this warning by default (maybe we can just set it to debug level or something, so it's not visible by default).

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,27 @@
...
+    end
+    global shared_threadpool = pthreadpool_create(NNPACK_CPU_THREADS)
```
MikeInnes: This is good but will be slow to access, can we store it in a const ref or something.

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,86 @@
...
+    b = zeros(Float32, size(y, 3))
+    conv!(y, x, w, b, pad = pad_, stride = stride_, dilation = dilation, algo = UInt32(algo), threadpool = threadpool)
+end
+
+function conv(x::AA{4}, w::AA{4}, b::AA{1}; pad = 0, stride = 1, dilation = 1, algo = UInt32(0), nthreads = NNPACK_CPU_THREADS)
+    threadpool = nthreads != NNPACK_CPU_THREADS ? pthreadpool_create(nthreads) : shared_threadpool
+    dilation == 1 || dilation == (1, 1) || error("NNPACK does not support dilation > 1")
+    pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+    ((size(x, 1) - size(w, 1) + 2 * pad_[1]) % stride_[1] == 0 && (size(x, 2) - size(w, 2) + 2 * pad_[2]) % stride_[2] == 0) || error("Choose the stride, pad and kernel size properly")
+    w = reverse(reverse(w, dims=1), dims=2) # Needs to be fixed once the flipkernel and crosscor PR is merged
```
MikeInnes: Should be possible to combine this into one with:
```suggestion
    w = reverse(w, dims=(1,2)) # Needs to be fixed once the flipkernel and crosscor PR is merged
```

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,86 @@
...
+    y = similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_))
+    b = zeros(Float32, size(y, 3))
+    conv!(y, x, w, b, pad = pad_, stride = stride_, dilation = dilation, algo = UInt32(algo), threadpool = threadpool)
+end
+
+function conv(x::AA{4}, w::AA{4}, b::AA{1}; pad = 0, stride = 1, dilation = 1, algo = UInt32(0), nthreads = NNPACK_CPU_THREADS)
+    threadpool = nthreads != NNPACK_CPU_THREADS ? pthreadpool_create(nthreads) : shared_threadpool
```
MikeInnes: I think we should just avoid exposing `nthreads` for now, and always use the shared pool. If it's something people want we can add it in later.

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,86 @@
+const AA{N} = AbstractArray{Float32,N}
```
MikeInnes: I think that we should catch Float64 here as well, at least for convs and possibly also for softmax and pooling (depending on benchmark results). Doing the conversion to Float32 and feeding to nnpack is probably quicker than  using our slow convolutions.

src/nnpack/nnlib.jl
avik-pal: `reverse` function is expecting an integer as dims. Its not taking a tuple

src/nnpack/nnlib.jl
avik-pal: This will definitely make things faster but we also end up losing features like dilation, also arbitrary stride kernel combinations will give errors. Also if we are internally making all computations in Float32 can't we simply make the conversion in Flux, we would just need one conversion in that case instead of repeated conversion back and forth from Float64 -> Float32

avik-pal:
Most of the test failures are due to accuracy issues I believe.

src/nnpack/nnlib.jl
staticfloat: If we're calling into `NNPACK`, I think it makes sense to "poison" the chain of operations; e.g. if we've configured Flux to be dead-set on using NNPACK, then `conv(X<:Float64, W<:Float32)` should return a `Float32`; narrowing the output type due to the internal kernel working in f32.  This should cause downstream operations to use `conv(X<:Float32, W<:Float32)`, thereby avoiding the constant re-quantization.

If we just document this, and allow people who really care about precision to trivially re-cast to `Float64` in the places that it matters, then I think we should be able to get the best of both worlds.

As for dilation and arbitrary stride kernels, I think those should fall back to the slow convolution algorithms.

staticfloat:
@avik-pal I am trying this branch out now because I'm trying to train large vision models (like VGG19).  I fixed all the precision problems in the test set on my local machine by using the `≈ Float32.(...)` trick in more places.  There are two classes of errors that remain:

* Lack of dilation support: `NNPACK does not support dilation > 1` at `test/conv.jl:36` and `test/conv.jl:45`.

* `Choose the stride, pad and kernel size properly` errors at `test/conv.jl:24`, `test/conv.jl:138`, `test/conv.jl:139`, `test/conv.jl:140` and `test/conv.jl:150`.

As stated above in a code comment, I think it makes sense to fall back to the slow algorithms in these cases.  I don't know how easy it is to dispatch that, but I think it important that we continue to allow dilated and strided convolutions.

staticfloat:
This also gives an ambiguity error when used from Flux:

```
ERROR: MethodError: getfield(NNlib, Symbol("#kw##conv"))()(::NamedTuple{(:stride, :pad, :dilation),Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64},Tuple{Int
64,Int64}}}, ::typeof(NNlib.conv), ::Array{Float64,4}, ::TrackedArray{…,Array{Float64,4}}) is ambiguous. Candidates:
  (::getfield(NNlib, Symbol("#kw##conv")))(::Any, ::typeof(NNlib.conv), x::AbstractArray, w::TrackedArray) in Flux.Tracker
  (::getfield(NNlib, Symbol("#kw##conv")))(::Any, ::typeof(NNlib.conv), x::AbstractArray{Float64,4}, w::AbstractArray{Float64,4}) in NNlib
```

Minimal reproducer:
```
using Flux
x = randn(10,10,1,1)
C = Conv((3, 3), 1 => 1)
C(x)
```

src/nnpack/nnlib.jl
avik-pal: The idea to document it is fair enough.
However, I am not yet sure how to call those functions for dilation, now that they are overloaded.

avik-pal:
@staticfloat It should be fixed now.

src/nnpack/nnlib.jl
```diff
@@ -8,40 +8,40 @@ const AA1 = Union{AA{2}, AA{3}, AA{4}, AA{5}}
...
+maxpool!(y::A, x::A, k; pad = map(_->0,k), stride = k) where A<:AA{4} =
     nnp_max_pooling_output(x, y, k, padding = expand(Val{length(k)}, pad), stride = expand(Val{length(k)}, stride), threadpool = shared_threadpool)
 
-conv(x::AbstractArray{Float64, 4}, w::AbstractArray{Float64, 4}; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) =
-    Float64.(conv(Float32.(x), Float32.(w), pad = pad, stride = stride, dilation = dilation, algo = algo))
+conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where A<:AbstractArray{Float64, 4} =
+    conv(Float32.(x), Float32.(w), pad = pad, stride = stride, dilation = dilation, algo = algo)
 
-function conv(x::AA{4}, w::AA{4}; pad = 0, stride = 1, dilation = 1, algo = UInt32(0))
+function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where A<:AA{4}
```
staticfloat: You have to be careful with these kinds of changes because they are not semantically the same; this `conv()` method now only takes `conv(::AA{Float32}, ::AA{Float32})` or `conv(::AA{Float64}, ::AA{Float64})` but not `conv(::AA{Float32}, ::AA{Float64})`.  Is that what you intended?

avik-pal: Yes, since the support for different types in conv hasn't been merged yet IIRC, I was avoiding making such a change in this PR. I believe it would be confusing to have different behavior in different places. Also the AA is short for AbstractArray{Float32}, I guess I should change it.

src/nnpack/nnlib.jl
MikeInnes: Perhaps you can use `invoke`?

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,26 @@
...
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["NNPACK_CPU_THREADS"])
+    catch
+        global NNPACK_CPU_THREADS = 4
+    end
+    global shared_threadpool = Ref(pthreadpool_create(NNPACK_CPU_THREADS), 1)
```
MikeInnes: I think you need to create the Ref as a global const and then fill it in `@init`, so that it can be type inferred.

src/nnpack/libnnpack.jl
```diff
@@ -0,0 +1,145 @@
...
+    threadpool = threadpool === nothing ? pthreadpool_create() : threadpool
+    nnp_fully_connected_inference(size(x, 1), size(y, 1), x, w, y, threadpool)
+    y
+end
+
+function nnp_max_pooling_output(batch_size, channels, input_size, input_padding, pooling_size, pooling_stride, input, output, threadpool)
+    @check ccall((:nnp_max_pooling_output, libnnpack), nnp_status, (Csize_t, Csize_t, nnp_size, nnp_padding, nnp_size, nnp_size, Ptr{Cfloat}, Ptr{Cfloat}, pthreadpool_t), batch_size, channels, input_size, input_padding, pooling_size, pooling_stride, input, output, threadpool)
+end
+
+function nnp_max_pooling_output(x::AbstractArray{Float32,4}, y::AbstractArray{Float32,4}, kernel::Tuple; padding = 0, stride = 1, threadpool = nothing)
```
MikeInnes: Shouldn't these functions all use the shared threadpool now?
```suggestion
function nnp_max_pooling_output(x::AbstractArray{Float32,4}, y::AbstractArray{Float32,4}, kernel::Tuple; padding = 0, stride = 1, threadpool = shared_threadpool[])
```

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,119 @@
...
+    nnp_softmax_output(x, x, threadpool = shared_threadpool)
+
+softmax!(y::A, x::A) where A<:AbstractVecOrMat{Float64} = softmax!(Float32.(y), Float32.(x))
+
+softmax!(y::A, x::A) where A<:AbstractVecOrMat{Float32} =
+    nnp_softmax_output(x, y, threadpool = shared_threadpool)
+
+softmax(x::A) where A<:AbstractVecOrMat{Float64} = softmax(Float32.(x))
+
+softmax(x::A) where A<:AbstractVecOrMat{Float32} =
```
MikeInnes: Doesn't NNlib effectively already have this definition?

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,119 @@
...
+maxpool!(y::A, x::A, k; pad = map(_->0,k), stride = k) where A<:AbstractArray{Float32, 4} =
+    nnp_max_pooling_output(x, y, k, padding = expand(Val{length(k)}, pad), stride = expand(Val{length(k)}, stride), threadpool = shared_threadpool)
+
+conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where A<:AbstractArray{Float64, 4} =
+    conv(Float32.(x), Float32.(w), pad = pad, stride = stride, dilation = dilation, algo = algo)
+
+function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where A<:AbstractArray{Float32, 4}
+    dilation == 1 || dilation == (1, 1) || error("NNPACK does not support dilation > 1")
+    pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+    ((size(x, 1) - size(w, 1) + 2 * pad_[1]) % stride_[1] == 0 && (size(x, 2) - size(w, 2) + 2 * pad_[2]) % stride_[2] == 0) || error("Choose the stride, pad and kernel size properly")
```
MikeInnes: Seems like there might be some shared code among these checks, perhaps they can be factored out into a function?

src/nnpack/nnlib.jl
```diff
@@ -0,0 +1,119 @@
...
+    conv(Float32.(x), Float32.(w), Float32.(b), pad = pad, stride = stride, dilation = dilation, algo = algo)
+
+function conv(x::A1, w::A1, b::A2; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where {A1<:AbstractArray{Float32, 4}, A2<:AbstractArray{Float32, 1}}
+    dilation == 1 || dilation == (1, 1) || error("NNPACK does not support dilation > 1")
+    pad_, stride_ = padtuple(x, pad), padtuple(x, stride)
+    ((size(x, 1) - size(w, 1) + 2 * pad_[1]) % stride_[1] == 0 && (size(x, 2) - size(w, 2) + 2 * pad_[2]) % stride_[2] == 0) || error("Choose the stride, pad and kernel size properly")
+    conv!(similar(x, cdims(size(x), dilation_dims(w, dilation), pad_, stride_)), x, w, b, pad = pad_, stride = stride_, dilation = dilation, algo = UInt32(algo))
+end
+
+crosscor(x::A1, w::A1, b::A2; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where {A1<:AbstractArray{Float64, 4}, A2<:AbstractArray{Float64, 1}} =
```
MikeInnes: Would it be possible to implement `conv!` with support for `flipkernel`, and then avoid implementing these methods explicitly? (i.e. NNlibs current fallbacks should end up doing the same thing)

avik-pal: Actually the NNPACK needs a bias term. The NNlib definitions do not take the bias term into consideration so I added these functions for consistency. As for the normal ones without bias it already uses the ones in NNlib.

src/nnpack/nnlib.jl
avik-pal: Yes it kind of does. But we do call into NNPACK softmax for all cases.

src/nnpack/NNPACK.jl
avik-pal: I have made some changes. Is this what you mean?

avik-pal:
Any particular reason we are making travis test on Julia 0.7? The tests pass locally for julia 1.0

src/nnpack/interface.jl
```diff
@@ -0,0 +1,143 @@
...
+end
+
+#NOTE: Commenting out the activation functions until sure what to do
+
+# relu(x::AA1) = nnp_relu_output(x, inplace ? x : similar(x), threadpool = shared_threadpool[])
+
+# leakyrelu(x::AA1, a = oftype(x/1, 0.01)) =
+#     nnp_relu_output(x, inplace ? x : similar(x), negative_slope = a, threadpool = shared_threadpool[])
+
+softmax!(x::A) where A<:AbstractVecOrMat{Float64} = softmax!(Float32.(x))
```
CarloLucibello: this seems very wrong for 3 reasons:
1) we are not modifying x as `softmax!(x)` is supposed to do; 
2) If I choose to use Float64 arrays you have to respect that, I don't want things
to be rounded to 32 bits internally without me (the user) knowing it. 
3) also, the return type is changing to float32 array

src/nnpack/interface.jl
```diff
@@ -0,0 +1,143 @@
...
+
+softmax!(x::A) where A<:AbstractVecOrMat{Float32} =
+    nnp_softmax_output(x, x, threadpool = shared_threadpool[])
+
+softmax!(y::A, x::A) where A<:AbstractVecOrMat{Float64} = softmax!(Float32.(y), Float32.(x))
+
+softmax!(y::A, x::A) where A<:AbstractVecOrMat{Float32} =
+    nnp_softmax_output(x, y, threadpool = shared_threadpool[])
+
+softmax(x::A) where A<:AbstractVecOrMat{Float64} = softmax(Float32.(x))
```
CarloLucibello: this is very wrong, see my comment above

src/nnpack/interface.jl
avik-pal: There was a prior discussion on the internal conversion issue.

Here's @staticfloat 's comment on the issue

> If we're calling into NNPACK, I think it makes sense to "poison" the chain of operations; e.g. if we've configured Flux to be dead-set on using NNPACK, then conv(X<:Float64, W<:Float32) should return a Float32; narrowing the output type due to the internal kernel working in f32. This should cause downstream operations to use conv(X<:Float32, W<:Float32), thereby avoiding the constant re-quantization.
If we just document this, and allow people who really care about precision to trivially re-cast to Float64 in the places that it matters, then I think we should be able to get the best of both worlds.

However, we are currently making any `Float64` conversion to `Float32` to proceed. This speeds up the training by a factor of 4x. So I believe its worth making the conversion and mentioning about it in the documentation.

Also for the first point, `softmax!` defined for `Float32` will be making the inplace modification of x. And we are calling that from the function for `Float64`. Were you refering to this?

CarloLucibello: >If we're calling into NNPACK, I think it makes sense to "poison" the chain of operations; e.g. if we've configured Flux to be dead-set on using NNPACK, then conv(X<:Float64, W<:Float32) should return a Float32; narrowing the output type due to the internal kernel working in f32. This should cause downstream operations to use conv(X<:Float32, W<:Float32), thereby avoiding the constant re-quantization.
If we just document this, and allow people who really care about precision to trivially re-cast to Float64 in the places that it matters, then I think we should be able to get the best of both worlds.

I disagree with that, we should follow julia promotion conventions and have
`conv(X<:Float64, W<:Float32) -> Float64`. 

If one wants to work only in float32, he would simply have to cast the input X to float32 and every operation in the successive layers will be float32.

On the contrary, having `conv(X<:Float64, W<:Float32) -> Float32`, beside being surprising to the user, would mean that if I want the "work with 64 bit activation" behaviour, I would have to cast
to float64 at *each* hidden layer. 

And here we are discussing a Float64 x Float32 -> Float32 operation, a Float64 -> Float32 operation as the proposed `softmax` makes even less sense. If NNPACK doesn't support float64 operations (is this the case?) we have to fall back to the julia implementation for float64 ones. 

We should not enforce something on the users, they should be in power, if the want speed they should use float32, but if they precision it has to possible for them to stick to float64.

>Also for the first point, softmax! defined for Float32 will be making the inplace modification of x. And we are calling that from the function for Float64. Were you refering to this?

Problem is that when you call `Float32.(x)` you are creating a new array, so `x` is not modified at all.

staticfloat: NNPACK does not support float64 operations, as far as I know, and this PR was (in my understanding) code that only gets enabled when Flux is in an NNPACK-specific mode, as stated in the block of text you quoted above.  If that's not the case then I agree we need to change something, but if Flux is setup to use NNPACK as its backing execution engine, this behavior makes sense.

avik-pal: Currently, NNPACK gets called by default as long as the hardware supports it. But we can simply add a flag which can make us use the NNlib default functions instead of NNPACK. I believe if the user is allowing NNPACK to load then there should not be any issue if we internally do computation in Float32 because NNPACK cant handle Float64.

Also `conv(X<:Float64, W<:Float32)` will still error in this PR. The datatypes of the arrays must match for the convolution to take place. I could try to make the changes here itself. But it also involves making a lot of changes in base nnlib functions. Hence if you do `conv(X<:Float64, W<:Float64)` we get `Float32` outputs in case you enable NNPACK.

If accurary is an issue and still user wants the speed of NNPACK, we might want the user to pass another flag which loads the NNPACK `Float32` functions but restricts the usage of NNPACK for `Float64`.

> Problem is that when you call Float32.(x) you are creating a new array, so x is not modified at all.

Yes, you are right. I will correct this.

avik-pal:
I have updated the travis.yml script to test on Julia 1.0 . Let me know if there was any particular reason to test on Julia0.7

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,28 @@
...
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["NNPACK_CPU_THREADS"])
+    catch
+        global NNPACK_CPU_THREADS = 4
+    end
+    global shared_threadpool = Ref(pthreadpool_create(NNPACK_CPU_THREADS))
```
MikeInnes: What I mean is, make `shared_threadpool` const and do `shared_threadpool[] = pthreadpool_create(...)`. Mutable globals are slow, but a ref can be type-inferred.

src/nnpack/interface.jl
```diff
@@ -0,0 +1,139 @@
...
+    return pad_, stride_, fallback
+end
+
+function softmax!(x::A) where A<:AbstractVecOrMat{Float64}
+    x = Float32.(x)
+    softmax!(x)
+end
+
+softmax!(x::A) where A<:AbstractVecOrMat{Float32} =
+    nnp_softmax_output(x, x, threadpool = shared_threadpool[])
```
MikeInnes: These kwargs seem unecessary, since `threadpool` defaults to `shared_threadpool` anyway

src/nnpack/interface.jl
```diff
@@ -0,0 +1,139 @@
...
+        maxpool_cpu!(similar(x, pdims(size(x), k, pad_, stride_)), x, k, pad = pad_, stride = stride_)
+    else
+        maxpool!(similar(x, pdims(size(x), k, pad_, stride_)), x, k, pad = pad_, stride = stride_)
+    end
+end
+
+maxpool!(y::A, x::A, k; pad = map(_->0,k), stride = k) where A<:AbstractArray{Float64, 4} =
+    maxpool!(Float32.(y), Float32.(x), k, pad = pad, stride = stride)
+
+maxpool!(y::A, x::A, k; pad = map(_->0,k), stride = k) where A<:AbstractArray{Float32, 4} =
```
MikeInnes: Is AbstractArray really right here? Can you actually pass anything other than `Array` to nnpack?

src/nnpack/interface.jl
```diff
@@ -0,0 +1,100 @@
...
+        ∇conv_data!(zeros(Float32, size(x)), dy, x, w; pad = pad_, stride = stride_, dilation = dilation, algo = UInt32(algo))
+    end
+end
+
+function ∇conv_data!(dx::A, dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0), flipkernel = 0) where A<:AbstractArray{Float32, 4}
+    flipkernel == 0 && (w = reverse(reverse(w, dims=1), dims=2))
+    nnp_convolution_input_gradient(dx, x, dy, w, padding = pad, stride = stride, algo = algo)
+end
+
+function ∇conv_filter(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1, algo = UInt32(0)) where A<:AbstractArray{Float32, 4}
```
MikeInnes: I wonder about the fact that you're implementing both the in-place and out-of-place versions here. Couldn't you just add the in place ones, and NNlib's existing methods would dispatch the out-of-place ones?

MikeInnes:
I went ahead and removed the Float64 methods. I think I had suggested that originally, but since then I think it's better to handle that on the Flux side.

src/nnpack/nnlib.jl
MikeInnes: Ok that sounds good

MikeInnes:
I resolved some of my own comments here, but would still like to understand the use of AbstractArray and whether the compatibility checks can be moved to the in-place versions of these functions.

One other thing, I just did a quick benchmark and`conv` actually seems to be significantly slower on this branch. Any idea what's happening there?

```julia
@btime NNlib.conv($(rand(Float32, 100, 100, 3, 2)), $(randn(Float32, 2, 2, 3, 3)))
```

avik-pal:
I believe it is due to the small batch size. When I tried to compare the time, the speed was slow for batch sizes like 4, but for higher batch sizes like 256 or 128, the speed improves significantly (around 5 times more img/sec for VGG. 
Here are some MXNET Benchmarks https://aws.amazon.com/blogs/machine-learning/speeding-up-apache-mxnet-using-the-nnpack-library/. Though it is not slower but NNPACK does not make much difference at low batch size.
But still I think if there is no speedup even for large batch sizes we will have to investigate.

As for the compatibility checks they can be moved to the inplace functions. In that case  should completely drop the other functions, right?

Also AbstractArray was a wrong choice, it needs to be fixed.

avik-pal:
I tried some benchmarks
```julia
batches = [1, 2, 4, 8, 16, 32, 64, 128, 256];
w = randn(Float32, 3, 3, 3, 16)
for i in batches
    x = randn(Float32, 224, 224, 3, i)
    push!(times, @belapsed $NNlib.conv($x, $w))
    println("$i ==> $(i/times[end])") # Images processes per second
end
```
__Higher the better__

|Batch Size|NNPACK 4 threads|No NNPACK|
|:------------:|:------------------------:|:--------------:|
|1|9.3|355|
|2|18.8|349|
|4|38.7|347|
|8|75.99|337|
|16|116.55|278|
|32|181.12|164|
|64|264.1|275|
|128|371.93|243|
|256|456.7|262|


These numbers don't look very good to me and will be worth taking a look into.
However the end to end benchmarks for VGG19 looks pretty good (which is actually very strange).
nnpack benchmarks
The x axis is the batch_size index ( from the code) and y axis is the number of images per second.
y3 --> No NNPACK
y2 --> 4 threads
y1 --> 8 threads

avik-pal:
I removed the bias argument from the `conv` functions and am simply sending zeros as the bias. Doing this allows us to only have the nplace functions.

avik-pal:
@MikeInnes The NNPACK version works very well when the channel dimension is much deeper. For channel dim 64, and batch size 32 we get 31 images per sec while the current NNlib version gives us only 12 images per second.
However, seeing the critical slowdowns for certain cases it might be worth defining edge cases for which we fall back to the NNlib implementation.

avik-pal:
Here is the discussion of integrating NNPACK into Pytorch https://github.com/pytorch/pytorch/pull/2826. Seems like they have also used heuristics for determining whether or not to use NNPACK.
As expected NNPACK is slow for the cases we had benchmarked earlier https://github.com/pytorch/pytorch/pull/2826#issuecomment-333285282.

avik-pal:
IMO we should prioritize the items like:
1. Performance (whether to use or not use NNPACK). This is critical as we might sometimes end up a lot slower using NNPACK.
2. Supporting more platforms. (Supporting Windows is a difficult part as NNPACK itself doesn't support it, though there is a [version](https://github.com/zeno40/nnpack-windows) which does. I am not sure if it would be good to rely on that)
3. Numerical Conversion.

These should be followed by the code organizational issues.

Also, I believe we should have a set of functions which simply take the operation descriptor (say Conv) and for the given set of parameters returns the optimal algorithm and backend to use. Something like [CUDNN](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnFindConvolutionForwardAlgorithmEx) offers. So in future, if we extend to newer algorithms and backends we can simply extend these functions, and since the `conv` will be calling these to determine its optimal course of action, we need not make any change there. This is not even close to the tool that you are suggesting but will definitely aid in the development of one.

staticfloat:
> Performance (whether to use or not use NNPACK). This is critical as we might sometimes end up a lot slower using NNPACK.

I'm not surprised that NNPack is slower sometimes, but I am surprised that it's so much slower.  Do you have any idea why?

> Supporting Windows is a difficult part as NNPACK itself doesn't support it

Looking at the NNPACK code, I think we'll be able to cross-compile it without _too_ much difficulty.  You can leave this to me, I'll take care of it.  :)

> I believe we should have a set of functions which simply take the operation descriptor (say Conv) and for the given set of parameters returns the optimal algorithm and backend to use.

Let's do that for now.  It will be more clear than what we have now, and there will be no real performance benefits from compile-time determination yet, that will only come once we have other options for convolution that aren't just `ccall`'s.

avik-pal:
> I'm not surprised that NNPack is slower sometimes, but I am surprised that it's so much slower. Do you have any idea why?

I am not sure as to why this is the case. But I think it is mainly on the side of NNPACK and not the wrapper (though we should still confirm this using benchmarks). https://github.com/pytorch/pytorch/pull/2826#issuecomment-333585126 here there is a comparison between pytorch conv and nnpack for small batch size and NNPACK is quite slow.

staticfloat:
That's kind of surprising, because [MXNet says it should be fast](https://mxnet.incubator.apache.org/versions/master/faq/nnpack.html#nnpack-performance); I'm not sure when I'll have time to look at this, but eventually I will if you can't figure it out.

Just FYI, I'm planning on merging the transposed convolution PR first, which will break some of this PR I think.

staticfloat:
Hey @avik-pal just FYI I'm 80% of the way through [a large overhaul of NNlib](https://github.com/FluxML/NNlib.jl/compare/sf/overhaul); I'm going to open a PR tomorrow that changes a LOT of things, but will make this PR easier to deal with.  Most likely I intend to get my PR merged in (after the other maintainers and major contributors of this package have had a chance to weigh in on the changes) and then I will help adapt this PR to the new structure.

avik-pal:
Performance
@staticfloat I have updated the interface for NNPACK. Should I change the definations of `conv` for the 2d case to call nnpack by default and fallback to im2col if there is any issue?

Apart from this I need to add some tests for the functions.

Also I am not sure how the performance was fixed :P.

staticfloat:
Hey, that's great!  I'm not surprised that our performance characteristics are different; but I'm guessing the image sizes you're using are very different from your previous test?  Previously you were getting ~1 image per second at batch size 1, now you're getting 400 images per second at batch size 1, so my guess is your most recent test is using significantly smaller images?

I would like to see a comparison with a reasonable batch size (like 64) across image sizes going from 8x8 up to something large, like 512x512; just to make sure there aren't terrible performance cliffs or something.

Overall, this is really encouraging.

src/nnpack/interface.jl
```diff
@@ -0,0 +1,25 @@
...
+        :∇conv_data    => :_nnpack,
+        :∇conv_filter  => :_nnpack,
+    )
+    @eval begin
+        @timeit_debug to function $(Symbol("$(front_name)$(backend)!"))(
+                        out::Array{T1,4}, in1::Array{T2,4}, in2::Array{T3,4},
+                        cdims::ConvDims; kwargs...) where {T1, T2, T3}
+            # Output must of the same type as in the function signature
+            T1.($(Symbol("$(front_name)$(backend)!"))(Float32.(out), Float32.(in1),
+                                                      Float32.(in2), cdims; kwargs...))
```
staticfloat: Let's create two methods here; one that is specialized on `T1`, `T2`, `T3` all equal to `Float32`, and then one that is generic.  Let's include a `@warn "Automatically converting $(size(in1)) input tensor to Float32" maxlog=1` in the generic version so that users are aware that this is sub-optimal and can manually convert when necessary.

src/nnpack/interface.jl
```diff
@@ -0,0 +1,25 @@
...
+        end
+    end
+end
+
+
+function maxpool_nnpack!(y::Array{T1, 4}, x::Array{T2, 4}, pdims::PoolDims;
+                         kwargs...) where {T1, T2}
+    # We want the output to be of the same type as desired
+    T1.(maxpool_nnpack!(Float32.(y), Float32.(x), pdims; kwargs...))
```
staticfloat: Same thing here; always warn if we have the wrong datatypes.

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,38 @@
...
+    status = nnp_initialize()
+    if status == nnp_status_unsupported_hardware
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["NNPACK_CPU_THREADS"])
+    catch
+        global NNPACK_CPU_THREADS = 4
```
staticfloat: We should default this to `Sys.CPU_THREADS`

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,38 @@
...
+    else
+        return true
+    end
+end
+
+@init begin
+    check_deps()
+    status = nnp_initialize()
+    if status == nnp_status_unsupported_hardware
+        @warn "HARDWARE is unsupported by NNPACK so falling back to default NNlib"
```
staticfloat: No reason to capitalize `HARDWARE`.

src/nnpack/NNPACK.jl
```diff
@@ -0,0 +1,38 @@
...
+    else
+        include(nnlib_interface_path)
+    end
+    try
+        global NNPACK_CPU_THREADS = parse(UInt64, ENV["NNPACK_CPU_THREADS"])
+    catch
+        global NNPACK_CPU_THREADS = 4
+    end
+    shared_threadpool = pthreadpool_create(NNPACK_CPU_THREADS)
```
staticfloat: You need dereference the `Ref` via `shared_threadpool[] = pthreadpool_create(NNPACK_CPU_THREADS)`, otherwise this value does not get saved.

staticfloat: I actually think this PR right now does not use threads at all, because we are always passing `C_NULL` to the runtime.  It will be interesting to run the performance benchmarks again after changing this!

src/nnpack/error.jl
```diff
@@ -0,0 +1,83 @@
...
+        msg = "NNPACK STATUS OUT OF MEMORY"
+    elseif status == nnp_status_insufficient_buffer
+        msg = "NNPACK STATUS INSUFFICIENT BUFFER"
+    elseif status == nnp_status_misaligned_buffer
+        msg = "NNPACK STATUS MISALIGNED BUFFER"
+    end
+    NNPACKError(status, msg)
+end
+
+macro check(nnp_func)
```
staticfloat: Please rename this to `nnpack_check` to avoid conflicting with other things.

src/nnpack/libnnpack.jl
```diff
@@ -0,0 +1,135 @@
...
+function nnp_softmax_output(batch_size, channels, input, output, threadpool)
+    @check ccall((:nnp_softmax_output, libnnpack), nnp_status, (Csize_t, Csize_t, Ptr{Cfloat}, Ptr{Cfloat}, pthreadpool_t), batch_size, channels, input, output, threadpool)
+end
+
+function nnp_softmax_output(x::VecOrMat{Float32}, y::VecOrMat{Float32}; threadpool = shared_threadpool[])
+    nnp_softmax_output(ndims(x) == 2 ? size(x, 2) : 1, size(x, 1), x, y, threadpool)
+    y
+end
+
+#FIXME: Output of fully connected not consistent with `kernel * input`
```
staticfloat: Is this still the case?  If not, please delete these comments.

src/nnpack/libnnpack.jl
```diff
@@ -0,0 +1,135 @@
...
+function nnp_max_pooling_output(y::Array{Float32,4}, x::Array{Float32,4}, kernel::Tuple; padding = 0, stride = 1, threadpool = shared_threadpool[])
+    input_size = nnp_size(Csize_t.((size(x, 1), size(x, 2)))...)
+    pooling_size = nnp_size(Csize_t.(kernel)...)
+    input_padding = nnp_padding(Csize_t(padding[2]), Csize_t(padding[1]), Csize_t(padding[2]), Csize_t(padding[1]))
+    pooling_stride = nnp_size(Csize_t.(stride)...)
+    nnp_max_pooling_output(size(x, 4), size(x, 3), input_size, input_padding, pooling_size, pooling_stride, x, y, threadpool)
+    y
+end
+
+#TODO: Add wrapper for convolution inference
```
staticfloat: If this is done, please remove this comment.

src/nnpack/libnnpack.jl
avik-pal: It is not done. I was focussing on getting the convolutions right first before fixing it.

src/nnpack/libnnpack.jl
avik-pal: It is not done.

src/nnpack/interface.jl
avik-pal: if all of them are equal to Float32 won't it be exactly same as the function defined in nnpack/impl.jl

staticfloat: You're right; you've already got the base case defined, so just put a `@warn` in this one.

src/nnpack/NNPACK.jl
avik-pal: NNPACK doesn't seem to scale very well with increasing threads. On a machine with 8 default threads, 4 seems to be much better for performance. (I tested this after changing the shared_threadpool deferencing)

avik-pal:
Actually the benchmarks with 1 img per sec was for VGG16. This was for a single convolution.
Anyways I did the shared_threadpool dereferencing. Now I am getting pretty good results for high batch size but it is pretty terrible for small batch size. 
```
1 ==> 35.345990591180076
2 ==> 70.08730811014935
4 ==> 136.73528909463602
8 ==> 241.78609564931003
16 ==> 441.75463727098787
32 ==> 679.1439704356744
64 ==> 902.974004140305
128 ==> 1000.2636007170265
256 ==> 1063.1722630815611
```

staticfloat:
Hmm.  Okay.  Is it possible for us to have a `Dict{}` with multiple thread pools, one that is `C_NULL`, to disable multithreading, one with 2 threads, one with 4 threads, etc.... up to the maximum number of CPU threads?  We can then write a function that maps batch size/spatial dims/channel number to the number of threads, and intelligently assign the optimal number of threads.  Eventually we may want the user to be able to tune this with a benchmark suite, but for now let's just add the capability by creating a `select_threadpool(cdims::DenseConvDims)` method, and have it always return the `C_NULL` threadpool unless the batch size is really large.

Looking at the [NNPack convolution implementation](https://github.com/Maratyszcza/NNPACK/blob/c039579abe21f5756e0f0e45e8e767adccc11852/src/convolution-output.c#L392-L396), it appears that threading will only benefit us with lots of output channels and large batch sizes, so I think we can just do some quick benchmarking, choose reasonable defaults for when we'll kick in with larger numbers of threads, and call this a big win.  :)

staticfloat:
I also built you [a new `NNPACK`](https://github.com/JuliaPackaging/Yggdrasil/releases/tag/NNPACK-v2018.06.22-0) that builds for MacOS as well as a few other platforms like musl Linux.

avik-pal:
So do we need to setup this dict at the time when the user is installing NNlib, or do we run the benchmarks and set these as the defaults for now. In case of the default what should I use 4 as the max threads?

src/NNlib.jl
```diff
@@ -24,8 +24,6 @@ include("impl/depthwiseconv_im2col.jl")
 # Direct implementations of pooling
 include("impl/pooling_direct.jl")
 
-to = TimerOutput()
```
staticfloat: This is important to keep.

staticfloat:
> So do we need to setup this dict at the time when the user is installing NNlib

For now, let's just hardcode it.  We may eventually have a quick little micro-benchmark that takes all installed backends, compares them, and chooses the fastest at package install time, but for now I think just having something intelligent hardcoded is much better than nothing.

> In case of the default what should I use 4 as the max threads?

Sure, let's start with that.

staticfloat:
Hmmm, we're also going to need to do something to guard against invalid `stride`, `dilation`, etc... kinds of parameters.  I was just trying this out, and `stride=2` causes Julia to segfault because we allocate something smaller than what NNPACK is expecting.

avik-pal:
>  we're also going to need to do something to guard against invalid stride, dilation, etc...

Yes we need to handle that. Do we put the check in `conv_nnpack` or in `conv` where we check the validity and then call `conv_nnpack`. Similar to the design we had prior to the overhaul PR.

staticfloat:
I think we define this by using type parameters:

```
conv(x, w, cdims::DenseConvDims{2,(1,1),P,(1,1),F}) where {P,F} = conv_nnpack(x, w, cdims)
```

That will only redirect `conv()` from `conv_im2col()` -> `conv_nnpack()` if number of spatial dimensions (`N`) is `2`, `stride` is `(1,1)` and dilation is `(1,1)`.

avik-pal:
It will actually work for all cases where `(im_size + pad_1 + pad_2 - kernel` is divisible by `stride` how do we handle this like u mention?

staticfloat:
Are you sure?  Looking at the code, it doesn't look to me like NNPACK does stride at all; there's no stride parameter in the convolution routines.

avik-pal:
Ah yes, it doesn't support stride for convolution. But it does for maxpool.

staticfloat:
@avik-pal what's the status on this?  Is this ready for general testing?  I think I'd like to release `v0.6.0` of NNlib first, then merge this in right away afterwards, for those of us living on the bleeding edge to try out and have fun with.  :)

avik-pal:
The heuristics part of choosing the operation is not yet implemented. I will have to try to get it done by next week probably. Apart from this, it should be ok for general testing.

avik-pal:
I have added some basic heuristics. They are not ideal but is much faster than what we currently have in NNlib. Also I am bounding the max threads to be 8 irrespective of what number above 8 the user enters. Anything above 8 generally worsens the performance.

I have added the new build file. But I do not have access to those systems. So someone would have to verify if this PR works there.

I am working on fixing the tests they seem to be numerical issue.

avik-pal:
Indeed the test fails are due to the numerical accuracies. Should I change the equalities to `isapprox`?

staticfloat:
> Should I change the equalities to `isapprox`?

Yes, please do, I will test on a couple of systems over here once you do.

avik-pal:
@staticfloat tests are passing locally for me.

staticfloat:
It's working fine on my Macbook Pro and my Linux server.  This is great work Avik, I think we're ready to merge!

PR #71, opened 2018-10-04
[WIP] mode parameter for Convolution/cross-convolution

ayush1999:
@MikeInnes Would expressing the `mode` argument work in each function call work, or do I need to add another `crosscor` function?

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -39,8 +39,14 @@ end
 ∇conv_data(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray =
   ∇conv_data!(zero(x), dy, x, w; pad = pad, stride = stride, dilation = dilation)
 
+∇crossconv_data(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray =
+  ∇crossconv_data!(zero(x), dy, x, w; pad = pad, stride = stride, dilation = dilation)
+
 ∇conv_filter(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray =
   ∇conv_filter!(zero(w), dy, x, w; pad = pad, stride = stride, dilation = dilation)
+
+∇crossconv_filter(dy::A, x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray =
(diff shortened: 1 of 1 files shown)


ayush1999:
Also, there was some discussion in https://github.com/FluxML/Flux.jl/pull/308 about the way to express these changes in Flux. Could you give me a final decision on it so that I can go ahead and add them to Flux?

MikeInnes:
We can pass `mode` around in the implementation, but let's not expose it in the `conv` or `conv!` interface. Instead let's add `crosscor` and `crosscor!`; these should both be trivial definitions that forward to `conv(!)` with the mode changed.

src/conv.jl
```diff
@@ -45,6 +57,13 @@ function conv!(y::AbstractArray{T,3}, x::AbstractArray{T,3}, w::AbstractArray{T,
     return y
 end
 
+function crossconv!(y::AbstractArray{T,3}, x::AbstractArray{T,3}, w::AbstractArray{T,3};
+               pad = 0, stride = 1, dilation = 1) where T
+    args = map(x -> reshape(x, size(x,1),1,size(x,2),size(x,3)), (y, x, w))
+    crossconv!(args..., pad = (pad...,0), stride = (stride...,1), dilation = (dilation...,1))
```
MikeInnes: That's `crosscor`. It should also be a thin wrapper that doesn't need as much duplication as you've added here; if you just call `conv!` here with the right mode, for example, you won't need specific 2D and 3D wrappers.

ayush1999: But `conv!` doesn't take the `mode` parameter in its arguments at the moment, (and you also advised me not to expose mode in `conv!`).

MikeInnes: I think it would be ok to have it in `conv!` (perhaps as `flipkernel`) but just not document it. That makes it a bit easier to support `crosscor!` without a lot of duplication. You can also then avoid the gradient wrappers, since we can just support flipkernel in Flux.

ayush1999:
@MikeInnes I've removed all gradient wrappers and added the `flipkernel` argument to `conv!`.

MikeInnes:
Ok perfect. Thanks a lot!

jekbradbury:
It still uses “crossconv”, which should instead by “crosscor”

MikeInnes:
Yeah that's fixed in https://github.com/FluxML/NNlib.jl/commit/57276431d19890c6122b50f5d58e91741cac4236

PR #73, opened 2018-10-17
Add a lower level conv interface that doesn't use keyword arguments

Keno:
The keyword arguments to the various conv arguments affect the output
shape. If the arrays track their shape, this puts significant strain
on the compiler to track the keyword arguments through the various
layers of canonicalization (as well as needing to be captured by
value in Zygote's derivative code). To make things easier on the
compiler, create a couple of structs that behave as functions, but
have type parameters to configure the convolution. The compiler
is much happier about that and can successfully infer Zygote's
gradient function (after both Flux and Zygote are changed to
use this interface instead of the keyword arguments of course).
I don't think it's particularly pretty, but it seems to work
fine for now and we can fix it as part of a larger overhaul
of NNlib's intrinsics. To avoid disrupting users, I keep the
keyword argument versions as well (though to get the type
inference benefits one needs to avoid using those).

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -24,23 +24,38 @@ padtuple(x::Tuple,p::Integer) = map(_->p, head(head(x)))
 padtuple(x::Tuple,p::Tuple) = p
 padtuple(x::AbstractArray,p) = padtuple(size(x),p)
 
-function conv(x::A, w::A; pad = 0, stride = 1, dilation = 1) where A<:AbstractArray
+# These are defined as an interface for users that are sensitive to type inference
+# (e.g. Zygote/XLA). By avoiding keyword arguments, we can make things a little easier
+# on the compiler for now.
+struct _conv{stride, pad, dilation}; end
+struct _∇conv_data{stride, pad, dilation, flipkernel}; end
+struct _∇conv_filter{stride, pad, dilation, flipkernel}; end
+
+# Convenience wrappers for the above with keyword arguments
+conv(x::AbstractArray, w::AbstractArray; pad = 0, stride = 1, dilation = 1) =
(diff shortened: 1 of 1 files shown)


MikeInnes:
This makes me a little uneasy given how much this is a kludge for the XLA work. Presumably these type parameters are also going to propagate "upwards" (e.g. they have to be explicitly present in the Conv layer), which is not that nice. But I guess we can live with it for now, if compiler improvements are going to make it unnecessary in future.

Is it possible to add a test case to show where this fixes inference, or will that end up being overly contrived?

Keno:
XLA isn't the problem (e.g. the VGG forward pass is fine without this). Zygote not doing keyword arguments very well and capturing those by value rather than type parameter for the backpropagator closure is. If you'd prefer, we could probably hack this into Zygote.

MikeInnes:
Sure, but unless I misunderstood your original issue, that only comes up when trying to put shapes in the type system, correct?

Changing Zygote's gradient definition to lift those parameters into the type system does seem like it could be cleaner.

Keno:
> Sure, but unless I misunderstood your original issue, that only comes up when trying to put shapes in the type system, correct?

Well, yes and no the issue is general of course, but it doesn't matter too much if you're not trying to have the shapes available in the backwards pass (whether or not you actually put them in the type system or not).

> Changing Zygote's gradient definition to lift those parameters into the type system does seem like it could be cleaner.

Ok, I can look into that. We'll also have to fix Zygote's handling of keyword arguments, which is a general issue.

staticfloat:
Superseded by #94.  We have type-system-level convolution dispatch now!

Issue #68, opened 2018-09-19
Softmax and crossentropy are prone to produce NaNs


pevnak:
Hi,

I have found examples, where softmax or logitcrossentropy overflows. I will try to find the problem, but I will file the issue first.

This overflows logitcrossentropy. 
```
o = param(reshape([ -431.0;  279.0;   427.0], 3, 1))
y = [true, false, false]
f = Flux.logitcrossentropy(o, y);
Flux.Tracker.back!(f); 
Flux.Tracker.grad(o)
```
I believe the problem is in  `softmax` in this line `∇logsoftmax(Δ, xs) = ∇softmax(Δ ./ softmax(xs), xs)` in logsoftmax.jl:26.

A second example just overflows the softmax as
```
y = [true, false, false]
o = param(reshape([ -1047, -981, 1891],3,1))
Flux.Tracker.back!(sum(softmax(o))); 
Flux.Tracker.grad(o)
```
The problem here is that `softmax(o)` outputs `[0.0, 0.0, 1.0]`. The problem here is in  ∇softmax, which uses exp without checking that it might overflow.

pevnak:
This is now fixed in #75 and can be closed after merging.

PR #75, opened 2018-10-30
Softmax

pevnak:
This is a fix to an issue #68 (Softmax and crossentropy are prone to produce NaNs). I have added tests that the gradient does not produce any NaNs). I have also add more test of gradients of these functions to `runtests.jl` and also a check that gradients of these function do not overflows.


MikeInnes:
Seems reasonable to me; we can't add Flux as a test dependency though.

pevnak:
I have fixed that by adding calculated gradients.

MikeInnes:
Works for me. Thanks!

PR #79, opened 2018-12-05
RFC: Loosen the signature for conv

JobJob:
So it accepts AbstractArrays of the same eltype, ndims, but possibly different AbstractArray types


JobJob:
Oh just saw that this is a dup of #63

PR #83, opened 2019-01-01
implment gelu

chengchingwen:
Implementation of gaussian error linear unit


chengchingwen:
@MikeInnes Anything I missed?

PR #88, opened 2019-01-28
improve performance for pooling in forward mode

### LINKED ISSUE #87: The slicing in the inner loops in the pooling layers are expensive
Ref https://gist.github.com/KristofferC/34178cf812f2b3ea468ca217a103c4a6

Removing the slicing (and adding an `@inbounds`) gives a 30x speedup for the maxpooling.

KristofferC:
Fixes #87 

I tried versions using `view`s and generators but they were all slow.

### DIFF
--- a/src/impl/pool.jl
+++ b/src/impl/pool.jl
@@ -2,7 +2,7 @@ function max_pooling2d_fwd!(x::AbstractArray{T,4}, y::AbstractArray{T,4},
                             width::Int, height::Int, channels::Int, num::Int, pooled_width::Int,
                             pooled_height::Int, kernel_w::Int, kernel_h::Int, pad_w::Int, pad_h::Int,
                             stride_w::Int, stride_h::Int) where T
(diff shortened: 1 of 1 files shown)


src/impl/pool.jl
```diff
@@ -11,7 +11,11 @@ function max_pooling2d_fwd!(x::AbstractArray{T,4}, y::AbstractArray{T,4},
     hstart = max(hstart, 0) + 1
     wstart = max(wstart, 0) + 1
 
-    y[pw, ph, c, n] = maximum(x[wstart:wend, hstart:hend, c, n])
```
vchuravy: Did you try just using a `view`?

KristofferC: Look at top comment :p. Anyway, to elaborate, the view gets allocated and the loop is super tight so it has terrible performance.

MikeInnes:
Awesome stuff, thanks @KristofferC.

Issue #77, opened 2018-11-28
generalize softmax


CarloLucibello:
the softmax functions should be generalized to handle reduction across any dimensions, e.g.:
```julia
softmax(x, dims=2)
softmax(x, dims=(1,3))
```

bhvieira:
Is this solved? I have some interest in this to add attention support in Flux.

CarloLucibello:
Looking at this https://github.com/FluxML/NNlib.jl/blob/342928eb4478da9c7b1433ec75c8eb8a9b155747/src/softmax.jl
it seems that the issue with `softmax` has been solved, but `softmax!` and `logsoftmax` still don't support reduction over arbitrary dimensions.

mcabbott:
The `softmax` on master (which takes `dims=1`) is only half as fast as the old one:
```julia
mm = rand(20,100);                                          
@btime softmax($mm);  # tagged, with softmax!,  23.076 μs (1 allocation: 15.75 KiB)
@btime softmax1($mm); # master, with dims=1,    47.657 μs (13 allocations: 33.45 KiB) 
```
Was this discussed somewhere? Are there goals besides being generic? Some variants which are almost as fast:
```julia
function softmax2(xs::AbstractArray{T}; dims=1) where {T}
    temp = maximum(xs, dims=dims)
    out = exp.(xs .- temp)
    out ./= sum!(temp, out)
end
function softmax3(xs::AbstractArray{T}; dims=1) where {T}
    max = maximum(xs, dims=dims)
    out = exp.(xs .- max)
    out ./ sum(out, dims=dims)
end

@btime softmax2($mm); # re-using temp,   27.382 μs (11 allocations: 16.83 KiB) 
@btime softmax3($mm); # no mutation,     26.462 μs (13 allocations: 33.45 KiB)  
```

CarloLucibello:
`softmax` and `logsoftmax` now support  `dims` keyword, so this can be closed. 
@mcabbott fill free to open a new Issue/PR if you think there is some performance problem.
The current implementation seems to be exactly the same as your `softmax3` 
https://github.com/FluxML/NNlib.jl/blob/c82a76d40cf5c34e4a1eed2655aa30b464e3e408/src/softmax.jl#L28

PR #94, opened 2019-02-21
Major overhaul of NNlib

staticfloat:
As I was on vacation last week, I tried to write a convolutional autoencoder, but was frustrated by the slow performance of our `im2col` implementation.  While I alleviated that somewhat, I also got frustrated by what I saw as unnecessary complexity in `NNlib`, and inadequate testing coverage.  To that end, I have performed a major overhaul of `NNlib`, arriving at what I believe to be a much more "coherent" API, with significantly improved inline documentation/comments, and a test set that approaches the border of exhaustive.  Many thanks to @MikeInnes, @tejank10, @avik-pal and everyone else who has braved this codebase before me.  This kind of code is gnarly, and I owe the previous authors for giving me a foundation to massage, because there's no way I would have wanted to write this from scratch.

This is a giant monolithic commit because so many of the changes are interleaved that I couldn't really find a good way to break it up.  I realize this will be a nightmare to review, however my hope is that things will be "clean" enough, fast enough, and well-tested enough that it will be an obvious win to merge it without too many further changes.   Here's the list of highlight changes:

* I added a new type-system-level operation shape system.  Things like `DenseConvDims`, `DepthwiseConvDims`, `PoolDims`, etc... allow the type system to know things about stride, padding, dilation, flipped kernels, input shape, output shape, etc...  This is beneficial to things like `im2col`, and will also be beneficial to tangential projects like XLA.jl
* I changed the naming of all spatial operations to be completely shape-based, e.g. no more `conv2d()`, instead you just invoke `conv()` with a 4-dimensional tensor.
* I added "direct" pure-Julia implementations of all convolution operations.  Not intended to be highly performant, they're supposed to be used as last-ditch fallbacks and for testing, so that we can check all our implementations against eachother.
* Beefed up testing significantly, adding "fuzzing" tests that take many minutes to run, but which have caught a LOT of bugs.  I now have much higher confidence that NaN's aren't going to sneak into the output because of funny dilation/stride combinations.
* I removed all `@threads` pending future threading improvements, as in my tests it either generated invalid results or was slower than the serialized versions.
* Removed `log_fast()` as it doesn't seem to be used anywhere.

Things I still need to do before this can be merged:

- [ ] Performance testing.  I've made the implicit assumption that the reshape-use-generic-3d-versions-of-things strategy isn't going to be any slower than what we had before; benchmark this and prove it.
- [x] Update any broken `Flux` APIs in a branch over there to make sure that it can coexist with these changes easily (this should be pretty easy) (Exists here: https://github.com/FluxML/Flux.jl/tree/sf/nnlib_overhaul)
- [ ] Update CuArrays/other dependent GPU packages

Things this work should enable/make easier in the future:
* This should make it much easier to integrate things like NNPACK, MAGMA, etc...  These acceleration packages should be able to define things like `conv_nnpack!()` and should be trivially plug-in-able.  We can then benchmark the various implementations against eachother, and choose the right kernel using a lookup table.
* I've taken pains to make sure that nonallocating versions of all function calls are truly as nonallocating as possible.  This should enable us to, in Flux, track memory allocations the first time, then pass in preallocated buffers for everything and decrease our large allocations significantly.


src/pooling.jl
```diff
@@ -0,0 +1,123 @@
...
+        function $(Symbol("$(front_name)!"))(
+                dx::AbstractArray{T,5}, dy::AbstractArray{T,5},
+                x::AbstractArray{T,5}, pdims::PoolDims; kwargs...) where {T}
+            $(Symbol("$(front_name)_$(backend)!"))(dx, dy, x, pdims; kwargs...)
+        end
+    end
+end
+
+
+# Our strategy for 1d and 2d convolution is to reshape from to 3d convolutions, which
```
jekbradbury:
```suggestion
# Our strategy for pooling is to reshape to an array with three spatial dimensions, which
```

src/softmax.jl
```diff
@@ -52,4 +17,69 @@ independent.
       0.244728
       0.665241
 """
-softmax
+softmax(xs) = softmax!(similar(xs), xs)
+
+function softmax!(out::AbstractVecOrMat{T}, xs::AbstractVecOrMat{T}) where {T}
```
jekbradbury: Is it possible (without hurting performance) to make this generic over the dimension index? (related: it would be pretty nice if using the clean broadcast versions in the comments didn't hurt performance either—is that because the broadcast infrastructure doesn't insert the equivalent of `@inbounds`?)

src/conv.jl
```diff
@@ -1,266 +1,157 @@
...
+    for name in (:conv, :depthwiseconv)
+        @eval begin
+            function $(Symbol("$(name)$(backend)"))(
+                    x::AbstractArray{xT,N}, w::AbstractArray{wT,N},
+                    cdims::ConvDims; kwargs...) where {xT, wT, N}
+                yT = promote_type(xT, wT)
+                # Annoyingly, we must allocate with `zeros()` because if we were to use
+                # the faster `similar()`, it may have NaNs within it, which will poison
+                # the output because we support accumulation (even with `beta = 0` the
+                # NaNs poison us as NaN * 0 == NaN).  This is a bit of a shame, but it's
```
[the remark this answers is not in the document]
MikeInnes: In case it's not clear why: `NaN*false == 0.0`. That's a neat trick that I had no idea about.

MikeInnes:
This is great, awesome work Elliot. It'd be great to have a once-over from others who have looked at this code, but otherwise I'm mostly only concerned about the API, which should be simple to review. 

Presumably we'll also need to update CuArrays to add the GPU wrappers; might be worth setting that up on a branch so that we can see how it looks.

src/conv.jl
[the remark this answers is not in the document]
staticfloat: Okay this is neat; looks like I can use `beta=0.0` for the BLAS calls (they do not get `NaN` poisoned for some reason; perhaps they have some special handling for this case) and `beta=false` for the pure Julia implementations.  Thanks for the tip, @jekbradbury!

src/softmax.jl
staticfloat: In my experience, broadcasting creates small objects, and so if you're doing lots of small broadcasts, you can easily get drowned in small allocations.  It's almost always faster to just write the loop yourself, unfortunately.

staticfloat: I'm going to punt any improvements here off onto https://github.com/FluxML/NNlib.jl/issues/77; this PR is big enough as it is.  ;)

src/pooling.jl
```diff
@@ -0,0 +1,123 @@
...
+                    x::AbstractArray{xT,N}, pdims::PoolDims; kwargs...) where {xT, N}
+                y = zeros(xT,  output_size(pdims)..., channels_out(pdims), size(x, N))
+                return $(Symbol("$(name)$(backend)!"))(y, x, pdims; kwargs...)
+            end
+            
+            # Backprops too
+            function $(Symbol("∇$(name)$(backend)"))(
+                    dy::AbstractArray{T,N}, x::AbstractArray{T},
+                    pdims::PoolDims; kwargs...) where {T, N}
+                dx = zeros(T, input_size(pdims)..., channels_out(pdims), size(dy, N))
```
avik-pal:
```suggestion
                dx = zeros(T, input_size(pdims)..., channels_in(pdims), size(dy, N))
```
Doesn't make a difference as `in_channels = out_channels` for pooling but dx should have dimensions of input x.

staticfloat: You're right; I like this.  Thanks!

src/conv.jl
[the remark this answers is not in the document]
StefanKarpinski: Yes, `false` is a "strong zero" in the sense that `false*x` is `zero(x)` and `true` is a "strong unit" in the sense that `true*x` is always `x`.

Manifest.toml
```diff
@@ -96,9 +63,11 @@ uuid = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 deps = ["Distributed", "InteractiveUtils", "Logging", "Random"]
 uuid = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
-[[UUIDs]]
-deps = ["Random"]
-uuid = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+[[TimerOutputs]]
+deps = ["Crayons", "Printf", "Test", "Unicode"]
+path = "/Users/sabae/.julia/dev/TimerOutputs"
```
StefanKarpinski: This is WIP, I presume.

staticfloat: Correct; we're about to land https://github.com/KristofferC/TimerOutputs.jl/pull/50 after which I'll eliminate this `dev` business.  Good eye, but I guess I should expect that you of all people would be checking out the `Manifest` diff.

staticfloat:
Alright; a little bit of progress, and some performance measurements.  Looks like it's not as much of a win as I had originally measured; we tend to win on large image sizes, but not necessarily on small image sizes.  This is hardly surprising, but still a little disappointing.  Here are minimum benchmark results for `conv2d` across different image sizes showing this effect:

conv2d

conv2d_data

conv2d_filter

Note that the `conv2d_data()` results are more or less expected, as I know that our `col2im()` implementation is slow; it's not as memory-friendly as it should be, some loops need to be re-ordered, but I'll need to sit down and think about how that can be done sometime in the future.

As expected, the biggest perf gains are when there is no padding to deal with, because one of the big pieces of this PR was to separate padding considerations from the inner loop body.

There are some nasty inference gotchas that can arise seemingly out of the woodwork, suddenly killing performance.  (e.g. when applying `@timeit_debug` annotations to methods).  This codebase will be a really good test suite for a future static inference proving tool, as it's the kind of code that we really should be able to prove everything about, modulo compiler limitations.

MikeInnes:
Status? Since this is quite disruptive in terms of other people's PRs, it'd be great to get it landed -- once you're happy with it of course.

staticfloat:
Alright I'm calling this good.  Future performance work can happen later, and since I've got a `Flux` branch that works with this I think we can merge now and patch up e.g. the GPU ecosystem before the next release.

MikeInnes:
Great, please merge at your leisure then.

Issue #85, opened 2019-01-17
The relu definition is a bit "scary" for array inputs.


KristofferC:
The following seems like an easy mistake to make:

```jl
julia> x = randn(5)
5-element Array{Float64,1}:
 -1.0483606281411753
 -0.13524020825581384
 -2.184303973828124
  0.4308162306092796
 -0.673766167758703

julia> relu(x)
5-element Array{Float64,1}:
 0.0
 0.0
 0.0
 0.0
 0.0

julia> relu.(x)
5-element Array{Float64,1}:
 0.0
 0.0
 0.0
 0.4308162306092796
 0.0
````

**Comment by MikeInnes** [comment]
I agree but I'm not sure what to do about it. I guess we can just disable `relu` for arrays, and perhaps re-enable if someone complains about it.

**Comment by KristofferC** [comment]
Or add a vectorized definition for it.

```
relu(x::AbstractArray) = clamp.(x, 0, Inf)
```

or something.

**Comment by MikeInnes** [comment]
Strictly speaking you could argue that this changes the meaning of `relu`, and it's not consistent with the way that Base has gradually moved away from special-cased vectorised operations. But it's hard to imagine that being a big problem in practice.

Happy to take a PR for that fix.

**Comment by KristofferC** [comment]
> Base has gradually moved away from special-cased vectorised operations. 

Base typically have type annotations though

```
julia> sin(rand(5))
ERROR: MethodError: no method matching sin(::Array{Float64,1})
```

**Comment by KristofferC** [comment]
But yes, there are cases where `f(x::AbstractArray)` is not equal to `f.(x)`, like for trigonometric functions for matrices.

**Comment by MikeInnes** [comment]
Yeah true. Maybe the better fix is to just define `relu(::Real)` only; it might make sense to do that across our activation functions.

**Comment by staticfloat** [comment]
I would support a `help.jl` file filled with things like:

```julia
relu(x::Array) = error("Use explicitly-broadcasted invocations such as `relu.(x)` to apply activation functions to tensors!")
```

These little gotchas can be very helpful to users coming from less-enlightened frameworks.

staticfloat:
Fixed by #98

PR #98, opened 2019-03-25
Allow only real input for activation functions

devmotion:
This PR implements what was discussed in https://github.com/FluxML/NNlib.jl/issues/85. A simple macro ensures that activation functions are defined for real inputs only and calling them with an array results in a (hopefully informative) error message.


MikeInnes:
I like the idea but would prefer not to annotate each definition like this. How about `σ(x::Real) = ...` and then later in the file, `@onlyreal σ, relu, ...` to add the error message to all functions. A simple `@eval` loop might actually be easier than a macro here.

devmotion:
I agree, as soon as I had written this initial implementation I thought that would have been easier :smile: I guess, it might also simplify overloading existing methods - e.g., I was wondering whether it would make sense to have a common implementation of `softplus` in NNlib and StatsFuns (the implementation [there](https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/basicfuns.jl#L66-L67) seems to be numerically more stable, and there exists a [PR](https://github.com/JuliaStats/StatsFuns.jl/pull/57) to improve it even further)? What's the reason for having different implementations in different packages? I guess it's not intended to add a dependence on Rmath to NNlib, so maybe it would be better to move those implementations to a separate package on which both StatsFuns and NNlib could depend (or to remove the R dependency from StatsFuns :stuck_out_tongue:).

MikeInnes:
> What's the reason for having different implementations in different packages?

Only Conway's law. And I think there's some advantage in exposing `softplus` as a name, just because it's familiar. Of course that doesn't mean we should have multiple impls.

I think a StatsFuns dependency would be ok, especially since Rmath is going away eventually. We probably want all of it to work well with the GPU and AD ecosystem anyway.

MikeInnes:
Note that we want to keep the manifest (it's useful for development/testing, especially if we need to use unreleased functionality temporarily). If it needs an update feel free to do that.

devmotion:
Ah I see, sorry for that! Cleaning the requirements lead to test errors on Travis, so I thought the easiest solution would be to remove the Manifest all together. I updated it.

src/activation.jl
```diff
@@ -108,12 +108,18 @@ end
...
 
 See [Deep Sparse Rectifier Neural Networks](http://proceedings.mlr.press/v15/glorot11a/glorot11a.pdf).
 """
-softplus(x) = ifelse(x > 0, x + log1p(exp(-x)), log1p(exp(x)))
+softplus(x::Real) = ifelse(x > 0, x + log1p(exp(-x)), log1p(exp(x)))
+
+# Provide an informative error message if activation functions are called with an array
+for f in (:σ, :σ_stable, :logσ, :relu, :leakyrelu, :elu, :gelu, :swish, :selu, :softsign, :softplus)
+  @eval $(f)(x::AbstractArray, args...) =
+    error("Use explicit invocations such as `", $(string(f)), ".(x)` to apply activation functions to tensors!")
```
MikeInnes:
```suggestion
    error("Use broadcasting (`", $f, ".(x)`) to apply activation functions to arrays.")
```

I think `$f` will print the same but worth checking quickly.

MikeInnes:
Unfortunately the changes to the manifest etc are conflicting now, but that should be easy to fix. I suggested a tweak to the error message but otherwise this is good to go!

src/activation.jl
devmotion: Actually in my initial implementation I used `$f`, but I added `string` since I was not satisfied with the output. With just `$f` I get, e.g., 
> ERROR: Use broadcasting (`NNlib.σ.(x)`) to apply activation functions to arrays.
whereas with `$(string(f))`
> ERROR: Use broadcasting (`σ.(x)`) to apply activation functions to arrays.
is printed.

MikeInnes: I see, a bit surprising that they behave differently, but that makes sense.

MikeInnes:
Perfect, thank you!

PR #99, opened 2019-03-26
Changed output size calculation to support kernel size such as (x,y)

thebhatman:
This patch fixes the issue of convolution breaking for filters such as (x,y) where x not equal to y. 
https://github.com/FluxML/Flux.jl/issues/700

### DIFF
--- a/src/impl/conv.jl
+++ b/src/impl/conv.jl
@@ -18,7 +18,7 @@ img_size(c::ConvDims{I,K,C,S,P,D,F}) where {I, K, C, S, P, D, F} = I
 # Calculate the output dimensions of this convolution
 function output_size(c::ConvDims{I,K,C,S,P,D,F}) where {I, K, C, S, P, D, F}
     O_w = div(I[1] + P[1] + P[2] - (K[1] - 1) * D[1] - 1, S[1]) + 1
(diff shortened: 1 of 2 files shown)


jw3126:
Can you add a testcase? I am not very familiar with the API of NNLib, but I think something like
```julia
ker31 = randn(3,1)
ker13 = randn(1,3)
ker33 = conv(ker13, ker31)
data = randn(10,10)
@test conv(ker33, data) ≈ conv(ker13, conv(ker31, data))
```
is a good test.

thebhatman:
I have added the test cases.

dhairyagandhi96:
LGTM, Thanks!

dhairyagandhi96:
LGTM, Thanks! Perhaps add the above mentioned test in as well?

thebhatman:
The problem with the above test cases is that data, ker13, ker31 have to be 4d tensors to be given as input to conv. Also it is not possible to do ker13 *ker31 for 4d tensors if I assume size(ker13) = (1, 3, 1, 1) and size(ker31) = (3,1,1,1)

dhairyagandhi96:
4d tensors would cause issue, yeah. Good catch.

dhairyagandhi96:
thanks, again

PR #108, opened 2019-04-04
Add `logcosh` activation

pshashk:
Added numerically stable `log(cosh(x))` https://github.com/FluxML/Flux.jl/pull/68.


test/activation.jl
```diff
@@ -125,4 +128,6 @@ end
             @eval @test logsigmoid.($T[-100_000, 100_000.]) ≈ $T[-100_000, 0.]
         end
     end
+
+    @test logcosh(1_000.0) + log(2) == 1_000.0
```
staticfloat: I like this test a lot.  Nice work.

PR #112, opened 2019-04-24
Added Pixel Shuffle Function

shreyas-kowshik:


shreyas-kowshik:
Shifted function from : https://github.com/FluxML/Flux.jl/pull/735 , as discussed with @dhairyagandhi96 .

I have created a new file specifically for these sort of operations on convolutions as adding this to `conv.jl` din't seem to fit in right, as it is filled with optimization and convolution utility functions.

All such operations in future can be added to `conv_op.jl`. Is it fine?

aterenin:
I've also implemented a pixel shuffle, as described here: https://arxiv.org/abs/1609.05158

My implementation is different from yours (just a `reshape` and `permutedims`, no need to partition) and, on running the code, doesn't shuffle pixels precisely the same way. Do you have a reference describing in detail the operation implemented here?

shreyas-kowshik:
I've followed the following for refernce : 
https://github.com/brade31919/SRGAN-tensorflow/blob/7f7ec0f929feb64a111221dc992f740cd48e11eb/lib/ops.py#L85-L112

shreyas-kowshik:
@dhairyagandhi96 @staticfloat What else needs to be done more in this?

tbenst:
Is this feature available in Flux yet? Considering porting code from pytorch that uses this. Thanks!

CarloLucibello:
superseded by #260

PR #121, opened 2019-05-31
maxpool and meanpool added

thebhatman:
Now pooling layers can be called simply as maxpool(x, (2,2)), without having to stick to the PoolDims API.


avik-pal:
@thebhatman I think it would be good to have a couple of tests verifying this API.

src/pooling.jl
```diff
@@ -136,3 +136,20 @@ if is_nnpack_available()
         return func(x, pdims; kwargs...)
     end
 end
+
+expand(N, i::Tuple) = i
+expand(N, i::Integer) = ntuple(_ -> i, N)
+
+function maxpool(x, k::NTuple{N, Integer}; pad = 0, stride = k) where N
+    pad = expand(Val(2*N), pad)
```
MikeInnes: Why is `pad` twice as long here? I'd expect padding to only apply to the dimensions you actually pool.

thebhatman: Yeah I actually had pad the same length as the kernel. But then I looked at this implementation of MaxPool in Flux https://github.com/FluxML/Flux.jl/blob/8ee6af1bee39bd9f6ed8e6cdf854ea9aaccb14dd/src/layers/conv.jl#L287. Here it is twice the length. Should I make it the same as length of kernel?

MikeInnes: @staticfloat do you know why this is?

thebhatman: I tried with 2*N and N separately. In both cases, the output size is the same.

avik-pal: I believe this is `2 * N` for supporting asymmetric padding. If kernel is of length 2, then for 2 dimensions we will have to specify padding along 4 directions.

MikeInnes: Ah, that makes sense.

Assuming we can just pass `N` in for symmetric padding, and let that be handled later on, that's better here.

staticfloat: Yep; it should all get canonicalized to `2*N` later on, once the `PoolDims` object is created.

staticfloat:
I intentionally removed this API from the last release because I wanted to encourage usage of the `PoolDims` API; it has the opportunity (not realized yet, but may be in the future, pending future compiler enhancements and greater usage of StaticArrays) for much higher performance than this style of API.

Is there a reason you need non-`PoolDims` APIs, or is it just for ease of programming?

MikeInnes:
From my side, mainly for ease of programming and interface consistency; it's unfortunate that the curried version of this `MaxPool((2,2),x)` is so different from the direct version `maxpool(x, PoolDims(x, (2, 2)))`.

Can you also elaborate a bit on the performance considerations? It's not obvious to me why `maxpool(x, (2, 2))` should be slower / how it gives more information to the compiler.

staticfloat:
> Can you also elaborate a bit on the performance considerations?

What the `PoolDims` does is push the shapes of all operations (kernel size, stride, dilation, etc...) into the type domain, which means that, with aggressive enough type specialization, we have a hope of things like argument checking, small for loops, etc... being compile-time folded away and unrolled.  For large images, this shouldn't have much of an impact at all, but for small operations, I expect this could give a handsome speed boost once the compiler is able to handle it, and once we have `StaticArray`-like types more thoroughly integrated into the ecosystem.

It's not interactive usage I'm concerned about here, it's usage by things like `Flux`, which should always be using the `PoolDims` API (and not creating the `PoolDims` every time, but instead creating it once and then using it again and again so that the compiler can infer as far as it possibly can).  My only concern is that we don't want oblivious users to bake in slow codepaths by using this API.

However, if we want it, we can definitely create wrappers for all APIs, `conv()`, `maxpool()`, etc...  It's up to you, I don't care that strongly, just want to nudge people in the right direction.

src/pooling.jl
avik-pal: @thebhatman As per the discussion, it makes more sense to pass `N` here.

src/pooling.jl
```diff
@@ -136,3 +136,20 @@ if is_nnpack_available()
...
+    stride = expand(Val(N), stride)
+    pdims = PoolDims(x, k; padding = pad, stride = stride)
+    return maxpool(x, pdims)
+end
+
+function meanpool(x, k::NTuple{N, Integer}; pad = 0, stride = k) where N
+    pad = expand(Val(2*N), pad)
```
avik-pal: Same here

thebhatman: Done. I have changed it to N.

MikeInnes:
Allocating a `PoolDims` upfront makes sense, but it's a little weird to tell users that given that we don't do it in Flux right now (and can't in general since MaxPool and Chain and so on have to accept inputs of multiple shapes, though having shape inference in `Chain` might help with this).

I guess the question is whether `PoolDims` will ultimately need dynamic information or not. Right now, AFAICT, it doesn't; it just needs `ndims(x)-2` and `(2,2)` to be in the type system, neither of which are inherently un-inferable if we coax the compiler a bit. So at least as things stand now, `maxpool(x, (2,2))` can be just as fast as the explicitly-constructed version.

If `PoolDims` will eventually depend directly on `size(x)`, that reasoning breaks down. But in that case it seems like we'd still want a fallback for `x` of unknown shape, since otherwise we're forcing a performance hit on a lot of more dynamic models.

staticfloat:
Agreed; let's cross this bridge when we actually have compiler support for it.

@thebhatman can you add wrappers for `conv()` and `depthwiseconv()` as well?  Just packing kwargs into `DenseConvDims` and `DepthwiseConvDims` objects.  Note that if you want to get the gradients, having the original `*Dims` object around makes it a lot easier (you don't have to track a bunch of other values, then invert them to get the proper stride/dilation/padding for inverse operations).  Thank God that `Zygote` allows capturing the forward-pass arguments, makes defining [the adjoint for `conv`](https://github.com/FluxML/Zygote.jl/blob/master/src/lib/nnlib.jl#L17-L25) much easier.

avik-pal:
This PR looks good to me. The appveyor build fail is just a network error. Should I go ahead and merge it?

PR #122, opened 2019-05-31
Avoid error while trying to build NNPACK on unsupported platforms

avik-pal:
Replace the error in NNPACK build with a warning.


MikeInnes:
:+1: Thanks!

PR #127, opened 2019-06-10
Added wrappers for conv and depthwiseconv

thebhatman:


src/conv.jl
```diff
@@ -161,3 +163,19 @@ if is_nnpack_available()
         return conv_nnpack(x, w, cdims; kwargs...)
     end
 end
+
+function conv(x, w::AbstractArray{T, N}; stride = 1, pad = 0, dilation = 1) where {T, N}
```
staticfloat: I think you're missing the `flipped` boolean parameter.

src/conv.jl
```diff
@@ -161,3 +163,19 @@ if is_nnpack_available()
...
+    stride = expand(Val(N-2), stride)
+    pad = expand(Val(N-2), pad)
+    dilation = expand(Val(N-2), dilation)
+    cdims = DenseConvDims(x, w; stride = stride, padding = pad, dilation = dilation)
+    return conv(x, w, cdims)
+end
+
+function depthwiseconv(x, w::AbstractArray{T, N}; stride = 1, pad = 0, dilation = 1) where {T, N}
```
staticfloat: I think you're missing the `flipped` boolean parameter.

thebhatman: Done. I have added the flipped parameter. Thank you!

test/conv.jl
```diff
@@ -639,3 +639,21 @@ end
...
+    @test size(conv(x, w1; stride = (1, 2), pad = (2, 3))) == (12, 7, 16, 10)
+end
+
+@testset "depthwiseconv_wrapper" begin
+    x = rand(10, 10, 3, 10)
+    w = rand(2, 2, 3, 3)
+    w1 = rand(3, 4, 3, 3)
+    @test size(depthwiseconv(x, w)) == (9, 9, 9, 10)
+    @test size(depthwiseconv(x, w; stride = (2, 2), pad = (2, 2))) == (7, 7, 9, 10)
+    @test size(depthwiseconv(x, w1; stride = (1, 2), pad = (2, 3))) == (12, 7, 9, 10)
```
staticfloat: One last nitpick; for both of these, let's exercise every parameter.  `stride`, `pad`, `dilation`, `flipped`.

thebhatman: Sure. I have now added two more tests with all parameters.

PR #129, opened 2019-06-12
Remove timer outputs as they mess with type inference

### LINKED ISSUE #125: Convolution is not inferable
e.g.
```julia
x = rand(10, 10, 3, 2)
w = rand(2, 2, 3, 4)
@code_typed conv(x, w, DenseConvDims(x, w)) # AbstractArray{yT, 4} where yT
```

@staticfloat any idea what might be going wrong here?

avik-pal:
@staticfloat I am removing the TimerOutputs given they mess with the type inference. Let me know if there are any other possible solutions to this issue
Fixes #125

### DIFF
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -1,9 +1,6 @@
 module NNlib
 using Requires, TimerOutputs
 
-const to = TimerOutput()
-
-
 # Include APIs
 include("dim_helpers.jl")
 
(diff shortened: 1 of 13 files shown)


src/NNlib.jl
```diff
@@ -1,9 +1,6 @@
 module NNlib
 using Requires, TimerOutputs
```
staticfloat: Don't forget this one.  :P

staticfloat:
Looks like you missed some in `src/nnpack/impl.jl`

staticfloat:
Also, let's add tests to ensure that `conv()` is inferred.

staticfloat:
Tada!  Passing tests.  Excellent.

PR #130, opened 2019-06-22
Make softmax! dimension-agnostic

merckxiaan:
I tried reducing the allocations for a dimension-agnostic softmax implementation but there is still a few allocations. On my pc benchmarking this against the original implementation gives very similar results however.
As for `logsoftmax`, I'm not quite sure what's happening in the original implementation so I've not yet tried to generalize that.
EDIT:
Of course, I should note I also haven't changed the derivatives yet since I'm not sure what they should be.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -20,7 +20,7 @@ independent.
 softmax(xs; dims=1) = softmax!(similar(xs), xs, dims)
 
 function softmax!(out::AbstractArray{T}, xs::AbstractArray{T}, dims) where {T}
-    max = maximum(xs, dims)
+    max = maximum(xs, dims=dims)
     out .= exp.(xs.-max)
     out ./= sum(out, dims=dims)
     return out
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -3,23 +3,23 @@ export softmax, softmax!, ∇softmax, ∇softmax!,
 
 """
     softmax(xs) = exp.(xs) ./ sum(exp.(xs))
+
 [Softmax](https://en.wikipedia.org/wiki/Softmax_function) takes
 log-probabilities (any real vector) and returns a probability distribution that
 sums to 1.
+
 If given a matrix it will treat it as a batch of vectors, with each column
 independent.
+
     julia> softmax([1,2,3.])
     3-element Array{Float64,1}:
       0.0900306
       0.244728
       0.665241
 """
-softmax(xs) = softmax!(similar(xs), xs)
-
 function softmax(xs::AbstractArray{T}; dims=1) where {T}
     max = maximum(xs, dims=dims)
-    out = exp.(xs .- max)
-    out = out ./ sum(out, dims=dims)
+    out = exp.(xs .- max) ./ sum(exp.(xs .- max), dims=dims)
 end
 
 function softmax!(out::AbstractVecOrMat{T}, xs::AbstractVecOrMat{T}) where {T}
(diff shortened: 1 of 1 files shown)


merckxiaan:
`softmax(Flux.param(rand(10,10)), dims=2)` throws an error: 
`MethodError: no method matching Float64(::Tracker.TrackedReal{Float64})`.
I believe this is because `similar(TrackedArray)` returns a regular array but I'm not quite sure why the error doesn't occur when using `dims=1`?
Thanks
Jules

MikeInnes:
If we're going to do this, I think the best way might be to just remove the in-place version entirely and write the concise array-level version.

Probably, the gradient issue is because `softmax(x)` has an explicit gradient but `softmax(x; kw...)` doesn't, or something similar. After adding this we'll have to have PRs to Tracker, Zygote etc.

src/softmax.jl
```diff
@@ -17,32 +17,12 @@ independent.
       0.244728
       0.665241
 """
-softmax(xs) = softmax!(similar(xs), xs)
+softmax(xs; dims=1) = softmax!(similar(xs), xs, dims)
```
MikeInnes: Instead of forwarding to `softmax!`, can you just write an out-of-place version here? That'll fix the gradient issue.

MikeInnes:
Sorry, I don't think we should actually remove definitions here even if they aren't used directly by NNlib (e.g. gradients and in place versions). If you can simplify how they're implemented that's still useful. But they get overloaded by things like CUDNN so we'll need to keep them for compatibility at least.

merckxiaan:
Right, I'm still figuring out all this git stuff.
So only the `softmax(xs; dims)` should be added and the rest should just stay, correct?

src/softmax.jl
```diff
@@ -3,14 +3,11 @@ export softmax, softmax!, ∇softmax, ∇softmax!,
 
 """
     softmax(xs) = exp.(xs) ./ sum(exp.(xs))
-
```
MikeInnes: There are a few unrelated docstring changes here that'd be good to remove.

src/softmax.jl
```diff
@@ -19,6 +16,12 @@ independent.
 """
 softmax(xs) = softmax!(similar(xs), xs)
 
+function softmax(xs::AbstractArray{T}; dims=1) where {T}
+    max = maximum(xs, dims=dims)
+    out = exp.(xs .- max)
+    out = out ./ sum(out, dims=dims)
```
MikeInnes: I think it'd be easier to write `out` in one line here (with the benefit of fusing everything).

Also, if you add this method we'll need to remove the first `softmax(xs)` method, since they overlap in signature.

Otherwise looking good!

MikeInnes:
Would it be easy to generalise the definition of the gradient to handle dimensions as well? Not necessarily essential but would be a big help in updating our AD, etc.

merckxiaan:
I'm not sure what the derivative function should look like, presumably it isn't as straightforward as adding `dims=dims` everywhere?
I believe this is something that should be implemented by someone more experienced than myself.

A huge thank you also for guiding me through this pr as this is my first real code contribution, albeit a rather small one.

jekbradbury:
> I'm not sure what the derivative function should look like, presumably it isn't as straightforward as adding `dims=dims` everywhere?

I'm pretty sure that's all you need to do.

MikeInnes:
Great, thanks. We'll need to update the CUDA wrappers so that CUDNN gets called, but this shouldn't in itself break stuff.

MikeInnes:
Thanks @merckxiaan!

PR #135, opened 2019-09-26
logsoftmax with dims

mcabbott:
For #77, this adds a version of `logsoftmax` which takes `dims` keyword, and `∇logsoftmax` too.  

It also adjusts the implementation for `softmax` from #130 so as not to exponentiate twice; this is about twice as fast, see https://github.com/FluxML/NNlib.jl/issues/77#issuecomment-527090413 .

Both of these mutate their `out` array, which could easily be avoided if desired, at the cost of twice the memory.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -68,7 +68,7 @@ end
 way than directly taking the log of the softmax function, which is commonly used in
 computing cross entropy loss.
 """
-function logsoftmax(xs::AbstractArray{T}; dims=1) where {T}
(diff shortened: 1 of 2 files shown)


MikeInnes:
It would be best to avoid in-place operations, since these are designed to be AD-compatible. Otherwise it looks good.

mcabbott:
OK, done.

MikeInnes:
Perfect, thanks a lot!

PR #143, opened 2019-11-14
Add compat section to Project.toml and rm REQUIRE

dhairyagandhi96:


Project.toml
```diff
@@ -9,6 +9,11 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
+[compat]
+julia = "0.7, 1"
```
MikeInnes: The require file asks for Julia 1.0; are we sure this works on 0.7?

dhairyagandhi96: Ah, yeah, I will remove that

PR #145, opened 2019-11-18
Added Mish activation

digantamisra98:
Referring to issue - [144](https://github.com/FluxML/NNlib.jl/issues/144)

### DIFF
--- a/test/activation.jl
+++ b/test/activation.jl
@@ -135,6 +135,7 @@ end
         @test mish(9) == 9*tanh(log(1 + exp(9)))
         xs = Float32[1 2 3; 1000 2000 3000]
         @test typeof(mish.(xs)) == typeof(xs)
+    end
 
     @test leakyrelu( 0.4,0.3) ≈  0.4
     @test leakyrelu(-0.4,0.3) ≈ -0.12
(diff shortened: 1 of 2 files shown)


bhvieira:
This would be nice to have, +1!

staticfloat:
Looks good to me; can you add some simple tests, (take a look at the tests in https://github.com/FluxML/NNlib.jl/blob/master/test/activation.jl#L123-L141 for a simple example) so that we can make sure we don't accidentally break it in the future?

digantamisra98:
@staticfloat Hi, thanks for the review. Have added the tests.

test/activation.jl
```diff
@@ -127,6 +130,12 @@ end
         @test elu(-4) ≈ (exp(-4) - 1)
     end
 
+    @testset "mish" begin
+        @test mish(-5) ≈ -0.033576237730161704
+        @test mish(9) == 9*tanh(log(1 + exp(9)))
+        xs = Float32[1 2 3; 1000 2000 3000]
+        @test typeof(mish.(xs)) == typeof(xs)
```
staticfloat: I think you're missing an `end` here. :)

digantamisra98: Fixed. Thanks!

staticfloat:
Fantastic!  Thanks!

PR #146, opened 2019-11-30
An attempt to combine dense, depthwise and groupwise conv through `DenseConvDims`

arhik:
With #https://github.com/JuliaGPU/CuArrays.jl/pull/521#issue-346772509 and 
https://github.com/JuliaGPU/CuArrays.jl/pull/523#issue-347194687

We can define depthwise and groupwise convolutions using DenseConvDims (Naming will become confusing then). 

- [ ] I need to make corresponding changes in `direct` and `im2col` backend codes.

@staticfloat Please feel to comment on this. Open to any changes or suggestions. I will do my consider your changes in #142.


staticfloat:
Cool @arhik!  To make it easier for you, I just merged #142 so that you can more easily rebase.

Here are my thoughts:

* We don't need to call this `DenseConvDims` anymore; it's just `ConvDims` at this point.  :)

* It would be helpful to have a succinct description of what to do when `groupcount` is not equal to `1` or `input_channels()`.  E.g. what does `groupcount == 7` look like?  If you can succinctly describe what should happen, I can help you adjust the implementations of the direct and im2col implementations.

* Can we make use of NNPACK at all?

* We're going to need some more tests.  :)

src/conv.jl
```diff
@@ -130,28 +123,17 @@ for backend in (Symbol(), :_direct, :_im2col)
...
-    # We do the conv/depthwiseconv filter backprops separately, as the shape calculation
-    # for `w` is slightly different for depthwise than for normal dense convolution.
+    # This filter back prop covers dense/depthwise/groupwise conv filter backprops, as groupcount alone 
+    # is a deciding factor from cudnn's perspective. For backends im2col and direct needs to be handled.
     @eval begin
         function $(Symbol("∇conv_filter$(backend)"))(
                         x::AbstractArray{xT,N}, dy::AbstractArray{yT,N},
                         cdims::ConvDims; kwargs...) where {xT, yT, N}
-            dw = similar(dy, kernel_size(cdims)..., channels_in(cdims),
+            dw = similar(dy, kernel_size(cdims)..., div(channels_in(cdims),group_count(cdims)),
```
arhik: @staticfloat. This is only change. Only weights dimensions shrink as in here by groupcount value in third axis. if groupcount == 1 Nothing changes. When groupcount == 2 (lets say), then one group of weights operate only on the half of the input channels. and produce only one output channel. These weights have to be shared across all the channels so we will have repeat untill it matches output number of channels.

arhik: When groupcount == 7; we should make sure input channels can be divided exactly into 7 groups. Then we should check if output channels are multiple of groupcount(Since one group can only produce one output).

arhik: We should then distribute these 7 groups to operate on different input channels in blocks of div(input_channels(), 7)

ToucheSir:
Can this PR be moved to https://github.com/FluxML/NNlibCUDA.jl, or has it diverged too much to be workable there?

PR #148, opened 2019-12-17
handle dims properly in gradient of logsoftmax

tanhevg:
Original implementation was not passing `dims` to `softmax`

I picked this up when testing gradients for FluxML/Zygote.jl#425


MikeInnes:
:+1: Thanks!

PR #149, opened 2019-12-18
Preserve the type in differentiation

matsueushi:
The current implementations of `leakyrelu`, `elu` and `selu` return `Float64` gradients for `Float32` inputs. 

```julia
julia> using NNlib, Zygote

julia> leakyrelu'(1f0)
1.0

julia> leakyrelu'(-1f0)
0.009999999776482582

julia> elu'(1f0)
1.0

julia> elu'(-1f0)
0.3678794503211975

julia> selu'(1f0)
1.0507010221481323

julia> selu'(-1f0)
0.6467686295509338
```

cf. https://github.com/FluxML/Flux.jl/issues/963. This PR is intended to preserve float precision for differentiation.

```julia
using NNlib, Zygote, Test

ACTIVATION_FUNCTIONS = [σ, relu, leakyrelu, elu, gelu, swish, selu, softplus, softsign, logcosh];
function test_deliv_float_precision_preserving(a)
    @testset "$(a): " begin
        for T in [Float32, Float64]
            for val in [-10, -1, 0, 1, 10]
                val = @inferred a'(T(val))
                @test typeof(val) == T
            end
        end
    end
end

@testset "Float derivative inference" begin
    test_deliv_float_precision_preserving.(ACTIVATION_FUNCTIONS)
end
```

Before
```shell
Test Summary:              | Pass  Fail  Total
Float derivative inference |   85    15    100
  σ:                       |   10           10
  relu:                    |   10           10
  leakyrelu:               |    5     5     10
  elu:                     |    5     5     10
  gelu:                    |   10           10
  swish:                   |   10           10
  selu:                    |    5     5     10
  softplus:                |   10           10
  softsign:                |   10           10
  logcosh:                 |   10           10
ERROR: Some tests did not pass: 85 passed, 15 failed, 0 errored, 0 broken.
```

After
```shell
Test Summary:              | Pass  Total
Float derivative inference |  100    100
```

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -51,7 +51,7 @@ Leaky [Rectified Linear Unit](https://en.wikipedia.org/wiki/Rectifier_(neural_ne
 activation function.
 You can also specify the coefficient explicitly, e.g. `leakyrelu(x, 0.01)`.
 """
-leakyrelu(x::Real, a = oftype(x/1, 0.01)) = max(a*x, x/1)
+leakyrelu(x::Real, a = oftype(x/1, 0.01)) = max(a*x, x/one(x))
 
 
 """
@@ -62,7 +62,7 @@ Exponential Linear Unit activation function.
 See [Fast and Accurate Deep Network Learning by Exponential Linear Units](https://arxiv.org/abs/1511.07289).
 You can also specify the coefficient explicitly, e.g. `elu(x, 1)`.
 """
-elu(x, α = one(x)) = ifelse(x ≥ 0, x/1, α * (exp(x) - one(x)))
+elu(x, α = one(x)) = ifelse(x ≥ 0, x/one(x), α * (exp(x) - one(x)))
 
 
 """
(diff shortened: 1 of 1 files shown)


staticfloat:
Thanks @matsueushi!

dhairyagandhi96:
Should we also add the testset in our regular tests?

matsueushi:
Do I need to add the test to test/activation.jl? Currently NNlib.jl doesn't have Zygote.jl dependency. Or is it better to update the tests of Zygote.jl?

staticfloat:
I would add a type-stability test to NNlib (just ensure that the activation functions don't change the type unnecessarily).

matsueushi:
If you are talking about the values of the activation functions, they are already defined in test/activation.jl. I modified it to test gradients.

https://github.com/FluxML/NNlib.jl/blob/1c35815a2586fcb56e3a28d064cf45f48e4e8df5/test/activation.jl#L5-L14

staticfloat:
Right, so what I mean is that we should have that same test for the gradients of the activation functions. :)

matsueushi:
Thanks, I see what you mean.

PR #154, opened 2019-12-23
Refactor tests and make NNPACK optional

dhairyagandhi96:
This is aimed towards the multiple failures noted in NNlib due to the testing scheme failing to catch broken cases, even when errors were thrown

Relates to #153 and #152 

Additionally the inference tests with the `@inferred` macro have been replaced with checking types from the output, this remains as a TODO to fix inference, and find what is causing inference to fail.

For now, this refactor ensures that tests that have errored (esp. on CI), show up correctly. The difference comes up as `tests/conv.jl` has the skips in place for checking fuzzing tests, returns and further tests aren't marked with failures

cc @staticfloat @ChrisRackauckas @MikeInnes

### DIFF
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -5,7 +5,7 @@ using Requires
 include("dim_helpers.jl")
 
 # NNPACK support
-if get(ENV, "NNLIB_USE_NNPACK", "false")
+if get(ENV, "NNLIB_USE_NNPACK", "false") == "true"
   if Sys.islinux() || Sys.isapple()
       include("nnpack/NNPACK.jl")
   else
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -6,7 +6,8 @@ include("dim_helpers.jl")
 
 # NNPACK support
 if Sys.islinux() || Sys.isapple()
-    include("nnpack/NNPACK.jl")
+    # include("nnpack/NNPACK.jl")
+    is_nnpack_available() = false
 else
     is_nnpack_available() = false
 end
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -5,12 +5,9 @@ using Requires
 include("dim_helpers.jl")
 
 # NNPACK support
-if get(ENV, "NNLIB_USE_NNPACK", "false") == "true"
-  if Sys.islinux() || Sys.isapple()
-      include("nnpack/NNPACK.jl")
-  else
-      is_nnpack_available() = false
-  end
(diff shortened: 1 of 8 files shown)


dhairyagandhi96:
As a test, I might add a dummy function in there to throw and make sure that CI reports a failure.

dhairyagandhi96:
Things look as expected on Travis, AppVeyor needs a bit of hand holding to build

dhairyagandhi96:
Added return type annotations to the NNPACK codebase to help inference along. Earlier it was inferring `Any`, and this should help resolve the type to the expected `Array`.

dhairyagandhi96:
The inference works, but that breaks mixed precision `conv_<filter/data>_im2col`.

dhairyagandhi96:
So this also introduces a switch to turn off NNPACK, which is unset by default. Good for backwards compatibility and fleshing out the implementation.

src/NNlib.jl
```diff
@@ -5,10 +5,14 @@ using Requires
 include("dim_helpers.jl")
 
 # NNPACK support
-if Sys.islinux() || Sys.isapple()
-    include("nnpack/NNPACK.jl")
+if get(ENV, "NNLIB_USE_NNPACK", "false") == "true"
+  if Sys.islinux() || Sys.isapple()
```
KristofferC: This is then something that needs to be set during precompile time. And if you want to change it, how do you cause the package to be recompiled for it to take affect?

dhairyagandhi96: Good catch, maybe I should move this to an `__init__` function.

KristofferC: That has some other negatives in the sense that the NNPack code will no longer be preconpilable. One way would probably be to not use dispatch to select the backend code but instead have something like

```
function conv(...)
    if get(ENV(...))
         _NNPack_conv(...)
    else
        generic_conv(...)
    end
end
```

But that would require some restructuring of code that might be a bit annoying. 

The easiest is probably to move to use a build step. The build file looks at the ENV variable and creates a file that includes the NNPack depending on the env variable. If one wants to change this option, one sets the nrw env var and the runs `build` on the package.

dhairyagandhi96: You mean doing this in the `build.jl` file and write to `deps.jl`?

dhairyagandhi96: Could you please take a look? I've made it so.

dhairyagandhi96:
Merging this now, to get the ecosystem up and running. Hopefully it doesn't complicate debugging and testing too much @staticfloat ?

PR #155, opened 2019-12-30
conv.jl: fix @debug message

bcdarwin:


bcdarwin:
Also, perhaps this should be higher than `debug`?

Issue #158, opened 2020-01-14
Conv gives incorrect results when multithreading


oxinabox:
using lastet released v0.6.3 on julia 1.3.
without NNPack
```
julia> NNlib.is_nnpack_available()
false

julia> versioninfo()
Julia Version 1.3.2-pre.0
Commit 2e6715c045 (2019-12-31 00:49 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.0.0)
  CPU: Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

The following code works fine if I have not set threads.
But the assertion fails, if I have say `export JULIA_NUM_THREADS=8`.

```
using NNlib
using Random
Random.seed!(1)
x = rand(10, 3, 2)
w = rand(3, 3, 3)
cdims = DenseConvDims(x, w)


r1 = conv(x, w, cdims)
sleep(0.1)
r2 = conv(x, w, cdims)
sleep(0.1)
r3 = conv(x, w, cdims)
sleep(0.1)
r4 = conv(x, w, cdims)

@assert r1 == r2 == r3 == r4
```

andreasnoack:
Ref https://github.com/FluxML/Flux.jl/issues/982#issuecomment-571471224 and https://github.com/FluxML/NNlib.jl/commit/1365788c453f282221b6d6483d25a626f81e8535#r36687517

dhairyagandhi96:
Yup, its exclusively with threading enabled that this happens, looks like a thread safety issue of some sort.

I think the way to get around it would be to chunk our memory per thread?
cc @staticfloat (ref the NNPACK related releases)

PR #159, opened 2020-01-14
test multithreaded

oxinabox:
Since i ran into #158 
I figure good to test multhreaded.

TravisCI only gives us 2 cores so thats what we can do.

This gives us at least  a chance that CI will catch something going wrong this way.


dhairyagandhi96:
Can we continue to test on older Julia versions? Otherwise lgtm

oxinabox:
1.1 isn't LTS, nor current stable.
We could keep it, but why 1.1 and not 1.2?

---

Hmm seems like the env isn't getting caught into the matrix of Travis tests...

dhairyagandhi96:
Ideally would like to include 1.2 in the testing suite, as a sanity check

Maybe try to break it down with GROUP in Travis?

CarloLucibello:
I think testing on the lowest supported julia version and the last one is fine. Also, we are also doubling the computational load of CI here. I'll have CI run again before merging

CarloLucibello:
ops, this should have been merged time ago. Rebase?

CarloLucibello:
this has been done in a recent PR

PR #162, opened 2020-01-30
Update test cases of activation functions

matsueushi:
- Add mish (https://github.com/FluxML/NNlib.jl/pull/145) to the list of the activation functions
- Type stability test for gradients (https://github.com/FluxML/NNlib.jl/pull/149)

I slightly modified the definition of `gelu` because the previous definition didn't pass the gradient test in Julia 1.0 and 1.1. (https://travis-ci.org/matsueushi/NNlib.jl/builds/643663660?utm_medium=notification&utm_source=github_status)

It is related to `pi` and I found a similar issue. https://github.com/JuliaMath/Calculus.jl/issues/144

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ julia = "1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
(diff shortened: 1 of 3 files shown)


staticfloat:
Let's see if CI is happier now that you've upgraded the deps.

matsueushi:
Thanks!

PR #163, opened 2020-01-31
Big interface overhaul

staticfloat:
Eliminate as much runtime checking as possible; create metaprogramming
forwards from frontend to backend in a better (yet still imperfect) manner.

We are reserving the right for the frontend names such as `conv()` to
decide which backend to use at runtime, but for now, all `conv()` calls
(but not all pooling calls) are dispatched at compile-time, which is nice.


matsueushi:
Are we going to drop support for Julia 1.0-1.2? NNPACK_jll requires >= 1.3.
https://github.com/JuliaBinaryWrappers/NNPACK_jll.jl/blob/c49efc824b69448f272665b07513578c25ee1d8c/src/NNPACK_jll.jl#L3-L19

Issue #164, opened 2020-02-02
Potential overflow of tanh


sambitdash:
As described with #78  there is a possibility of tanh getting into the same situation. `tanh` being a division of 2 symmetric functions the possibilities of numerical instability is still higher. I believe the NNlib should define its own `tanh` activation function that overrides the `Base:tanh` like:

Base.tanh(x) = 2σ_stable(x) - one(x)

Similarly, wherever possible all activation functions must be made derivatives of a safe version of `sigmoid`.

sambitdash:
Looks like tanh is safe as normal usage. exp(LARGE) maps to `Inf` leads to valid result. The fixes may be needed for `ForwardDifference` methods only.

sambitdash:
As per comments in #165 the tanh is safe from ForwardDiff so closing.

Issue #78, opened 2018-11-30
Overflow of Float64 logistic sigmoid derivative


ornithos:
I've encountered NaN derivatives when using the logistic sigmoid (σ) for moderate sized negative numbers. I think the boundary is about x=-709.783:

```julia
using Flux: Tracker
using ForwardDiff, NNlib

# Within Flux
x = Tracker.param(-709.783)
Tracker.back!(σ(x))
Tracker.grad(x) # = NaN

# Overflow in Dual Number:
σ(ForwardDiff.Dual(-709.783,1.0))  # Dual{Nothing}(0.0,NaN)
```

since (1 + exp(709.782)) ≈ 1.8 x 10^308. Given that there is a σ_stable function [defined](https://github.com/FluxML/NNlib.jl/blob/20aa7005c2ad2e511b4332a8af7920a2bac7a067/src/activation.jl#L12) presumably for precisely this issue when using Float32, should this be extended to all real numbers?

EDIT: A similar problem occurs in `softplus` at the same (negated) value, which may benefit from setting to the identity above e.g. x=80. (The other activation functions look ok.)

MikeInnes:
Yeah, happy to take patches here. We could do something like `ifelse(x < lower_bound(x), ...` where `lower_bound` dispatches on type.

MJ10:
So essentially there is just one definition
```julia
σ(x) = ifelse(x < lower_bound(x),  zero(x), one(x) / (one(x) + exp(-x)))
```
and we have define lower_bound as
```julia
lower_bound(x::Float32) = ...
lower_bound(x::Float64) = ...
...
```
Is this right or did you mean something else? @MikeInnes

MikeInnes:
Yeah that seems right to me.

MJ10:
```julia
σ(x) = ifelse(x < lower_bound(x), zero(x), one(x) / (one(x) + exp(-x)))

lower_bound(x::Float32) = -80
lower_bound(x::Float64) = -709
```
So this can replace the separate definitions for stable right? I'm not sure how to include the `Flux.Tracker.TrackerReal` types without requiring the package like `ForwardDiff`.

MikeInnes:
Yes, it can replace `σ_stable`. I would probably just do it for `Dual` types for now, since those are used by Flux's broadcast anyway. The fallback should be `typemin`.

MJ10:
```
[...]

PR #165, opened 2020-02-02
Adding ForwardDiff fix for sigmoid.

sambitdash:
This is a sample PR for changes needed for ForwardDiff.Duals. The derivatives can be computed as simple operations on orginal functions than resorting to situations where `Inf`/`Inf` or `Inf`*0  can arise leading to potential NaNs.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ julia = "1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 
 [targets]
-test = ["Test"]
+test = ["Test", "ForwardDiff"]
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -127,7 +127,7 @@ Return `log(cosh(x))` which is computed in a numerically stable way.
 logcosh(x::T) where T = x + softplus(-2x) - log(convert(T, 2))
 
 # Provide an informative error message if activation functions are called with an array
-for f in (:σ, :σ_stable, :logσ, :relu, :leakyrelu, :elu, :gelu, :swish, :selu, :softsign, :softplus, :logcosh)
+for f in (:σ, :logσ, :relu, :leakyrelu, :elu, :gelu, :swish, :selu, :softsign, :softplus, :logcosh)
   @eval $(f)(x::AbstractArray, args...) =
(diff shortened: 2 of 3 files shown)


sambitdash:
A detailed approrach may be worked out for all activation methods as discussed in #164 or #78.

MikeInnes:
Can you elaborate on the changes to ForwardDiff.Dual and why it needs this rewrite? Why can't we continue using `σ_stable`?

sambitdash:
@MikeInnes, 

My proposed solution is not just for sigmoid but any such function where such situation can arise, although it has been found in sigmoid to start with we can look at other potential situations on a case to case basis. 

In Julia if you were to define 1/(1+exp(-x)) it's stable. Julia's infinte maths takes care of cases like `1/Inf` cases and accurately maps to `0`. 

While it's possible in the computation of the ForwardDiff method the generic differentiation computation is not able to address such corner cases and leading to `Inf/Inf` or `0/0` cases which is mapping the data to `NaN`.

The earlier implementation clipped the negative value arbitrarily to -80 which as documented was considered as a hack (as per the comment in the code). The primary intent of such a code would have been valid had Julia not been able to handle `Inf` maths properly. That's not the case here. 

Here the issue is `ForwardDiff` code is not handling the numerical computation of derivatives in an `Inf` aware manner. And the proposed solution was spcific to addressing that issue. Since, derivative of sigmoid can be written as polynomial of sigmoid this can be very easily addressed by rewriting the `Dual`. Similarly, for `tanh` and other functions as well.

MikeInnes:
Ok, I thought you meant there had been recent forwarddiff changes that this was in response to.

Having a cutoff is not completely arbitrary since the point where `NaN`s arise is exactly the point where the result of `σ` is indistinguishable from zero. But analytically simplifying the gradient to avoid that issue entirely obviously seems better than defining cutoffs for all floating point types. So I agree this seems like a good way to go. We shouldn't need it for most functions since things like `tanh` are defined analytically by ForwardDiff already.

sambitdash:
If you are in agreement with these changes, I will add a few test cases as may be needed for this PR and submit specifically for `sigmoid` only.

cossio:
Please see https://github.com/FluxML/NNlib.jl/pull/198 for a simpler fix.

CarloLucibello:
this PR is not needed anymore, thanks in any case

PR #168, opened 2020-02-08
Activation functions added

sriyash421:
added some activation functions and unit tests for the same.
Tests passed for all cases except the gradient check in the softshrink activation function.

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -59,7 +59,7 @@ leakyrelu(x::Real, a = oftype(x / 1, 0.01)) = max(a * x, x / one(x))
 [Rectified Linear Unit](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 activation function.
 """
-relu6(x::Real) = min(relu(x), oftype(x/1, 6.0))
+relu6(x::Real) = min(relu(x), one(x)*6)
 
 """
     rrelu(x) = max(ax, x)
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -59,7 +59,7 @@ leakyrelu(x::Real, a = oftype(x / 1, 0.01)) = max(a * x, x / one(x))
 [Rectified Linear Unit](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 activation function.
 """
-relu6(x::Real) = min(relu(x), one(x)*6)
+relu6(x::Real) = min(relu(x), one(x)*oftype(x, 6))
 
 """
     rrelu(x) = max(ax, x)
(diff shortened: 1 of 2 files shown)


src/activation.jl
```diff
@@ -136,9 +166,24 @@ See [Mish: A Self Regularized Non-Monotonic Neural Activation Function](https://
...
+tanhshrink(x::Real) = x - tanh(x)
+
+"""
+    softshrink = (x ≥ λ ? x-λ : (-λ ≥ x ? x+λ : 0))
+
+See [Softshrink Activation Function](https://www.gabormelli.com/RKB/Softshrink_Activation_Function)
+"""
+function softshrink(x::Real, λ = oftype(x / 1, 0.5))
+    return ifelse(x >= λ, (x - λ) / one(x), ifelse(x <= -λ, (x + λ) / one(x), oftype(x / 1, 0.0)))
+end
```
matsueushi: It can be simplified as 
```julia
softshrink(x::Real, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
```

This definition passes the gradient test
```julia
julia> using Flux, Test, Zygote

julia> softshrink(x::Real, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
softshrink (generic function with 4 methods)

julia> @testset for T in [Float32, Float64]
           for val in [-10, -1, 0, 1, 10]
               val = @inferred softshrink(T(val))
               @test typeof(val) == T
           end
       end
Test Summary: | Pass  Total
T = Float32   |    5      5
Test Summary: | Pass  Total
T = Float64   |    5      5
2-element Array{Any,1}:
 Test.DefaultTestSet("T = Float32", Any[], 5, false)
 Test.DefaultTestSet("T = Float64", Any[], 5, false)

julia> @testset for T in [Float32, Float64]
           for val in [-10, -1, 0, 1, 10]
               val = @inferred softshrink'(T(val))
               @test typeof(val) == T
           end
       end
Test Summary: | Pass  Total
T = Float32   |    5      5
Test Summary: | Pass  Total
T = Float64   |    5      5
2-element Array{Any,1}:
 Test.DefaultTestSet("T = Float32", Any[], 5, false)
 Test.DefaultTestSet("T = Float64", Any[], 5, false)
```

sriyash421: After making the changes, the tests pass for all cases on my local machine but fail for one in he build. Could you please tell me why?

matsueushi: I'm not sure why it happens, but it might be related the operator `<=`. https://github.com/FluxML/Flux.jl/issues/1011

matsueushi: I made a PR https://github.com/FluxML/Zygote.jl/pull/509

staticfloat:
Looks good to me!

PR #170, opened 2020-02-14
docstring, type annotation

matsueushi:
I have cleaned up the docstrings and definitions of the activation functions, and fixed a missing comma in Base.show for ConvDims.

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -59,7 +59,7 @@ leakyrelu(x::Real, a = oftype(x / 1, 0.01)) = max(a * x, x / one(x))
 [Rectified Linear Unit](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 activation function.
 """
-relu6(x::Real) = min(relu(x), one(x)*oftype(x, 6))
+relu6(x::Real) = min(relu(x), oftype(x, 6))
 
 """
     rrelu(x) = max(ax, x)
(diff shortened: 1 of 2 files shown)


src/activation.jl
```diff
@@ -45,7 +45,7 @@ relu(x::Real) = max(zero(x), x)
 
 
 """
-    leakyrelu(x) = max(0.01x, x)
+    leakyrelu(x, a = 0.01) = max(ax, x)
```
CarloLucibello: I suggest using no whitespace in keyword arguments, as this seems to be the stylistic standard in julia
```suggestion
    leakyrelu(x, a=0.01) = max(a*x, x)
```

mcabbott: But those aren't keywords. I thought the style was something like `leakyrelu(x, [a = 0.01]) = max(ax, x)`.

matsueushi: > Optional arguments should be represented with their default values (i.e. f(x, y=1)) when possible, following the actual Julia syntax. Optional arguments which do not have a default value should be put in brackets (i.e. f(x[, y]) and f(x[, y[, z]])). 

https://docs.julialang.org/en/v1/manual/documentation/index.html#Documentation-1

Since our functions have default values, I would follow the style suggested by @CarloLucibello

staticfloat:
Looks good to me!

PR #175, opened 2020-02-24
Added some Activation functions

AdarshKumar712:
Added the following activation functions along with tests

- [x] hardsigmoid
- [x] hardtanh 
- [x] thresholdrelu or trelu
- [x]  lisht 

- hardsigmoid and hardtanh are piecewise linear approximations and computationally effiecient than their counterparts, sigmoid and tanh respectively
- Threshold relu is used in Autoencoders as Threshold gated RELU
- LiSHT: Non-Parametric Linearly Scaled Hyperbolic Tangent Activation Function
  Ref: https://arxiv.org/pdf/1901.05894.pdf

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -1,5 +1,5 @@
-export σ, sigmoid, relu, leakyrelu, relu6, rrelu, elu, gelu, swish, selu, celu, softplus, softsign, logσ,
-       logsigmoid, logcosh, mish, tanhshrink, softshrink, thresholdrelu, trelu, lisht, hardσ, hardsigmoid
+export σ, sigmoid, hardσ, hardsigmoid, hardtanh, relu, leakyrelu, relu6, rrelu, elu, gelu, swish, selu, celu, softplus, softsign, logσ,
+       logsigmoid, logcosh, mish, tanhshrink, softshrink, thresholdrelu, trelu, lisht
 
 """
     σ(x) = 1 / (1 + exp(-x))
@@ -18,13 +18,14 @@ const sigmoid = σ
 end
 
 """
-   hardσ(x) = x > 2.5 ? 1.0 : x < -2.5 ? 0 : 0.2 * x + 0.5
-Segment-wise linear approximation of sigmoid[HardSigmoid]()
+   hardσ(x) = max(0, min(1.0, 0.2 * x + 0.5))
 
-Note: It should not be use with Regression tasks.
+Segment-wise linear approximation of sigmoid
+Note: It should not be used with Regression tasks.
 """
-hardσ(x::Real) = oftype(x,x>2.5 ? 1.0 : x<-2.5 ? 0 : 0.2*x + 0.5)
+hardσ(x::Real) = oftype(x, max(0, min(1.0, 0.2 * x + 0.5)))
 const hardsigmoid = hardσ
+
 
 """
     logσ(x)
@@ -42,6 +43,12 @@ Return `log(σ(x))` which is computed in a numerically stable way.
 logσ(x::Real) = -softplus(-x)
 const logsigmoid = logσ
 
+"""
+    hardtanh(x) = max(-1, min(1, x))
+
+Segment-wise linear approximation of tanh. Cheaper  and  more  computational  efficient version of tanh.
+"""
+hardtanh(x::Real) = oftype(x,max(-1, min( 1, x)))
 
 """
     relu(x) = max(0, x)
@@ -154,6 +161,7 @@ function celu(x::Real, α::Real = one(x))
     return ifelse(x ≥ 0, x / one(x), α * (exp(x/α) - one(x)))
 end 
 
+
 """
     trelu(x, theta = 1.0) = x > theta ? x : 0 
 
@@ -162,6 +170,7 @@ See [ThresholdRelu](https://arxiv.org/pdf/1402.3337.pdf)
 """
 trelu(x::Real,theta = one(x)) = x > theta ? x : zero(x) 
 const thresholdrelu = trelu
+
 
 """
     softsign(x) = x / (1 + |x|)
@@ -211,7 +220,7 @@ See [Softshrink Activation Function](https://www.gabormelli.com/RKB/Softshrink_A
 softshrink(x::Real, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
 
 # Provide an informative error message if activation functions are called with an array
-for f in (:σ, :σ_stable, :hardσ, :logσ, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
+for f in (:σ, :σ_stable, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
     @eval $(f)(x::AbstractArray, args...) =
       error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
 end
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -160,9 +160,7 @@ end
 Continuously Differentiable Exponential Linear Units
 See [Continuously Differentiable Exponential Linear Units](https://arxiv.org/pdf/1704.07483.pdf).
 """
-function celu(x::Real, α::Real = one(x))
-    return ifelse(x ≥ 0, x / one(x), α * (exp(x/α) - one(x)))
-end 
+celu(x::Real, α::Real = one(x)) = ifelse(x ≥ 0, x / one(x), α * (exp(x/α) - one(x))) 
 
 
 """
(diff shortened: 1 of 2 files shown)


test/activation.jl
```diff
@@ -101,7 +113,7 @@ end
     end
 
     @testset "Test Integer64 and Integer32 inputs will force Float64 outputs" begin
-        test_value_int_input_forces_float64.(filter(x -> (x != relu && x != relu6), ACTIVATION_FUNCTIONS))
+        test_value_int_input_forces_float64.(filter(x -> (x != relu && x != relu6 ), ACTIVATION_FUNCTIONS))
```
staticfloat: I think this extra space is a mistake.

src/activation.jl
```diff
@@ -17,6 +17,15 @@ const sigmoid = σ
   σ(x::ForwardDiff.Dual{T,Float32}) where T = σ_stable(x)
 end
 
+"""
+   hardσ(x) = max(0, min(1.0, 0.2 * x + 0.5))
+
+Segment-wise linear approximation of sigmoid
+Note: It should not be used with Regression tasks.
```
staticfloat: Can you give a reason why you've added this note?  Is there a paper that explains this?

src/activation.jl
```diff
@@ -35,6 +44,14 @@ logσ(x::Real) = -softplus(-x)
 const logsigmoid = logσ
 
 
+"""
+    hardtanh(x) = max(-1, min(1, x))
+
+Segment-wise linear approximation of tanh. Cheaper  and  more  computational  efficient version of tanh.
+"""
+hardtanh(x::Real) = oftype(x / 1,max(-1, min( 1, x)))
```
staticfloat: Is there a paper that introduced `hardtanh()`?

test/activation.jl
AdarshKumar712: Oh! Yes, That was left there by mistake.

src/activation.jl
AdarshKumar712: I am not completely sure that if the following paper introduced this, but they have used this in NLP
http://www.jmlr.org/papers/volume12/collobert11a/collobert11a.pdf

AdarshKumar712: And they have claimed that this function is computationally cheaper than tanh itself.

src/activation.jl
AdarshKumar712: I have added this note because the errors will be higher in case of Regression tasks while using piecewise approximations and also it has been observed to get better results on DL Binary classification tasks.

test/activation.jl
AdarshKumar712: I will remove this

src/activation.jl
AdarshKumar712: Another thing is that here I have implemented it the way it has been implemented in Tensorflow as the coefficient of x is 0.2. In original paper, it was taken as 0.5.(https://arxiv.org/pdf/1511.00363.pdf). Is that ok?

test/activation.jl
```diff
@@ -1,6 +1,6 @@
 using NNlib, Test, Zygote
 
-ACTIVATION_FUNCTIONS = [σ, relu, leakyrelu, relu6, rrelu, elu, gelu, celu, swish, selu, softplus, softsign, logcosh, mish, tanhshrink, softshrink];
+ACTIVATION_FUNCTIONS = [σ,hardσ, hardtanh, relu, leakyrelu, relu6, rrelu, elu, gelu, celu, swish, lisht, selu, trelu, softplus, softsign, logcosh, mish, tanhshrink, softshrink];
```
matsueushi:
```suggestion
ACTIVATION_FUNCTIONS = [σ, hardσ, hardtanh, relu, leakyrelu, relu6, rrelu, elu, gelu, celu, swish, lisht, selu, trelu, softplus, softsign, logcosh, mish, tanhshrink, softshrink];
```
I think a space is missing

src/activation.jl
matsueushi:
```suggestion
hardtanh(x::Real) = max(-one(x), min(one(x), x))
```
I think `hardtanh` doesn't have to force floating point outputs. Our `relu` and `relu6` return integers for integer inputs.

AdarshKumar712: Actually I am unsure about how to decide on that. Though without forcing floating point output, all tests are passing(making this as an exception for int to forced floating point test). So, I guess this doesn't have to force floating point outputs. Also, I think the same goes for trelu as well. If that is so, I will make changes in both. Thanks for pointing that out.
Other than unwanted computation cost, are there any other reasons why we ensure such typecastings?

matsueushi: I'm not sure the background behind it, but this test 
https://github.com/FluxML/NNlib.jl/blob/c82a76d40cf5c34e4a1eed2655aa30b464e3e408/test/activation.jl#L106-L110

was added in the commit https://github.com/FluxML/NNlib.jl/commit/b698441cf0dfbf08875fa3364b44d95bad4ba104 by @staticfloat .

src/activation.jl
staticfloat: It should probably be a parameter with a default value (the TF value is fine, I think).

src/activation.jl
staticfloat: Oh, it's definitely cheaper, I can tell you that.  ;)

Reading that paper, they state that the `HardTanh` was first introduced in `(Collobert, 2004)` which turns out to be the same author's thesis: http://ronan.collobert.org/pub/matos/2004_phdthesis_lip6.pdf

So we'll just cite that as the originating paper.  :)

staticfloat: Unless we have a reason to force floating-point (e.g. for gradients) I prefer to keep datatypes as they are.

src/activation.jl
AdarshKumar712: Ok, I will change it to a parameter

src/activation.jl
AdarshKumar712: Ok, I will add that with the function

AdarshKumar712: I will the unwanted typecasting then

test/activation.jl
AdarshKumar712: I will add that

AdarshKumar712:
@staticfloat @matsueushi I have made the changes. Please have a look.

src/activation.jl
staticfloat: I don't think we should include that note, at least not without a publication explaining why.  I don't think we have enough proof to put that strong of an assertion into the source code.

test/activation.jl
```diff
@@ -114,7 +126,18 @@ end
             @test typeof(relu6(Int64(1))) == Int64
             @test typeof(relu6(Int32(1))) == Int32
         end
-
+        
+        @testset "hardtanh: " begin
+            # relu6 doesn't have to force floating point outputs
```
matsueushi:
```suggestion
            # hardtanh doesn't have to force floating point outputs
```

test/activation.jl
```diff
@@ -114,7 +126,18 @@ end
...
+            @test typeof(hardtanh(Int64(1))) == Int64
+            @test typeof(hardtanh(Int32(1))) == Int32
+        end
+        
+        @testset "trelu: " begin
+            # relu6 doesn't have to force floating point outputs
```
matsueushi:
```suggestion
            # trelu doesn't have to force floating point outputs
```

src/activation.jl
```diff
@@ -136,6 +166,19 @@ function celu(x::Real, α::Real = one(x))
...
+
+"""
+    trelu(x, theta = 1.0) = x > theta ? x : 0 
+
+Threshold Gated Rectified Linear   
+See [ThresholdRelu](https://arxiv.org/pdf/1402.3337.pdf)
+"""
+function trelu(x::Real,theta = one(x)) 
+    return ifelse(x> theta, x/ one(x), zero(x))
+end
```
matsueushi: Since we don't need to force floating point outputs, we can write just as
```julia
trelu(x::Real, theta=one(x)) = ifelse(x > theta, x, zero(x))
```

src/activation.jl
AdarshKumar712: Ok, I will remove that

AdarshKumar712:
Made all the changes asked above

matsueushi:
Looks good to me!

src/activation.jl
```diff
@@ -17,6 +17,15 @@ const sigmoid = σ
   σ(x::ForwardDiff.Dual{T,Float32}) where T = σ_stable(x)
 end
 
+"""
+   hardσ(x, a=0.2) = max(0, min(1.0, a * x + 0.5))
+
+Segment-wise linear approximation of sigmoid
+See: [BinaryConnect: Training Deep Neural Networks withbinary weights during propagations](https://arxiv.org/pdf/1511.00363.pdf)
+"""
+hardσ(x::Real, a=0.2) = oftype(x/1, max(zero(x/1), min(one(x/1), oftype(x/1,a) * x + oftype(x/1,0.5))))
```
CarloLucibello: I think division by `1` should be replaced by division by `one(x)` like fo other activations

AdarshKumar712: Actually one(x) or 1 doesn't make much difference here. In both the cases output precision remains the same. Moreover, division by is faster than division by one(x)
```
julia> x = Float32(1)
1.0f0

julia> @time x/one(x)
  0.002585 seconds (28 allocations: 1.875 KiB)
1.0f0

julia> @time x/1
  0.000009 seconds (5 allocations: 176 bytes)
1.0f0
```

PR #180, opened 2020-03-01
support FillArrays in convolutions

CarloLucibello:
Hotfix for https://github.com/FluxML/Flux.jl/issues/1055.
Those methods should go into FillArrays really, will open an issue there, but in the meanwhile let's fix this issue here since it can come up easily while doing simple tests on Flux conv layers


CarloLucibello:
will fix this in Zygote using the suggestions from https://github.com/JuliaArrays/FillArrays.jl/issues/88

MikeInnes:
It seems to me that the fix should be in NNlib, not Zygote. NNlib should be able to accept generic array types and Zygote shouldn't have to worry about what types it is differentiating.

PR #181, opened 2020-03-02
Add fallback batched_mul!

mcabbott:
This adds a simple fallback method for `batched_mul!`. 

It also narrows the signature of the methods from #100. I am not entirely sure that `StridedArray` is correct here, perhaps it should be `Array`? But for instance `using FillArrays; batched_mul(rand(3,3,3), Fill(1.0, 3,3,3))` used to be accepted, and give an error.

### DIFF
--- a/src/batched/batchedadjtrans.jl
+++ b/src/batched/batchedadjtrans.jl
@@ -1,35 +1,38 @@
 using LinearAlgebra
 import Base: -
 
-"""
+_batched_doc = """
+    batched_transpose(A::AbstractArray{T,3})
+    batched_adjoint(A)
+
+Equivalent to applying `transpose` or `adjoint` to each matrix `A[:,:,k]`.
+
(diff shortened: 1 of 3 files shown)


CarloLucibello:
why would this fail on nigthly?

mcabbott:
See if this helps, testing `==` may have picked up some numerical error, although I've no idea what changed.

CarloLucibello:
since you are at this, could also add a docstring to batched_adjoint and transpose, so that we can add a reference in flux docs?

CarloLucibello:
pff, still failing, this is annoying

mcabbott:
Sorry, I missed a few. And, sure, see what you think of this docstring.

mcabbott:
Don't merge yet, I found that `batched_mul(rand(1:99, 2,2,2), rand(2,2,2))` is an error, should promote.

CarloLucibello:
good to go?

PR #182, opened 2020-03-02
update travis

CarloLucibello:


CarloLucibello:
@staticfloat should we do this? If yes, could you deactivate appveyor for this repo?

CarloLucibello:
reconsider removing appveyor later, now I want to tag this and then flux

CarloLucibello:
omg, this took forever, we should remove appveyor

PR #183, opened 2020-03-05
Add comment for wrapper functions in CuArrays.jl 

matsueushi:
Some of activation functions defined in NNlib are incomplatible with GPU and they need `@cufunc` wrappers, which are defined in CuArrays.jl. https://github.com/JuliaGPU/CuArrays.jl/issues/614 

I found some activation function we added recently (e.g. `celu`) are GPU-unfriendly. I already submitted a PR https://github.com/JuliaGPU/CuArrays.jl/pull/615 to fix this issue, but I think it is better for us to consider GPU compatibility when we add a new function or update its definition  in future.


CarloLucibello:
thanks. Yes we should be more careful when adding new activations

Issue #176, opened 2020-02-26
Leaky relu failing tests


AdarshKumar712:
When I am trying to check for gradient float preserving tests seperately, it is failing tests:
```
julia> using Test

julia> using Zygote

julia> using NNlib

julia> typeof(leakyrelu'(Float32(0.5)))
Float64

julia> function test_gradient_float_precision_preserving(a)
           @testset "$(a): " begin
               for T in [Float32, Float64]
                   for val in [-10, -1, 0, 1, 10]
                       val = @inferred a'(T(val))
                       @test typeof(val) == T
                   end
               end
           end
       end
test_gradient_float_precision_preserving (generic function with 1 method)

julia> test_gradient_float_precision_preserving(leakyrelu)
leakyrelu: : Test Failed at REPL[5]:6
  Expression: typeof(val) == T
   Evaluated: Float64 == Float32
Stacktrace:
 [1] macro expansion at ./REPL[5]:6 [inlined]
 [2] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107 [inlined]
 [3] test_gradient_float_precision_preserving(::Function) at ./REPL[5]:3
leakyrelu: : Test Failed at REPL[5]:6
  Expression: typeof(val) == T
   Evaluated: Float64 == Float32
Stacktrace:
 [1] macro expansion at ./REPL[5]:6 [inlined]
 [2] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107 [inlined]
 [3] test_gradient_float_precision_preserving(::Function) at ./REPL[5]:3
leakyrelu: : Test Failed at REPL[5]:6
  Expression: typeof(val) == T
   Evaluated: Float64 == Float32
Stacktrace:
 [1] macro expansion at ./REPL[5]:6 [inlined]
 [2] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107 [inlined]
 [3] test_gradient_float_precision_preserving(::Function) at ./REPL[5]:3
leakyrelu: : Test Failed at REPL[5]:6
  Expression: typeof(val) == T
   Evaluated: Float64 == Float32
Stacktrace:
 [1] macro expansion at ./REPL[5]:6 [inlined]
[...]

PR #184, opened 2020-03-07
Doc corrections and unicode(alpha,beta,theta)

AdarshKumar712:
I have made some minor doc corrections and replaced alpha, beta, theta with their unicode chars.
I have also updated the leakyrelu activation function to address #176 , and replaced `x / one(x)` with `x / 1`. Please have a look at these changes.

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -23,9 +23,8 @@ end
 Segment-wise linear approximation of sigmoid
 See: [BinaryConnect: Training Deep Neural Networks withbinary weights during propagations](https://arxiv.org/pdf/1511.00363.pdf)
 """
-hardσ(x::Real, a=0.2) = oftype(x/1, max(zero(x/1), min(one(x/1), oftype(x/1,a) * x + oftype(x/1,0.5))))
+hardσ(x::Real, a=0.2) = oftype(x / 1, max(zero(x / 1), min(one(x / 1), oftype(x / 1, a) * x + oftype(x / 1, 0.5))))
 const hardsigmoid = hardσ
-
 
 """
     logσ(x)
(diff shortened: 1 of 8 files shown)


src/impl/conv_im2col.jl
```diff
@@ -11,13 +11,13 @@ end
 end
 
 """
-    conv_im2col!(y, x, w, cdims, col=similar(x); alpha=1, beta=0)
+    conv_im2col!(y, x, w, cdims, col=similar(x); α=1, β=0)
```
CarloLucibello: is this exported or used in flux? we should not break the interface

AdarshKumar712: I searched for this in the flux source files, but didn't found any function where this or other functions changed here, use alpha or beta parameters. However, flux re-export NNlib in its main file. I hope there shouldn't be any problem with Flux related interface.

AdarshKumar712:
@CarloLucibello can you please review these changes? I have removed the Unicode for alpha and beta as they may break the interface for some of the dependent packages, as you pointed out earlier.
Thanks!

src/activation.jl
```diff
@@ -155,53 +147,47 @@ See [Self-Normalizing Neural Networks](https://arxiv.org/pdf/1706.02515.pdf).
...
 
 """
-    trelu(x, theta = 1.0) = x > theta ? x : 0 
+    trelu(x, θ=1.0) = x > θ ? x : 0 
 
 Threshold Gated Rectified Linear.
 See [ThresholdRelu](https://arxiv.org/pdf/1402.3337.pdf)
 """
-trelu(x::Real,theta = one(x)) = ifelse(x> theta, x, zero(x))
+trelu(x::Real,θ = one(x)) = ifelse(x> θ, x, zero(x))
```
CarloLucibello, line 167:
```suggestion
trelu(x::Real, θ=one(x)) = ifelse(x> θ, x, zero(x))
```

src/activation.jl
```diff
@@ -155,53 +147,47 @@ See [Self-Normalizing Neural Networks](https://arxiv.org/pdf/1706.02515.pdf).
...
     softplus(x) = log(exp(x) + 1)
 
 See [Deep Sparse Rectifier Neural Networks](http://proceedings.mlr.press/v15/glorot11a/glorot11a.pdf).
 """
 softplus(x::Real) = ifelse(x > 0, x + log1p(exp(-x)), log1p(exp(x)))
 
-
 """
-    logcosh(x)
+    logcosh(x) = x + softplus(-2x) - log(2)
```
CarloLucibello, line 185: the mathematical definition better down in the docstring

AdarshKumar712: So this should be changed for all the existing functions? As in all the rest functions, mathematical definition is written beside the function.

AdarshKumar712: So this should be changed for other functions as well?

AdarshKumar712:
@CarloLucibello I have moved the mathematical definition to docstrings, for all the functions to maintain the uniformity. Please review these changes.

PR #185, opened 2020-03-14
gumbel Softmax added

sriyash421:
Added Gumbel softmax: a gradient estimator that replaces the non-differentiable sample from a categorical distribution with a differentiable sample

Doubt : Do I need to define functions for backward gradients as well?

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -1,5 +1,6 @@
+using Random
 export softmax, softmax!, ∇softmax, ∇softmax!,
-       logsoftmax, logsoftmax!, ∇logsoftmax, ∇logsoftmax!
+       logsoftmax, logsoftmax!, ∇logsoftmax, ∇logsoftmax!, gumbel_softmax
 
 """
     softmax(x; dims=1)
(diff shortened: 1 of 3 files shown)


src/softmax.jl
```diff
@@ -110,3 +111,34 @@ end
 
 ∇logsoftmax(Δ, xs; dims=1) = Δ .- sum(Δ, dims=dims) .* softmax(xs, dims=dims)
 ∇logsoftmax!(Δ, xs) = ∇softmax!(Δ, Δ, xs)
+
+"""
+gumbel_softmax(x; dims=1, tau=1, hard=false)
```
CarloLucibello: 4 space indenti here

src/softmax.jl
```diff
@@ -110,3 +111,34 @@ end
...
+with each column independent. Keyword `dims=2` will instead treat rows independently, 
+etc..."
+
+τ is temperature that controls how close to max it is
+
+If ``hard=True``, the returned samples will be one-hot
+"""
+function gumbel_softmax(xs::AbstractVecOrMat; dims=1, tau=1, hard=false)
+    rng = MersenneTwister(1234);
+    temp = rand!(rng,zeros(size(xs)))
```
CarloLucibello: you shouldn't sample for an rng alwat initializide with the same seed. 
Also, make sure this will work on both cpu and gpu

sriyash421: Thanks for reviewing it. I will work on the changes immediately.

ToucheSir:
Hey @sriyash421, would you still be interested on working on this? I think there's a basis here, the only 2 remaining pieces being the use of an `rng` argument (rather than hard-coding a `RandomDevice`) and Zygote compatibility (i.e. removing the mutation).

CarloLucibello:
closing as outdated

PR #187, opened 2020-03-22
Improvements to batched_mul, including PermutedDimsArray

mcabbott:
I think that we can allow `batched_gemm!` to be called on many (but not all) `PermutedDimsArray`s, and this is generally much faster than first calling `permutedims`. This PR is an attempt to implement this. It also extends `batched_mul!` to take `α, β` scales like `mul!`.

EARLIER:
* It adds methods to treat `PermutedDimsArray{<:Number,3,(2,1,3)}` as equivalent to `BatchedTranspose` (see https://github.com/FluxML/Zygote.jl/issues/552) and to allow `batched_adjoint ∘ batched_transpose` to be trivial on real-valued arrays.
* It deletes the method for `copy(::BatchedAdjoint)` etc, from #100, to return an `Array` like `copy(::Adjoint)`.

### DIFF
--- a/src/batched/batchedmul.jl
+++ b/src/batched/batchedmul.jl
@@ -5,7 +5,8 @@ using LinearAlgebra: BlasFloat, BlasReal
 
 using Base: promote_typejoin
 
-using ArrayLayouts: MemoryLayout, FirstMajor, SecondMajor, ConjLayout, StridedLayout, UnknownLayout
+using ArrayLayouts: MemoryLayout, UnitStride, AbstractColumnMajor, ConjLayout, StridedLayout, UnknownLayout
+const UnitStrideFirst = Union{UnitStride{1}, AbstractColumnMajor}
 
 include("./batchedadjtrans.jl")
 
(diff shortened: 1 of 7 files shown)


src/batched/batchedmul.jl
```diff
@@ -32,13 +35,13 @@ _unbatch(A::BatchedAdjOrTrans) = A.parent
 const _GemmFloat = Union{Float64, Float32, ComplexF64, ComplexF32}
 
 _BATCHED_GEMM_LIST = [
-    (:(StridedArray{T, 3}), 'N'),
-    (:(BatchedTranspose{T, <:StridedArray{T, 3}}), 'T'),
-    (:(BatchedAdjoint{T, <:StridedArray{T, 3}}), 'C')
+    (:(Array{T, 3}), 'N'),
+    (:(BatchedTranspose{T, <:Array{T, 3}}), 'T'),
+    (:(BatchedAdjoint{T, <:Array{T, 3}}), 'C')
```
CarloLucibello: why do we have to restrict these? Were they causing ambiguities with the cuarrays ones from https://github.com/JuliaGPU/CuArrays.jl/pull/619? the latter seem stricter than the stridedarray versions here....

mcabbott: Maybe there's no need. Does `StridedArray` sound right to you? 

In the opposite direction, I was wondering whether anything more general should be accepted. Here is one where `batched_gemm!` is happy to try but gives the wrong answer, I think:
```julia
julia> aa = rand(2,2,2); bb = rand(2,2,2);

julia> cc = PermutedDimsArray(bb, (1,3,2));

julia> strides(cc) # starts with 1
(1, 4, 2)

julia> dd = NNlib.batched_gemm!('N', 'N', 1.0, aa, cc, 0.0, similar(aa)) # runs
2×2×2 Array{Float64,3}:
[:, :, 1] =
 0.124876  0.271354
 0.324212  0.226239

[:, :, 2] =
 0.458338  2.53924e-315
 0.270784  1.06338e-314

julia> dd ≈ batched_mul(aa, cc) # fallback disagrees
┌ Debug: calling fallback method for batched_mul!
│   typeof(A) = Array{Float64,3}
│   typeof(B) = PermutedDimsArray{Float64,3,(1, 3, 2),(1, 3, 2),Array{Float64,3}}
│   typeof(C) = Array{Float64,3}
└ @ NNlib ~/.julia/dev/NNlib/src/batched/batchedmul.jl:66
false
```

mcabbott: I've now fixed the above error with strides of `batched_gemm!`.

mcabbott:
This has changed a lot. The current form uses https://github.com/JuliaMatrices/ArrayLayouts.jl to keep track dimensions & strides, via traits `UnitStrideFirst` and `UnitStride{2}`, added there in https://github.com/JuliaMatrices/ArrayLayouts.jl/pull/12 (which not yet merged).

See also #191 for an alternative approach. 

Either of these can be made to work with https://github.com/JuliaGPU/CuArrays.jl/pull/664, which will similarly allow permutations.

mcabbott:
This could possibly now be done with https://github.com/SciML/ArrayInterface.jl instead of https://github.com/JuliaMatrices/ArrayLayouts.jl . But I'm hesitant to add dependencies (which would probably need to be added to CUDA.jl too), and I think #191 is simpler lower-tech solution.

PR #191, opened 2020-04-03
Allow batched_mul to work through PermutedDimsArray, II

mcabbott:
This is an alternative to #187. 

It similarly allows `batched_mul` to work on many `PermutedDimsArray`s, but does this just by calling `strides(A)` and branching. It also extends `batched_mul!` to take `α, β` scales like `mul!`.

It adds two functions `storage_type` and `is_strided`, both of which recursively unwrap things. This avoids trying to dispatch on `BatchedAdjoint{PermutedDimsArray{...,CuArray}}`... instead it can separately check the underlying storage, and whether it should be safe to call `strides(A)`.

It also improves `batched_gemm!` to multi-thread the batch index (using https://github.com/JuliaLang/julia/pull/36360 to save & restore the number of BLAS threads), and to allow `size(A,3)==1` (batch only B,C).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-Compat = "3.13"
+Compat = "3.14"
 Requires = "0.5, 1.0"
 julia = "1.3"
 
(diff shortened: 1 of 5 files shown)


mcabbott:
Bump. Who has merge permissions here? @CarloLucibello,  @DhairyaLGandhi?

CarloLucibello:
ops, didn't see this, thanks

mcabbott:
Thanks!

Issue #188, opened 2020-03-29
gradient causes a 'MethodError: no method matching ∇maxpool' error for NNlib.maxpool


nirmal-suthar:
MWE : 
(same for meanpool)

```
julia> using Flux
julia> m = Chain(Conv((1,),2=>4),x -> maxpool(x, (4,), pad=2, stride=2))
Chain(Conv((1,), 2=>4), #3)

julia> gradient(() -> sum(m(rand(8,2,4))), params(m))
ERROR: MethodError: no method matching ∇maxpool(::FillArrays.Fill{Float64,3,Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}}, ::Array{Float64,3}, ::Array{Float64,3}, ::Tuple{Int64}; pad=2, stride=2)
Closest candidates are:
  ∇maxpool(::AbstractArray{T,N}, ::AbstractArray{T,N}, ::AbstractArray{T,N}, ::PoolDims; kwargs...) where {T, N} at /home/nirmal/.julia/packages/NNlib/FAI3o/src/pooling.jl:123
Stacktrace:
 [1] (::Zygote.var"#1770#1771"{Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol,Symbol},NamedTuple{(:pad, :stride),Tuple{Int64,Int64}}},Array{Float64,3},Tuple{Int64},Array{Float64,3}})(::FillArrays.Fill{Float64,3,Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}}) at /home/nirmal/.julia/packages/Zygote/KNUTW/src/lib/nnlib.jl:72
 [2] (::Zygote.var"#4341#back#1773"{Zygote.var"#1770#1771"{Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol,Symbol},NamedTuple{(:pad, :stride),Tuple{Int64,Int64}}},Array{Float64,3},Tuple{Int64},Array{Float64,3}}})(::FillArrays.Fill{Float64,3,Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}}) at /home/nirmal/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:55
 [3] #3 at ./REPL[4]:1 [inlined]
[...]

PR #193, opened 2020-04-06
Fix 'MethodError: no method matching ∇maxpool' error for NNlib.maxpool

nirmal-suthar:
For [#188 ](https://github.com/FluxML/NNlib.jl/issues/188)

- Added new methods for ∇maxpool and ∇meanpool to accept `k::NTuple{N, Integer}; pad=0, stride=k` instead of `pdims::PoolDims`, which causes error while calling gradient in NNlib.maxpool/NNlib.meanpool

- Added test

### DIFF
--- a/src/pooling.jl
+++ b/src/pooling.jl
@@ -140,6 +140,19 @@ end
 expand(N, i::Tuple) = i
 expand(N, i::Integer) = ntuple(_ -> i, N)
 
+function ∇maxpool(dy, y, x, k::NTuple{N, Integer}; pad=0, stride=k) where {N}
+    pad = expand(Val(N), pad)
+    stride = expand(Val(N), stride)
+    pdims = PoolDims(x, k; padding=pad, stride=stride)
+    ∇maxpool(dy, y, x, pdims)
+end
+
+function ∇meanpool(dy, y, x, k::NTuple{N, Integer}; pad=0, stride=k) where {N}
+    pad = expand(Val(N), pad)
+    stride = expand(Val(N), stride)
+    pdims = PoolDims(x, k; padding=pad, stride=stride)
+    ∇meanpool(dy, y, x, pdims)
+end
 
 """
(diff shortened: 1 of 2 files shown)


CarloLucibello:
I'm sorry, I just realized that probably the correct way to fix this is to specialize the adjoint here 
https://github.com/FluxML/Zygote.jl/blob/9cbf571b3379860f07c16989a3086d9a7d54653a/src/lib/nnlib.jl#L70
to 
```
@adjoint function maxpool(x, pdims::PoolDims; kw...)
```
Then Zygote should be able to automatically differentiate through the other `maxpool` method as well. 
Could you try this alternative approach, and in case it works file a PR over there? In that case,, adding a couple of simple tests there is enough. Sorry again for noticing late.

nirmal-suthar:
> here
>https://github.com/FluxML/Zygote.jl/blob/9cbf571b3379860f07c16989a3086d9a7d54653a/src/lib/nnlib.jl#L70
> to
> 
> ```
> @adjoint function maxpool(x, pdims::PoolDims; kw...)
> ```

Do I need to add an adjoint ` @adjoint function maxpool(x, k::NTuple; pad=0, stride=k)` as there is already an adjoint for `@adjoint function maxpool(x, pdims::PoolDims; kw...)`?

CarloLucibello:
I think that once you specialize on PoolDims you don't need any other adjoint

nirmal-suthar:
> I think that once you specialize on PoolDims you don't need any other adjoint

Yes, adding type-assertion to `pdims` fixes the issue. I have filed a PR [https://github.com/FluxML/Zygote.jl/pull/587](https://github.com/FluxML/Zygote.jl/pull/587).
Thanks :)

CarloLucibello:
ok, so closing this in favor of https://github.com/FluxML/Zygote.jl/pull/587

PR #194, opened 2020-04-14
Mutating softmax tests

sudo-rushil:
Added tests for softmax variants that mutate their inputs. These make sure that the mutating variants are consistent with the default softmax functions.

PS: As a side note,  I noticed that `∇logsoftmax!` was defined as

```julia
∇logsoftmax!(Δ, xs) = ∇softmax!(Δ, Δ, xs)
```

Is there supposed to be an analogous `∇logsoftmax!(Δ, Δ, xs)`, or is this intentional?

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -3,8 +3,8 @@ export softmax, softmax!, ∇softmax, ∇softmax!,
 
 """
     softmax(x; dims=1)
-    
-[Softmax](https://en.wikipedia.org/wiki/Softmax_function) turns input array `x` 
+
+[Softmax](https://en.wikipedia.org/wiki/Softmax_function) turns input array `x`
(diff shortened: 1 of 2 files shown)


CarloLucibello:
that is certainly a bug, we miss an implementation of `∇logsoftmax!(out, Δ, xs)`, similar to the one we have for softmax

sudo-rushil:
Would this have the desired behavior? It is defined analogously to ∇logsoftmax.

```julia
function ∇logsoftmax!(out::AbstractVecOrMat, Δ::AbstractVecOrMat, xs::AbstractVecOrMat)
    out .= Δ .- sum(Δ, dims=1) .* softmax(xs, dims=1)
end
```

CarloLucibello:
nice, thanks!

PR #195, opened 2020-04-20
Piecewise approximations to tanh and sigmoid

AStupidBear:


src/activation.jl
```diff
@@ -225,8 +225,40 @@ See [Softshrink Activation Function](https://www.gabormelli.com/RKB/Softshrink_A
 """
 softshrink(x::Real, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
 
+"""
+    ptanh(x)
+
+Piecewise Approximation of Tanh Function
+
+See [Activation Functions](https://www.gabormelli.com/RKB/Softshrink_Activation_Function).
```
mcabbott, line 233: I think you have the wrong link here. 

And I also think the docstring should explain how good/bad the approximation is, and ideally whether its gradient is continuous, etc.

src/activation.jl
```diff
@@ -225,8 +225,44 @@ See [Softshrink Activation Function](https://www.gabormelli.com/RKB/Softshrink_A
...
+    λ = oftype(x / 1, 0.26037)
+    ifelse(x > xθ, yθ, ifelse(x > o, yθ - λ * (x - xθ)^2, ifelse(x > -xθ, λ * (x + xθ)^2 - yθ, -yθ)))
+end
+
+"""
+    pσ(x)
+
+Piecewise Approximation of Sigmoid Function. 
+
+The maximum deviation between pσ and σ is 0.04.
```
mcabbott, line 251:
```suggestion
Uses 4 quadratic pieces, with absolute error < 0.02, and a continuous first derivative at the junctions.

```

AStupidBear: Thanks

CarloLucibello:
I'm not inclined toward adding more and more exotic activation to Flux. Is there any precedence of this piecewise activations in any other deep learning library? Do we have performance benchmarks?

AStupidBear:
@CarloLucibello Currently `tanh` and `sigmoid` in Base are not SIMD vectorizable. Users can speed up inference on cpu by overloading  `Base.tanh` with `NNlib.ptanh(x)` .

AStupidBear:
Here are some benchmarks.

```
using Random
using LoopVectorization
using BenchmarkTools
using NNlib
ENV["OMP_NUM_THREADS"] = 1

function ptanh(x::Real)
    l = oftype(x / 1, 1)
    o = oftype(x / 1, 0)
    xθ = oftype(x / 1, 1.92033)
    yθ = oftype(x / 1, 0.96016)
    λ = oftype(x / 1, 0.26037)
    ifelse(x > xθ, yθ, ifelse(x > o, yθ - λ * (x - xθ)^2, ifelse(x > -xθ, λ * (x + xθ)^2 - yθ, -yθ)))
end

function pσ(x::Real)
    l = oftype(x / 1, 1)
    o = oftype(x / 1, 0)
    h = oftype(x / 1, 0.5)
    x = x / oftype(x / 1, 4.1)
    ifelse(x > l, l, ifelse(x < -l, o, h + x * (l - h * abs(x))))
end

batch_size = 128
input_size = 64
hidden_size = 64

W = randn(Float32, hidden_size, input_size);
x = randn(Float32, input_size, batch_size);
```
```
julia> @btime ptanh.($x);
  3.529 μs (2 allocations: 32.08 KiB)

julia> @btime tanh.($x);
  191.660 μs (2 allocations: 32.08 KiB)

julia> @btime @avx tanh.($x);
  16.769 μs (2 allocations: 32.08 KiB)

julia> @btime $W * $x;
  15.868 μs (2 allocations: 32.08 KiB)

julia> @btime pσ.($x);
  3.481 μs (2 allocations: 32.08 KiB)

julia> @btime σ.($x);
  105.581 μs (2 allocations: 32.08 KiB)

julia> @btime @avx @. 1 / (1 + exp(-x));
  7.767 μs (10 allocations: 32.23 KiB)

julia> @btime $W * $x;
  15.745 μs (2 allocations: 32.08 KiB)

```

AStupidBear:
@CarloLucibello As you can see, `tanh` and `sigmoid` are much slower than the piecewise and the vectorized version and even slower than matrix multiplicatio, blocking us from doing do inference on cpu efficiently with recurrent layers. Other deep learning libraries don't have this because they either use MKL-DNN, or have SIMD-vectorized versions. 

Therefore, I strongly recommend merging this PR first as an alternative to speedup inference. Vectorized version should be considered later after LoopVectorization can be hooked to Zygote.

CarloLucibello:
Nice. Could you also add benchmarks with Zygote.gradient, for future reference?
CI failure seems unrelated, let's trigger it again

mcabbott:
That's a nice speedup! 

I'm not too sure what the gradient times are measuring, Zygote has a special case for `tanh`, which should cost nothing next to the forward pass:
https://github.com/FluxML/Zygote.jl/blob/master/src/lib/broadcast.jl#L99
```
julia> x = randn(Float32, 64, 128);

julia> @btime tanh.($x);
  163.032 μs (2 allocations: 32.08 KiB)

julia> @btime @fastmath tanh.($x); # doesn't vectorise here
  161.859 μs (2 allocations: 32.08 KiB)

julia> @btime gradient(x -> sum(tanh.(x)), $x);
  166.425 μs (19 allocations: 64.55 KiB)
```

One concern here is that the gradient of these is exactly zero after some point, which is very different to `tanh`'s decay. A simple remedy for this would be to just define them to have the true gradient, since this is not expensive. @AStupidBear do you perhaps you have reasonably real-life examples on which this could be checked?

AStupidBear:
@CarloLucibello Can we overload broadcastings of some activation functions such as `tanhh`/`sigmoid` by incorporating `LoopVectorization.jl`? See https://github.com/FluxML/Flux.jl/pull/1169

AStupidBear:
@mcabbott like this?
```
@adjoint function broadcasted(::typeof(ptanh), x::Numeric)
  y = ptanh.(x)
  y, ȳ -> (nothing, ȳ .* conj.(1 .- y.^2))
end
```

AStupidBear:
abc

mcabbott:
Oh right, that may not work quite as well as I hoped... whether it's better or worse than the actual gradient of `ptanh` seems hard to guess.

AStupidBear:
I usually use the piecewise verision for inference, not for training. But it can train, at least for time series forecasting (although not tested on lanugage models). Here is an example to train a cosine wave (like neural ode). This must be run with Flux.jl#tracker, not Flux.j#zygote. See https://github.com/FluxML/Flux.jl/issues/1168

```julia
using Flux, Random, Statistics

Random.seed!(1234)

import Flux.Optimise: apply!

mutable struct ClipValue{T}
    thresh::T
end

apply!(o::ClipValue, x, Δ) = clamp!(Δ, -o.thresh, o.thresh)

t = range(0, stop = 5, length = 500)
x = zeros(Float32, 0, 1, length(t))
y = cos.(2π .* reshape(t, 1, 1, :))
# x = randn(Float32, 10, 1, 100)
# y = mean(x, dims = 1)

model = Chain(GRU(0, 100), GRU(100, 10), GRU(10, 1))

function ptanh(x::Real)
    l = oftype(x / 1, 1)
    o = oftype(x / 1, 0)
    xθ = oftype(x / 1, 1.92033)
    yθ = oftype(x / 1, 0.96016)
    λ = oftype(x / 1, 0.26037)
    ifelse(x > xθ, yθ, ifelse(x > o, yθ - λ * (x - xθ)^2, ifelse(x > -xθ, λ * (x + xθ)^2 - yθ, -yθ)))
end

function pσ(x::Real)
    l = oftype(x / 1, 1)
    o = oftype(x / 1, 0)
    h = oftype(x / 1, 0.5)
    x = x / oftype(x / 1, 4.1)
    ifelse(x > l, l, ifelse(x < -l, o, h + x * (l - h * abs(x))))
end

function loss(x, y)
    xs = Flux.unstack(x, 3)
    ys = Flux.unstack(y, 3)
    ŷs = model.(xs)
    l = 0f0
    for t in 1:length(ŷs)
        l += Flux.mse(ys[t], ŷs[t])
    end
    return l / length(ŷs)
end

opt = Flux.Optimise.Optimiser(ClipValue(0.5f0), ADAM(1e-4, (0.9, 0.999)))

cb = () -> (println(Flux.data(loss(x, y))); Flux.reset!(model))

NNlib.tanh(x::T) where T <: Union{Float32, Float64} = ptanh(x)
NNlib.σ(x::T) where T <: Union{Float32, Float64} = pσ(x)
# import Base.Broadcast: broadcasted
# using Zygote: @adjoint, Numeric
# @adjoint function broadcasted(::typeof(ptanh), x::Numeric)
#     y = ptanh.(x)
#     y, ȳ -> (nothing, ȳ .* conj.(1 .- y.^2))
# end

Flux.@epochs 1000 Flux.train!(loss, params(model), repeat([(x, y)], 100), opt, cb = cb)
```

AStupidBear:
This PR will be useless once https://github.com/FluxML/NNlib.jl/pull/199 is merged

PR #196, opened 2020-04-20
BatchedAdjOrTrans interface tests

sudo-rushil:
Added tests to ensure that the BatchedAdjOrTrans lazy wrapper correctly interfaces with Base.


CarloLucibello:
thanks!

PR #199, opened 2020-05-10
use LoopVectorization to vectorize activation functions and softmax

AStupidBear:
```julia
using BenchmarkTools
using NNlib
using Zygote
```

Old NNlib

```julia
julia> for f in (tanh, σ)
           println(f)
           @btime $f.(x)
           @btime Zygote.gradient($x) do z
               sum($f.(z))
           end
       end
tanh
  88.858 μs (3 allocations: 16.16 KiB)
  114.096 μs (20 allocations: 32.61 KiB)
σ
  48.257 μs (3 allocations: 16.16 KiB)
  56.673 μs (5 allocations: 32.30 KiB)

julia> for d in 1:ndims(x)
           println(softmax, " dim=", d)
           @btime softmax($x, dims = $d)
           @btime Zygote.gradient($x) do z
               sum(softmax(z, dims = $d))
           end
       end
softmax dim=1
  58.959 μs (13 allocations: 32.86 KiB)
  133.949 μs (30 allocations: 98.19 KiB)
softmax dim=2
  51.954 μs (31 allocations: 34.23 KiB)
  110.524 μs (72 allocations: 101.53 KiB)
```

This PR:

```julia
julia> for f in (tanh, σ)
           println(f)
           @btime $f.(x)
           @btime Zygote.gradient($x) do z
               sum($f.(z))
           end
       end
tanh
  9.181 μs (1 allocation: 16.13 KiB)
  11.725 μs (20 allocations: 32.61 KiB)
σ
  2.344 μs (1 allocation: 16.13 KiB)
  4.755 μs (5 allocations: 32.30 KiB)

julia> for d in 1:ndims(x)
           println(softmax, " dim=", d)
           @btime softmax($x, dims = $d)
           @btime Zygote.gradient($x) do z
               sum(softmax(z, dims = $d))
           end
       end
softmax dim=1
  25.194 μs (13 allocations: 32.86 KiB)
  56.679 μs (30 allocations: 98.19 KiB)
softmax dim=2
  9.377 μs (31 allocations: 34.23 KiB)
  25.198 μs (72 allocations: 101.53 KiB)
```

Other activation functions can be sped up by overloading `Base.broadcasted` after the adjoint is defined in `LoopVectorization` (https://github.com/chriselrod/LoopVectorization.jl/issues/108).


CarloLucibello:
`vreduce` doesn't support a `dims` args yet

AStupidBear:
@CarloLucibello The next release of LoopVectorization will has that.

AStupidBear:
Tests are passed now. Any idea?

src/activation.jl
```diff
@@ -204,25 +209,29 @@ logcosh(x::Real) = x + softplus(-2x) - log(oftype(x, 2))
...
 
 # Provide an informative error message if activation functions are called with an array
 for f in (:σ, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
     @eval $(f)(x::AbstractArray, args...) =
       error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
 end
+
+for f in (:σ, :tanh)
+    @eval Base.broadcasted(::typeof($f), x::AbstractArray{T, N}) where {T <: Union{Float64, Float32}, N} = vmap($f, x)
+end
```
CarloLucibello, line 237: add newline

src/activation.jl
```diff
@@ -204,25 +209,29 @@ logcosh(x::Real) = x + softplus(-2x) - log(oftype(x, 2))
...
+softshrink(x::RealOrFloatType, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
 
 # Provide an informative error message if activation functions are called with an array
 for f in (:σ, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
     @eval $(f)(x::AbstractArray, args...) =
       error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
 end
+
+for f in (:σ, :tanh)
+    @eval Base.broadcasted(::typeof($f), x::AbstractArray{T, N}) where {T <: Union{Float64, Float32}, N} = vmap($f, x)
```
CarloLucibello, line 236: we can extend this to all activations, right?
Also, maybe better play and restrict the signature to `Array{T, N}`, since my understanding is that `vmap` doesn't support e.g. StaticArrays yet

CarloLucibello:
Is all of this Zygote friendly?

src/softmax.jl
```diff
@@ -26,9 +28,9 @@ julia> softmax([1, 2, 3])
 See also [`logsoftmax`](@ref).
 """
 function softmax(xs::AbstractArray; dims=1)
-    max_ = maximum(xs, dims=dims)
-    exp_ = exp.(xs .- max_)
-    exp_ ./ sum(exp_, dims=dims)
+    max_ = vreduce(max, xs, dims=dims)
+    exp_ = SLEEF.exp.(xs .- max_)
+    exp_ ./ vreduce(+, exp_, dims=dims)
```
CarloLucibello, line 33: also here, these operations won't support generic `AbstractArray`s. This is very annoying, but if we want this we'll have to create a copy of the original softmax specialized on Array`

chriselrod: Dispatch inside LoopVectorization functions and macros should automatically be hitting `SLEEFPirates.exp`. Is that not happening?

src/softmax.jl
```diff
@@ -98,19 +100,19 @@ function logsoftmax!(out::AbstractVecOrMat, xs::AbstractVecOrMat)
             end
             s = zero(eltype(out))
             for i = 1:size(out, 1)
-                s += exp(xs[i, j] - xi_max)
+                s += SLEEF.exp(xs[i, j] - xi_max)
```
chriselrod, line 103: On Linux with a recent glibc, `libmvec` provides a fast `exp` implementation. LoopVectorization with `@avx` will use this; it will often be faster than the autovectorizer working on the scalar `SLEEFpirates.exp`.
It's worth benchmarking, and of course always worth considering the compile time trade-off as well.

chriselrod: Hmm:
```julia
julia> using SLEEFPirates, LoopVectorization
[ Info: Precompiling SLEEFPirates [476501e8-09a2-5ece-8869-fb82de89a1fa]
[ Info: Precompiling LoopVectorization [bdcacae8-1622-11e9-2a5c-532679323890]

julia> function sumexp(x)
           s = zero(eltype(x))
           @inbounds @simd for i ∈ eachindex(x)
               s += SLEEFPirates.exp(x[i])
           end
           s
       end
sumexp (generic function with 1 method)

julia> function sumexpavx(x)
           s = zero(eltype(x))
           @avx for i ∈ eachindex(x)
               s += SLEEFPirates.exp(x[i])
           end
           s
       end
sumexpavx (generic function with 1 method)

julia> x64 = rand(1024);

julia> @btime sumexp($x64)
  4.613 μs (0 allocations: 0 bytes)
1737.405484403936

julia> @btime sumexpavx($x64)
  596.193 ns (0 allocations: 0 bytes)
1737.4054844039383

julia> @btime sum(exp, $x64)
  7.412 μs (0 allocations: 0 bytes)
1737.405484403936

julia> x32 = rand(Float32, 1024);

julia> @btime sumexp($x32)
  245.346 ns (0 allocations: 0 bytes)
1765.1919f0

julia> @btime sumexpavx($x32)
  295.561 ns (0 allocations: 0 bytes)
1765.1919f0

julia> @btime sum(exp, $x32)
  4.657 μs (0 allocations: 0 bytes)
1765.192f0
```
So `SLEEFPirates.exp` in a loop is faster with `Float32`. I'll look into this.

chriselrod: With `Float64`, autovectorization fails on Julia 1.4.1.
With `Float32`, avoiding ccall and instead inlining and everything, hoisting the constant loads, and unrolling is faster for longer input vectors.

chriselrod: Anyway, worth pointing out that reducing a floating point reduction isn't legal without some sort of annotation (like `@simd`).

CarloLucibello:
@AStupidBear bump on this, it would be really nice to have this performance improvement

chriselrod:
FWIW, I added a much faster AVX512-`Float32` `tanh` to SLEEFPirates.
I haven't released it yet, but if it will be used here, I can try to add AVX2 and `Float64` versions as well.

These were still slower than:
```julia
functh tanh_fast(x)
    exp2x = exp(x + x)
    (exp2x - 1)/(exp2x + 1)
end
```
But they are a little more accurate.

src/activation.jl
AStupidBear: We cannot extend this to all activations because the adjoint of `vmap` is not defined. `σ` and `tanh` are special because their broadcasting adjoints are explicitly definded in Zygote. See https://github.com/FluxML/Zygote.jl/blob/40db80ce0d9344d0a30d011cacd359630809db4d/src/lib/broadcast.jl#L99.

We should restrict the signature to `Array{T, N}` and also `SubArray`.

src/softmax.jl
AStupidBear: `SLEEF.exp.(...)` can handle generic `AbstractArrays` and generate vectorized code just as `@avx`. I will change that to be `@avx exp.(...)` or `vmap(exp, ...)` later, but `vmap` `vreduce` and `@avx` may not work on generic  `AbstractArrays`. @chriselrod Do we need to dispatch non `Array` type to Base methods in `LoopVectorization`?

AStupidBear:
> Is all of this Zygote friendly?

Since all those modifications are in kernel functions used in the forward and backward pass of Zygote, they are  Zygote friendly.

src/softmax.jl
AStupidBear: Thanks! I will change that to `@avx`.

src/softmax.jl
AStupidBear: Another way is to add seperate methods in NNlib instead of LoopVectorization to deal with different  `AbstractArrays`. What do you think? @CarloLucibello @chriselrod

chriselrod: LoopVectorization should be "compatible" with generic `AbstractArrays` in that it calls `check_args` on them:
```julia
julia> @macroexpand @avx for i ∈ eachindex(x, y)
           s += x[i] * y[i]
       end
quote
    begin
        var"##loopi#366" = LoopVectorization.maybestaticrange(eachindex(x, y))
        var"##i_loop_lower_bound#367" = LoopVectorization.maybestaticfirst(var"##loopi#366")
        var"##i_loop_upper_bound#368" = LoopVectorization.maybestaticlast(var"##loopi#366")
    end
    if LoopVectorization.check_args(x, y)
        var"##vptr##_x" = LoopVectorization.stridedpointer(x)
        var"##vptr##_y" = LoopVectorization.stridedpointer(y)
        local var"##s_0"
        begin
            $(Expr(:gc_preserve, :(var"##s_0" = begin
          begin
              var"##Tloopeltype##" = promote_type(eltype(x), eltype(y))
              var"##Wvecwidth##" = LoopVectorization.pick_vector_width_val(eltype(x), eltype(y))
          end
          LoopVectorization._avx_!(Val{(0, 0, 0, LoopVectorization.unwrap(var"##Wvecwidth##"))}(), Tuple{:LoopVectorization, :getindex, LoopVectorization.OperationStruct(0x0000000000000001, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, LoopVectorization.memload, 0x01, 0x01), :LoopVectorization, :getindex, LoopVectorization.OperationStruct(0x0000000000000001, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, LoopVectorization.memload, 0x02, 0x02), :LoopVectorization, :LOOPCONSTANTINSTRUCTION, LoopVectorization.OperationStruct(0x0000000000000000, 0x0000000000000000, 0x0000000000000001, 0x0000000000000000, LoopVectorization.constant, 0x00, 0x03), :LoopVectorization, :vfmadd_fast, LoopVectorization.OperationStruct(0x0000000000000001, 0x0000000000000001, 0x0000000000000000, 0x0000000000010203, LoopVectorization.compute, 0x00, 0x03)}, Tuple{LoopVectorization.ArrayRefStruct{:x,Symbol("##vptr##_x")}(0x0000000000000001, 0x0000000000000001, 0x0000000000000000), LoopVectorization.ArrayRefStruct{:y,Symbol("##vptr##_y")}(0x0000000000000001, 0x0000000000000001, 0x0000000000000000)}, Tuple{0, Tuple{4}, Tuple{3}, Tuple{}, Tuple{}, Tuple{}, Tuple{}}, Tuple{:i}, (var"##i_loop_lower_bound#367":var"##i_loop_upper_bound#368",), var"##vptr##_x", var"##vptr##_y", s)
      end), :x, :y))
        end
        s = LoopVectorization.reduced_add(var"##s_0", s)
    else
        $(Expr(:inbounds, true))
        local var"#20#val" = for i = eachindex(x, y)
                    #= REPL[3]:2 =#
                    s = Base.FastMath.add_fast(s, Base.FastMath.mul_fast(x[i], y[i]))
                end
        $(Expr(:inbounds, :pop))
        var"#20#val"
    end
end
```
If `check_args` returns true, it will call `_avx_!`, which will use LoopVectorization to optimize the loop.
If `check_args` returns false, it'll use `@inbounds @fastmath`.
```julia
julia> using LinearAlgebra

julia> LoopVectorization.check_args(rand(3,3))
true

julia> LoopVectorization.check_args(Symmetric(rand(3,3)))
false
```

chriselrod: This is for loops. For broadcasts, it dispatches on the return type. So it's probably better to have explicit checks, until I switch to handling it more similarly to loops.

chriselrod: `SLEEF.exp` in a broadcast is probably fine. For `Float32`, I found it was faster for long loops than using LoopVectorization, because when LLVM vectorized it, it unrolled it and interleaved all the individual instructions of the `exp`, letting out of order execution work to compute a few in parallel (on top of the SIMD).

chriselrod: Well, that's what I remembered. Not what I'm seeing at the moment:
```julia
julia> using SLEEFPirates

julia> x128 = rand(Float32, 128); y128 = similar(x128);

julia> x127 = rand(Float32, 127); y127 = similar(x127);

julia> @btime @. $y128 = SLEEFPirates.exp($x128);
  42.911 ns (0 allocations: 0 bytes)

julia> @btime @avx @. $y128 = SLEEFPirates.exp($x128);
  34.418 ns (0 allocations: 0 bytes)

julia> @btime @avx @. $y128 = exp($x128);
  34.660 ns (0 allocations: 0 bytes)

julia> @btime @. $y127 = SLEEFPirates.exp($x127);
  106.803 ns (0 allocations: 0 bytes)

julia> @btime @avx @. $y127 = exp($x127);
  34.787 ns (0 allocations: 0 bytes)

julia> x512 = rand(Float32, 512); y512 = similar(x512);

julia> @btime @. $y512 = SLEEFPirates.exp($x512);
  139.423 ns (0 allocations: 0 bytes)

julia> @btime @avx @. $y512 = exp($x512);
  125.525 ns (0 allocations: 0 bytes)

julia> x1024 = rand(Float32, 1024); y1024 = similar(x1024);

julia> @btime @. $y1024 = SLEEFPirates.exp($x1024);
  276.671 ns (0 allocations: 0 bytes)

julia> @btime @avx @. $y1024 = exp($x1024);
  255.173 ns (0 allocations: 0 bytes)
```

AStupidBear: In the latest commit, I found that `vmap` and `vreduce` don't work for `StaticArrays` while the `@avx` loops work. So I created functions like `fsum` `fmaximum` and `fmap` to get around this. If you'd like to modify those functions in `LoopVectorization`, I will remove those ugly definitions.

AStupidBear: The negligible performance degragation of migrating from `SLEEFPirates.exp.(...)` to `vmap(exp, ...)` seems fine to me.

src/softmax.jl
```diff
@@ -26,32 +33,25 @@ julia> softmax([1, 2, 3])
 See also [`logsoftmax`](@ref).
 """
 function softmax(xs::AbstractArray; dims=1)
-    max_ = maximum(xs, dims=dims)
-    exp_ = exp.(xs .- max_)
-    exp_ ./ sum(exp_, dims=dims)
+    max_ = fmaximum(xs, dims=dims)
+    exp_ = fmap(exp, xs .- max_)
```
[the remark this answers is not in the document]
chriselrod: Ah, yes, `@avx` broadcasting assumed that if `isone(stride(A,i))` is known at compile time, that it isn't being broadcast along that dimension.

CarloLucibello:
what's the status of this?

AStupidBear:
It's ready to get merged.

```julia
julia> using NNlib, StaticArrays; x = SMatrix{2, 2}(rand(2, 2))
2×2 SArray{Tuple{2,2},Float64,2,4} with indices SOneTo(2)×SOneTo(2):
 0.604445  0.330955
 0.975996  0.909042

julia> softmax(x)
2×2 SArray{Tuple{2,2},Float64,2,4} with indices SOneTo(2)×SOneTo(2):
 0.408166  0.359373
 0.591834  0.640627

julia> σ.(x)
2×2 SArray{Tuple{2,2},Float64,2,4} with indices SOneTo(2)×SOneTo(2):
 0.646673  0.581992
 0.726313  0.712804

julia> logsoftmax(x)
2×2 SArray{Tuple{2,2},Float64,2,4} with indices SOneTo(2)×SOneTo(2):
 -0.89608  -1.02339 
 -0.52453  -0.445308
```

src/softmax.jl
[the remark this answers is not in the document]
chriselrod: In about half an hour I'll release a new version of LoopVectorization that allows `LowDimArray{D,T,N}` to specify less than `N` dimensions in `D`. `LowDimArray{(false,)}(rand(1,5,6))` would then allow LoopVectorization's broadcasting to be aware that the first dimension is only `1`.

This could then be used as:
```julia
max_ = fmaximum(xs, dims=dims)
exp_ = if 1 ∈ dims
    @avx exp.(xs .- LowDimArrays{(false,)}(max_))
else dims isa Integer
    @avx exp.(xs .- max_)
end
```
(Although the issues about broadcasting being less robust about array types is still there.)

CarloLucibello:
Alright, let's merge this and tag a new release. If no problems come up, we can then extend apply vmap to all activations by providing custom adjoints, right @AStupidBear ?

AStupidBear:
@CarloLucibello Yes! But where should we put those definitions? Zygote?

CarloLucibello:
yes, here https://github.com/FluxML/Zygote.jl/blob/master/src/lib/nnlib.jl

AStupidBear:
Maybe it's better to dispatch other activation functions to `vmap`  and then define the adjoint for `vmap`? @chriselrod Any idea?

CarloLucibello:
I think we can simply copy the adjoint map, i.e. doing something similar to https://github.com/FluxML/Zygote.jl/pull/728

ChrisRackauckas:
yeah, we can just add `vmap` to the loop I have in that PR. That would need a LoopVectorization dependency then?

CarloLucibello:
yes. also needs some tests

Issue #200, opened 2020-05-21
NNPACK_CPU_THREADS and thread pool issue (ERROR: KeyError: key 3 not found)


Alexander-Barth:
I try to use the dev version of NNlib (the issue is also present with the released version v0.6.6) with NNPACK on Linux.

The CPU (AMD EPYC 7551P 32-Core) has AVX2 (required by NNPACK)

```
$ cat /proc/cpuinfo | grep avx2 | uniq
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm art rep_good nopl extd_apicid eagerfpu pni pclmulqdq ssse3 fma cx16 sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm cmp_legacy svm cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw arat npt nrip_save fsgsbase tsc_adjust bmi1 avx2 smep bmi2 rdseed adx smap clflushopt sha_ni xsaveopt xsavec xgetbv1
```

and NNPACK is available in NNlib

```
julia> NNlib.is_nnpack_available()
true
```

When I try to run this loop with julia 1.4.1:

```bash
for i in 1 2 4 8; do NNPACK_CPU_THREADS=$i julia --eval "using NNlib; @show NNlib.NNPACK_CPU_THREADS; @time NNlib.conv(zeros(Float32,200,200,100,50),ones(Float32,4,4,100,100));"; done
```

I get the following error messages:

```
NNlib.NNPACK_CPU_THREADS = 0.0
ERROR: KeyError: key 0 not found
Stacktrace:
 [1] getindex(::Dict{UInt64,Base.RefValue}, ::Int64) at ./dict.jl:477
 [2] select_threadpool(::DenseConvDims{2,(4, 4),100,100,(1, 1),(0, 0, 0, 0),(1, 1),false}, ::Int64) at /home/abarth/.julia/dev/NNlib/src/nnpack/performance.jl:4
[...]

PR #201, opened 2020-05-21
fix allocate_threadpool when NNPACK_CPU_THREADS is set

Alexander-Barth:
This should fix issue #200 .
With this patch I have the following results for the convolution:
```bash
for i in 1 2 4 8; do 
   NNPACK_CPU_THREADS=$i julia --eval "using NNlib; @show Int(NNlib.NNPACK_CPU_THREADS); @time NNlib.conv(zeros(Float32,200,200,100,50),ones(Float32,4,4,100,100));"
 done
```

```
Int(NNlib.NNPACK_CPU_THREADS) = 1
  7.966892 seconds (508.67 k allocations: 1.494 GiB, 0.44% gc time)
Int(NNlib.NNPACK_CPU_THREADS) = 2
  4.303962 seconds (508.67 k allocations: 1.494 GiB, 0.81% gc time)
Int(NNlib.NNPACK_CPU_THREADS) = 4
  3.020974 seconds (508.67 k allocations: 1.494 GiB, 1.14% gc time)
Int(NNlib.NNPACK_CPU_THREADS) = 8
  2.539355 seconds (508.67 k allocations: 1.494 GiB, 1.37% gc time)
```

### DIFF
--- a/src/nnpack/NNPACK.jl
+++ b/src/nnpack/NNPACK.jl
@@ -34,8 +34,8 @@ Allows NNPACK to intelligently choose which threadpool to use for getting the be
 performance.
 """
 function allocate_threadpool()
-    global NNPACK_CPU_THREADS = NNPACK_CPU_THREADS > 8 ? UInt64(8) : floor(log2(NNPACK_CPU_THREADS))
-    for i in 1:Int(NNPACK_CPU_THREADS)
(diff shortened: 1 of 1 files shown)


CarloLucibello:
thanks!

PR #202, opened 2020-05-30
Fix wrong maxpool gradient

yiyuezhuo:
Fix Flux issue [#1197](https://github.com/FluxML/Flux.jl/issues/1197).


yiyuezhuo:
Strangely, while new implementation gives correct gradient for some top layer (ex. Conv1 in ResNet) and old implementation gives wrong gradient. Gradient for other layers (bn1 in ResNet) seems wrong so I need to check it carefully.

PR #204, opened 2020-06-04
Update package to use jll

### LINKED ISSUE #192: Provide artifact for BB download
This is one of the few packages that does not yet have an artifact for BB download as noted in https://github.com/JuliaPackaging/Yggdrasil/issues/653

aviks:
Minimum Julia version is now  1.3

[Build script](https://github.com/JuliaPackaging/Yggdrasil/blob/master/N/NNPACK/build_tarballs.jl)

(As an aside, this package seems to check in the Manifest.toml. Is a good idea?)

Fix #192


aviks:
Failure on nightly is apparently due to IRTools not supporting julia nightly

```
WARNING: could not import Compiler.just_construct_ssa into Wrap
ERROR: LoadError: LoadError: InitError: UndefVarError: Params not defined
Stacktrace:
 [1] top-level scope at none:1
 [2] eval at .\boot.jl:331 [inlined]
 [3] define_typeinf_code2() at C:\Users\appveyor\.julia\packages\IRTools\BpoqK\src\reflection\reflection.jl:21
 [4] __init__() at C:\Users\appveyor\.julia\packages\IRTools\BpoqK\src\IRTools.jl:37
```

dhairyagandhi96:
Yes, nightly needs a new tag for IRTools

Generally looks good. We had noticed issues earlier when making nnpack the default, so will run the Flux test suite on this

ViralBShah:
I thought NNPACK was always used when I use Flux.

dhairyagandhi96:
That was when we had to transition. We'd made the Julia backend the default at the time

Project.toml
```diff
@@ -1,18 +1,18 @@
 name = "NNlib"
 uuid = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
-version = "0.6.6"
+version = "0.6.7"
```
giordano, line 3: I think a restriction of allowed Julia versions warrants a minor version bump

.travis.yml
```diff
@@ -4,11 +4,11 @@ language: julia
...
   - linux
   - osx
-  # - windows   
-  
+  # - windows
+
 julia:
-  - 1.0
   - 1.3
+  - 1.4
```
CarloLucibello, line 11:
```suggestion
  - 1
```

CarloLucibello: so that we always test on latest stable release

src/NNlib.jl
```diff
@@ -1,14 +1,19 @@
...
-include(joinpath(@__DIR__, "..", "deps", "deps.jl"))
-if check_deps() == nothing
+
+if isdefined(NNPACK_jll, :libnnpack)
   include("nnpack/NNPACK.jl")
 else
+  @warn "NNPACK not available for your platform: " *
+        "$( Pkg.BinaryPlatforms.platform_name(Pkg.BinaryPlatforms.platform_key_abi()))" *
+        "($( Pkg.BinaryPlatforms.triplet(Pkg.BinaryPlatforms.platform_key_abi())))
+        You will only be able to use only the default Julia NNlib backend"
```
CarloLucibello, line 16:
```suggestion
        You will be able to use only the default Julia NNlib backend"
```

src/NNlib.jl
```diff
@@ -1,14 +1,19 @@
 module NNlib
+using Pkg
```
CarloLucibello, line 2: do we need this?

aviks: Yes, for the error message to say what platform you are on. https://github.com/FluxML/NNlib.jl/pull/204/files#diff-04ed3dd551723d9193b5371386dea323R14

Project.toml
CarloLucibello: I think it's fine, julia users on any version won't see any breaking change. Propagating to the ecosystem a minor version update is slightly annoying and I am always happy to avoid it

dhairyagandhi96: I would suggest a version bump here too if it changes underlying behaviour of a package. In this case the pr makes use of nnpack by default, so should be reflected in versioning.

CarloLucibello:
I tried this PR in combination with Flux and Flux's test works fine. 

The NNPACK version seems quite outdated, it would be nice to have a bump for that, where can I ping for it?

Project.toml
aviks: > propagating to the ecosystem a minor version update is slightly annoying
That was my thought as well. 

>In this case the pr makes use of nnpack by default
I'm not sure I understand how that worked.  I thought the current code is functionally equivalent -- if the binaries are present, they'll be used. Just that binaries are now available for more platforms. But maybe I'm misunderstanding something.

aviks:
> The NNPACK version seems quite outdated, it would be nice to have a bump for that, where can I ping for it?

The build recipe is here: https://github.com/JuliaPackaging/Yggdrasil/blob/master/N/NNPACK/build_tarballs.jl , it needs to be updated to the latest tags. There seem to be quite a few dependencies, so might be a bit of work.

Project.toml
giordano: Quoting from https://github.com/JuliaRegistries/RegistryCI.jl/blob/729889777565fb0275110034ee933bd509b8962e/src/AutoMerge/guidelines.jl#L94-L100

> A patch release is not allowed to narrow the supported ranges of Julia versions

giordano: > Propagating to the ecosystem a minor version update is slightly annoying and I am always happy to avoid it

This is the problem of being in 0.x for too long.  Narrowing the supported ranges of Julia versions always requires a minor version bump, but in 1+ series this is not a big deal

CarloLucibello: > Quoting from https://github.com/JuliaRegistries/RegistryCI.jl/blob/729889777565fb0275110034ee933bd509b8962e/src/AutoMerge/guidelines.jl#L94-L100
> 
> > A patch release is not allowed to narrow the supported ranges of Julia versions

this applies only to `if (old_version >= v"1") || (new_version >= v"1")` though

giordano: Uhm, right :sweat_smile:

CarloLucibello:
ok, I'll merge, we can sort out the version number later

PR #207, opened 2020-06-07
Fix maxpool gradient

### LINKED ISSUE #205: Wrong MaxPool gradient
I tried to reused PyTorch weights in Flux and made the output fully consistent. But unfortunately, gradient results seem wrong. Finally I figured out that it's due to "Strided" MaxPool's wrong gradient. Following gists can be used to show it:

https://gist.github.com/yiyuezhuo/94a08637a6a80ac4a25eb89e5e605e16

`compare_master.ipynb` gist show present implementation comparison results.
image

By just replace it [=](https://github.com/FluxML/NNlib.jl/blob/master/src/impl/pooling_direct.jl#L179) with [+=](https://github.com/yiyuezhuo/NNlib.jl/blob/fix_maxpool_gradient/src/impl/pooling_direct.jl#L179). The most inconsistency are removed. 
image

But some inconsistency still hold. It's due to PyTorch's row major and Flux/Julia's column major difference. They follow their natural order to find the first element to apply gradients. So replace the iteration [order](https://github.com/yiyuezhuo/NNlib.jl/blob/fix_maxpool_gradient_further/src/impl/pooling_direct.jl#L165) can remove almost all inconsistency:

image

But this modification may damage performance, and just get vain gain.

Present [unit test](https://github.com/FluxML/NNlib.jl/blob/master/test/pooling.jl#L271) failed to detect this error, since `reshape(1:n, ...)` style example is unable to check the case in which  an entry is maximum on many windows. The `=` override other gradients passed to it though.

I have prepared two branches to Pull Request:

[fix_maxpool_gradient](https://github.com/yiyuezhuo/NNlib.jl/tree/fix_maxpool_gradient)
[fix_maxpool_gradient_further (change iteration order)](https://github.com/yiyuezhuo/NNlib.jl/tree/fix_maxpool_gradient_further)

Which one should I PR? (Well, since this [issue](https://github.com/FluxML/Flux.jl/issues/1197) is ignored entirely, maybe no PR will be accepted. But the gradient is still wrong, a solution will be fine imo.)

yiyuezhuo:
Fix #205


CarloLucibello:
Thanks for this! Does `meanpool` have similar problems?

yiyuezhuo:
No, I removed the modification in ` 924f472` commit.

CarloLucibello:
Are there some corner cases this fix doesn't cover?

yiyuezhuo:
If the caller has set `beta=1` to "fix" the problem themself, this fix will break their code. But at least Flux doesn't do this. If we consider compatibility with that, changing `beta` instead of `=` will be a better solution.

yiyuezhuo:
Tests for some kernel size, stride, pad, meanpool and assert have been added.

CarloLucibello:
thank you, tests are quite exhaustive.

Issue #210, opened 2020-06-10
maxpool NNPACK breaks nonzero padding with stride 1


dsweber2:
As the title says, using `maxpool` on the cpu with nonzero padding and a stride of 1 results in an error. On a previous version, `maxpool` with these parameters worked fine (not sure which, sometime within the past 3 weeks certainly). Strangely, the 1D version actually works. Examples and the actual error follows
```
using CUDA, NNlib
maxpool(cu(randn(25,2,1,3)),(2,1), pad=(1,0),stride=(1,1)) # works fine
maxpool(randn(25,2,1,3),(2,1), pad=(1,1),stride=(1,1)) # breaks
maxpool(randn(25,2,1,3),(2,1), pad=(1,1),stride=(2,2)) # fine
maxpool(randn(25,1,3),(2,), pad=2,stride=(1,)) # fine
```

```
ERROR: NNPACKError(code 12, NNPACK STATUS INVALID INPUT PADDING)
Stacktrace:
 [1] macro expansion at /home/dsweber/.julia/packages/NNlib/XKZ39/src/nnpack/error.jl:79 [inlined]
 [2] nnp_max_pooling_output at /home/dsweber/.julia/packages/NNlib/XKZ39/src/nnpack/libnnpack.jl:81 [inlined]
 [3] nnp_max_pooling_output(::Array{Float32,4}, ::Array{Float32,4}, ::Tuple{Int64,Int64}; padding::NTuple{4,Int64}, stride::Tuple{Int64,Int64}, threadpool::Ptr{Nothing}) at /home/dsweber/.julia/packages/NNlib/XKZ39/src/nnpack/libnnpack.jl:89
 [4] maxpool_nnpack!(::Array{Float32,4}, ::Array{Float32,4}, ::PoolDims{2,(2, 1),(1, 1),(1, 1, 1, 1),(1, 1)}) at /home/dsweber/.julia/packages/NNlib/XKZ39/src/nnpack/impl.jl:4
 [5] maxpool_nnpack!(::Array{Float64,4}, ::Array{Float64,4}, ::PoolDims{2,(2, 1),(1, 1),(1, 1, 1, 1),(1, 1)}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/dsweber/.julia/packages/NNlib/XKZ39/src/nnpack/interface.jl:44
[...]

PR #212, opened 2020-06-15
updates for NNpack

CarloLucibello:
maxpool tests are currently failing due to some type conversions. We didn't catch this since the tests were added in #207 at the same time of #204 . 

Removing the type casts I got the tests passing, but then #210 convinced me to bypass NNPACK's maxpool entirely as it is not completely reliable. 
I also, added NNPACK's conv gradient as defaults, for some reason only the forward was used before (@dhairyagandhi96 @MikeInnes @avik-pal @staticfloat do you know why?)

So, this PR does the following.
- use NNPACK only for Float32 operations, fallback to NNlib for Float64
- use NNPACK for conv gradient  
- don't use NNPACK for maxpool as a workaround for #210. Notice that NNPACK doesn't offer a gradient for maxpool, so the damage is a little mitigated.

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -106,7 +106,6 @@ end
 # allocation.  :P
 for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
     # First make auto-allocating versions of the conv()-like calls:
-    (backend == :_nnpack && !is_nnpack_available()) && continue
     for name in (:conv, :depthwiseconv)
         @eval begin
             function $(Symbol("$(name)$(backend)"))(
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -155,8 +155,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
     end
 end
 
-
-# # Use NNPACK if it is available and the operation is supported
+# Use NNPACK if it is available and the operation is supported
 if is_nnpack_available()
     function conv(x::Array{Float32, 4}, w::Array{Float32, 4},
                   cdims::DenseConvDims{2, K, C_in, C_out, (1, 1), P, (1, 1), F};
@@ -177,7 +176,6 @@ if is_nnpack_available()
     end
 end
 
-
 """
     conv(x, w; stride=1, pad=0, dilation=1, flipped=false)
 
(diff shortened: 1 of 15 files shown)


src/pooling.jl
```diff
@@ -103,7 +103,7 @@ end
 
 
 # Finally, let's generate auto-allocating versions of all our functions, for all backends:
-for backend in (Symbol(), :_direct, :_im2col)
```
CarloLucibello, line 106: the  im2col backend for pooling just doesn't exists

src/nnpack/interface.jl
```diff
@@ -1,48 +1,30 @@
...
-
-
-function maxpool_nnpack!(y::Array{T1, 4}, x::Array{T2, 4}, pdims::PoolDims;
-                         kwargs...) where {T1, T2}
-    @warn "Automatically converting input tensor to Float32. This will have performance implications" maxlog=1
-    # We want the output to be of the same type as desired
-    T1.(maxpool_nnpack!(Float32.(y), Float32.(x), pdims; kwargs...))
-end
+## NNPACK supports only Float32
+## Uncomment if we decide to cast to Float32
```
avik-pal, line 4: Not sure why we need to comment this out. If the user wants to use NNPACK with Float64 even after the warning I don't see any harm in typecasting and doing the operation. The presence of these functions shouldn't affect the exposed NNlib functions which should be using the fastest backend (which in this case would be the im2col/direct version)

CarloLucibello: right, will change

src/conv.jl
```diff
@@ -104,8 +104,9 @@ end
 # Finally, let's generate auto-allocating versions of all our functions, for all backends.
 # We `@timeit` these methods separately, as we want to know how much time is spent in
 # allocation.  :P
-for backend in (Symbol(), :_direct, :_im2col)
+for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
     # First make auto-allocating versions of the conv()-like calls:
+    (backend == :_nnpack && !is_nnpack_available()) && continue
     for name in (:conv, :depthwiseconv)
```
avik-pal, line 110: I don't think we have `depthwiseconv_nnpack`

PR #213, opened 2020-06-26
Remove redundant loops

pxl-th:
When dealing with padded regions in `im2col!`, loops in `if` branches are not needed in their current form, as they don't affect anything.
And it seems, like outer loops cover all `out_of_bounds` cases, if I understand correctly.


pxl-th:
Or if we want to _"deal with the entire plane"_ then we need to avoid looping over cleared out regions in `col_reshaped`.
Because then they might be overwritten... if i understand things correctly.

CarloLucibello:
yeah, I think who wrote the loop was trying to do something smarter, like setting an entire row to 0 once and for all, but that clearly wasn't the correct way so let's do the simple thing. Thanks for this contribution!

PR #226, opened 2020-07-25
Revert LoopVectorization

DhairyaLGandhi:
Following discussion in FluxML/Flux.jl#1272, #224, it is best to have a better pattern to introduce vectorisation in a more formal manner.

This reverts #199 for the time being.

### DIFF
--- a/src/activation.jl
+++ b/src/activation.jl
@@ -1,11 +1,6 @@
 export σ, sigmoid, hardσ, hardsigmoid, hardtanh, relu, leakyrelu, relu6, rrelu, elu, gelu, swish, selu, celu, softplus, softsign, logσ,
        logsigmoid, logcosh, mish, tanhshrink, softshrink, thresholdrelu, trelu, lisht
 
-import LoopVectorization: vifelse
-using LoopVectorization.SLEEFPirates: FloatType
-
-const RealOrFloatType = Union{Real, FloatType}
-
 ## Activation functions
 #
 # Some of activation functions have its wrapper function for GPU in CuArrays.jl.
@@ -17,9 +12,9 @@ const RealOrFloatType = Union{Real, FloatType}
 Classic [sigmoid](https://en.wikipedia.org/wiki/Sigmoid_function) activation
 function.
 """
-function σ(x::RealOrFloatType)
+function σ(x::Real)
     t = exp(-abs(x))
-    vifelse(x ≥ 0, inv(one(t) + t), t / (one(t) + t))
+    ifelse(x ≥ 0, inv(one(t) + t), t / (one(t) + t))
 end
 const sigmoid = σ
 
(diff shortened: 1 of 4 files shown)


src/activation.jl
```diff
@@ -17,9 +12,9 @@ const RealOrFloatType = Union{Real, FloatType}
 Classic [sigmoid](https://en.wikipedia.org/wiki/Sigmoid_function) activation
 function.
 """
-function σ(x::RealOrFloatType)
+function σ(x::Real)
```
CarloLucibello, line 15: What about removing these type restrictions? So that  people can overload broadcast without change the model to have looovectorization kick in. @AStupidBear

src/activation.jl
```diff
@@ -209,29 +204,25 @@ logcosh(x::RealOrFloatType) = x + softplus(-2x) - log(oftype(x, 2))
...
+softshrink(x::Real, λ = oftype(x/1, 0.5)) = min(max(zero(x), x - λ), x + λ)
 
 # Provide an informative error message if activation functions are called with an array
 for f in (:σ, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
     @eval $(f)(x::AbstractArray, args...) =
       error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
 end
-
-for f in (:σ, :tanh)
-    @eval Base.broadcasted(::typeof($f), x::Array{T, N}) where {T <: Union{Float64, Float32}, N} = vmap($f, x)
```
DhairyaLGandhi, line 236: This is also some aggressive piracy that is good to get rid of

src/activation.jl
DhairyaLGandhi: Probably although I don't know why that isn't possible without it?

DhairyaLGandhi:
Travis still seems to add LV in the test suite. Any idea why?

src/activation.jl
AStupidBear: Yes, it's better to remove the `Real` restriction so those functions can be used later in other packages.

DhairyaLGandhi: Currently all these functions are defined for Reals so I think it makes sense to keep that type annotation, but always happy to get rid of the type annotations that can make other types work more generically.

DhairyaLGandhi:
Any changes apart from

DhairyaLGandhi:
Any major changes apart from conflicts needed here @CarloLucibello ?

CarloLucibello:
ok, let's do this. Then we need some proper discussion on how to reorganize the code in JuliaML

PR #228, opened 2020-07-26
Added wrapper for convolution, bias, and activation

gartangh:
Required for:
https://github.com/JuliaGPU/CUDA.jl/pull/321

### DIFF
--- a/src/conv_bias_act.jl
+++ b/src/conv_bias_act.jl
@@ -1,0 +1,34 @@
+export conv_bias_act, conv_bias_act!
+
+function conv_bias_act(x::AbstractArray{xT,N}, w::AbstractArray{wT,N}, cdims::ConvDims, b::AbstractArray{bT,N}, σ=identity; kwargs...) where {xT, wT, bT, N}
+    y = similar(x, promote_type(xT, wT, bT), output_size(cdims)..., channels_out(cdims), size(x,N))
+    conv_bias_act!(y, x, w, cdims, b, σ; kwargs...)
+    return y
+end
+
+conv_bias_act!(y::AbstractArray{T,5}, x::AbstractArray{T,5}, w::AbstractArray{T,5}, cdims::ConvDims, b::AbstractArray{T,5}, σ=identity; kwargs...) where {T} =
+    σ.(conv!(y, x, w, cdims) .+ b)
+
+for N in (3, 4)
+    @eval begin
+        function $(Symbol("conv_bias_act!"))(
+                        y::AbstractArray{yT,$N}, x::AbstractArray{xT,$N},
+                        w::AbstractArray{wT,$N}, cdims::ConvDims,
+                        b::AbstractArray{bT,$N}, σ;
+                        kwargs...) where {yT, xT, wT, bT}
+            $(Symbol("conv_bias_act!"))(
+                insert_singleton_spatial_dimension(y, $(5 - N)),
+                insert_singleton_spatial_dimension(x, $(5 - N)),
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
How would we handle cases where bias is meant to be switched off? In that case, it would be nice for the current CUDNN dispatch to be the fallback to avoid allocating the intermediary that is supposed to be dropped. We can experiment with if those allocations cause issues for training larger models

gartangh:
We could add:
```Julia
  if (b == nothing || typeof(b) <: Flux.Zeros) && typeof(σ) == identity
    # bias is turned off and activation is identity
    conv(x, c.weight, cdims)
  else
    conv_bias_act(x, c.weight, cdims, b, σ)
  end
```
to Flux.jl, but I would rather catch the case of both: bias being turned off and activation not being used, in NNlib.jl:
```Julia
    if (b == nothing || typeof(b) <: Flux.Zeros) && typeof(σ) == identity
        # bias is turned off and activation is identity
        conv!(y, x, w, cdims; kwargs...)
    else
        conv_bias_act!(y, x, w, cdims, b, σ; kwargs...)
    end
```
So that case would no longer be an issue in CUDA.jl.

However, the other case remains: bias is turned off, but activation is not.

If the activation is ReLU, it would still be beneficial to merge the kernels with `cudnnConvolutionBiasActivationForward` in terms of speed, but not in terms of memory, because of the extra allocation for the bias.

Else, we could add extra checks to see if the bias is turned off:
```Julia
  if σ == NNlib.relu # always merge convolutions, bias, and relu, even when bias is turned off
    cudnnConvolutionBiasActivationForward(...)
  elseif algo == 1 && !(b == nothing  || typeof(b) <: Flux.Zeros) # only merge convolution and bias if the fastest algorithm is also the only supported algorithm and the bias is not turned off
    cudnnConvolutionBiasActivationForward(...)
    σ.(y)
  else # fallback
    if (b == nothing  || typeof(b) <: Flux.Zeros) # bias is turned off
      σ.(conv!(y, x, w, cdims, alpha=alpha1, algo=algo))
    else # bias is turned on
      σ.(add_bias(conv!(y, x, w, cdims, alpha=alpha1, algo=algo), b))
    end
  end
```

DhairyaLGandhi:
Ideally we should keep as much of the logic for this out of Flux as possible. What kind of performance impact do we see? Are there cases where allocating the bias would give us better performance?

DhairyaLGandhi:
bump

DhairyaLGandhi:
Is there a corresponding PR to Flux to use this API instead?

gartangh:
> Is there a corresponding PR to Flux to use this API instead?

@DhairyaLGandhi , that would be this PR: https://github.com/FluxML/Flux.jl/pull/1302.

src/conv_bias_act.jl
```diff
@@ -0,0 +1,34 @@
+export conv_bias_act, conv_bias_act!
+
+function conv_bias_act(x::AbstractArray{xT,N}, w::AbstractArray{wT,N}, cdims::ConvDims, b::AbstractArray{bT,N}, σ=identity; kwargs...) where {xT, wT, bT, N}
```
CarloLucibello, line 3: the bias here doesn't have dimension N

CarloLucibello: never mind the previous comment: in Flux's Conv layers, the bias (a vector) is reshaped to the same dimension of x and w when applied

CarloLucibello:
thanks for this effort!

Issue #197, opened 2020-04-28
Incorrect gradient of convolution w.r.t. weights


dfdx:
```julia
import Random
import NNlib
import NNlib: DenseConvDims


Random.seed!(42);


function ngradient(f, xs::AbstractArray...)
  grads = zero.(xs)
  for (x, Δ) in zip(xs, grads), i in 1:length(x)
    δ = sqrt(eps())
    tmp = x[i]
    x[i] = tmp - δ/2
    y1 = f(xs...)
    x[i] = tmp + δ/2
    y2 = f(xs...)
    x[i] = tmp
    Δ[i] = (y2-y1)/δ
  end
  return grads
end


function conv_loss(x, w)
    cdims = DenseConvDims(x, w; stride=1, padding=0, dilation=1)
    y = NNlib.conv(x, w, cdims)
    return sum(y)
end


x = rand(7, 7, 3, 10); w = rand(3, 3, 3, 1)
cdims = DenseConvDims(x, w; stride=1, padding=0, dilation=1)
y = NNlib.conv(x, w, cdims)
dy = ones(size(y))
    
ndx, ndw = ngradient(conv_loss, x, w)

dx = NNlib.∇conv_data(dy, w, cdims)
dw = NNlib.∇conv_filter(x, dy, cdims)

isapprox(dx, ndx, rtol=1e-5, atol=1e-5)  # true
isapprox(dw, ndw, rtol=1e-5, atol=1e-5)  # false
```
I recently updated NNlib from (I think) version 0.6.0 to the latest version 0.6.6 and my tests started to fail. `NNlib.∇conv_filter()` differs from both - numeric approximation (`ngradient`) from CUDNN implementation, and the difference is quite huge (like `[123, 128, ...]` vs `[112, 115, ...]`), so it's not about numeric instability. 

I tried to track back NNlib implementation, but huge portions of code were changed, including tests. Does anybody know what may have caused this issue?

CarloLucibello:
@dfdx for me the example script works fine on NNlib v0.6.6. Could you check if the problem still persist for you?

dfdx:
@CarloLucibello In your case, does the code work and show `true` in both 2 lines? 

I've just re-tested it and the problem still presents:

```julia
(@v1.4) pkg> st NNlib
Status `~/.julia/environments/v1.4/Project.toml`
  [872c559c] NNlib v0.6.6

julia> import Random

julia> import NNlib

julia> import NNlib: DenseConvDims

[...]

PR #235, opened 2020-09-16
Fix multithreading bug in ∇conv_filter_im2col

pazner:
Fixes issue #197.

The issue is caused by threading, and if `JULIA_NUM_THREADS` is greater than 1, it can be reproduced with the following MWE, which compares the gradients computed with `∇conv_filter_im2col` to those computed with `∇conv_filter_direct`.
```julia
using NNlib, Random

function compare_grad()
    Random.seed!(0)

    x = rand(50,1,10)
    w = rand(3,1,1)
    dy = rand(50,1,10)

    cdims = DenseConvDims(size(x), size(w); stride=1, padding=1)
    dw1 = NNlib.∇conv_filter_im2col(x, dy, cdims)
    dw2 = NNlib.∇conv_filter_direct(x, dy, cdims)
    maximum(abs, dw1-dw2)
end
```
In a correct implementation, `compare_grad()` should always return zero. Since the random seed is set to zero, the result should also be deterministic. However, when using multithreading it returns seemingly random large numbers:
```julia
julia> Base.Threads.nthreads()
4

julia> compare_grad()
28.640897825466183

julia> compare_grad()
15.199002081309956

julia> compare_grad()
26.932418843285404
```

This is caused because `∇conv_filter_im2col!` uses a threaded for-loop that attempts to accumulate the result to a single array `dw`, causing write conflicts.

I fixed this issue simply by disabling threading for this loop. Another option would be to introduce another thread-specific workspace for `dw`, and accumulate over threads, but I suspect this will not be worth it.

This PR also fixes some bugs involving `GC.@preserve` by removing commas from the argument list. Commas in this context cause the arguments to be treated as a tuple, as can be seen from the following example:
```julia
julia> @macroexpand @GC.preserve x y begin end
:($(Expr(:gc_preserve, quote
    #= REPL[11]:1 =#
end, :x, :y)))

julia> @macroexpand @GC.preserve x, y, begin end
:($(Expr(:gc_preserve, :((x, y, begin
          #= REPL[10]:1 =#
      end)))))
```

### DIFF
--- a/src/impl/conv_im2col.jl
+++ b/src/impl/conv_im2col.jl
@@ -95,9 +95,8 @@ function ∇conv_filter_im2col!(
     N = channels_out(cdims)
     K = prod(output_size(cdims))
     
-    @threads for batch_idx in 1:size(x,5)
-        # col_slice is a thread-local workspace
-        col_slice = view(col, :, :, threadid())
+    for batch_idx in 1:size(x,5)
+        col_slice = view(col, :, :, 1)
 
         im2col!(col_slice, view(x, :, :, :, :, batch_idx), cdims)
         GC.@preserve col_slice, dw, dy, begin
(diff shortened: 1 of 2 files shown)


CarloLucibello:
yeah, better disable multi-threading and fix the bug. We can try to reimplement it later in a correct way and with proper benchmarking

Issue #229, opened 2020-08-03
pool padding is buggy


denizyuret:
```
julia> x = randn(Float32,4,4,1,1)
x = randn(Float32,4,4,1,1)
4×4×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -2.03431    0.369436   0.404957  0.676001
  0.392665  -0.142985   1.75043   0.34975
  1.68848    0.559873  -0.883515  0.570732
 -2.00521    0.804538   0.685247  0.803716

julia> meanpool(x, PoolDims(x,2,padding=1))
meanpool(x, PoolDims(x,2,padding=1))
3×3×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -0.508577  0.193598  0.169
  0.520285  0.320951  0.230121
  0.0       0.372446  0.200929

julia> meanpool(cu(x), PoolDims(x,2,padding=1))
meanpool(cu(x), PoolDims(x,2,padding=1))
3×3×1×1 CuArray{Float32,4}:
[:, :, 1, 1] =
 -0.508577  0.193598  0.169
  0.520285  0.320951  0.230121
 -0.501303  0.372446  0.200929
```

jeremiedb:
@denizyuret Any development on this side? I'm a little worried to learn that such ubiquitous operators are buggy!

denizyuret:
I have not heard anything from NNlib developers.

jeremiedb:
@staticfloat Maybe you have a take on this one?

CarloLucibello:
this has been fixed in #238

Issue #218, opened 2020-07-04
Missing pooling options, inconsistencies with CuArrays


denizyuret:
There are two mean pooling operators in cudnn, one that includes and one that excludes the padded values. The current meanpool operation only supports including the padded values, we should add the other:

https://docs.nvidia.com/deeplearning/sdk/cudnn-api/index.html#cudnnPoolingMode_t

denizyuret:
Related also: `CUDNN_POOLING_MAX` excludes padding by default, this creates an inconsistency with CuArrays:
```
julia> x = randn(4,4,1,2)

julia> cx = CuArray(x)

julia> maxpool(x, PoolDims(x,2,padding=1))
3×3×1×2 Array{Float64,4}:
[:, :, 1, 1] =
 0.0       0.773338  0.0
 0.0       2.17463   0.751506
 0.299839  0.8081    0.489329

[:, :, 1, 2] =
 0.0      1.70753   0.103033
 1.26366  0.58986   0.905675
 0.0      0.927308  0.0

julia> maxpool(cx, PoolDims(x,2,padding=1))
3×3×1×2 CuArray{Float64,4}:
[:, :, 1, 1] =
 -0.976195  0.773338  -0.409254
 -0.231234  2.17463    0.751506
  0.299839  0.8081     0.489329

[:, :, 1, 2] =
 -0.119529  1.70753    0.103033
  1.26366   0.58986    0.905675
 -0.2345    0.927308  -0.978941
```

CarloLucibello:
the inconsistency has been fixed in #238, leaving this open for the missing pooling optinos

PR #238, opened 2020-11-01
fix meanpool bug #229

jeremiedb:
This fixes the errors returned by `meanpool` on edges. 
Core issue was that a `max(m, 0)` was applied to both mean and max pooling for padded elements. Instead, no operation should be performed for meanpool. 

For example, currently: 

```julia
julia> x = ones(Float32,4,4,1,1) .* -1
[:, :, 1, 1] =
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0

julia> pool = meanpool(x, PoolDims(x,2,padding=1))
3×3×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -0.25  -0.25  0.0
 -0.5   -1.0   0.0
  0.0    0.0   0.0

julia> pool = meanpool(CuArray(x), PoolDims(x,2,padding=1))
3×3×1×1 CuArray{Float32,4}:
[:, :, 1, 1] =
 -0.25  -0.5  -0.25
 -0.5   -1.0  -0.5
 -0.25  -0.5  -0.25
```

With this fix, the base cpu version now results in the same value. A test has been added. 

There's the issue raised by https://github.com/FluxML/NNlib.jl/issues/218 regarding the consistency between for maxpool between the CPU and GPU/CUDNN implementation that has not been addressed by this PR but that could be fixed in the meantime since it concerns the same code elements. It would simply involves to ignore the padded elements as well for the maxpool. 

Note: there's a couple of formatting that has been modified by Julia linter in vscode in the 2 modified files. I can revert those changes if needed, but I assume that these changes brought by the linter bring the code to expected format.

### DIFF
--- a/test/pooling.jl
+++ b/test/pooling.jl
@@ -323,6 +323,7 @@ end
 # using FiniteDifferences
 maxpool_answer_nature = Dict(
     "rank1" => Dict(
+        # kernel size 2, stride 1, pad 0
         "k2s1p0" => (size = (2,),
             stride = 1,
             pad = 0,
@@ -389,6 +390,7 @@ maxpool_answer_nature = Dict(
             ], 5, 1, 1),)
     ),
     "rank2" => Dict(
+        # kernel size 2, stride 1, pad 0
         "k2s1p0" => (size = (2, 2),
             stride = 1,
             pad = 0,
@@ -499,6 +501,7 @@ maxpool_answer_nature = Dict(
             ], 5, 5, 1, 1))
     ),
     "rank3" => Dict(
+        # kernel size 2, stride 1, pad 0
         "k2s1p0" => (size = (2, 2, 2),
             stride = 1,
             pad = 0,
--- a/test/pooling.jl
+++ b/test/pooling.jl
@@ -1,4 +1,4 @@
-#using NNlib, Test
+# using NNlib, Test
 
 maxpool_answer_dict = Dict(
     1 => Dict(
@@ -250,15 +250,12 @@ meanpool_answer_dict = Dict(
 
 for rank in (1, 2, 3)
     @testset "pool$(rank)d" begin
-        for (pool, ∇pool, answer_dict) in (
-                # Main API name
-                (maxpool, ∇maxpool, maxpool_answer_dict),
+        for (pool, ∇pool, answer_dict) in ((maxpool, ∇maxpool, maxpool_answer_dict),
                 (meanpool, ∇meanpool, meanpool_answer_dict),
 
                 # _direct name
                 (NNlib.maxpool_direct, NNlib.∇maxpool_direct, maxpool_answer_dict),
-                (NNlib.meanpool_direct, NNlib.∇meanpool_direct, meanpool_answer_dict),
-            )
+                (NNlib.meanpool_direct, NNlib.∇meanpool_direct, meanpool_answer_dict),)
 
             @testset "$(pool)$(rank)d" begin
                 y = answer_dict[rank]["y"]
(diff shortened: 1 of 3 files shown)


test/pooling.jl
```diff
@@ -324,8 +323,7 @@ end
 # using FiniteDifferences
 maxpool_answer_nature = Dict(
     "rank1" => Dict(
-        "k2s1p0" => ( # kernel size 2, stride 1, pad 0
```
CarloLucibello, line 327: we don't want to loose this comment

test/pooling.jl
```diff
@@ -396,12 +386,10 @@ maxpool_answer_nature = Dict(
                  0.333333333,
                  0.666666666,
                  0.333333333,
-            ], 5, 1, 1),
-        )
+            ], 5, 1, 1),)
     ),
     "rank2" => Dict(
-        "k2s1p0" => ( # kernel size 2, stride 1, pad 0
```
CarloLucibello, line 403: same here

CarloLucibello:
Nice hunting.
Those white spacing changes are mostly for the good but a little invasive, better avoid them in future PRs. As soon as the deleted comments are reinstated this should be ready to go

CarloLucibello:
a fix for #218 here or in another PR would be very welcome

jeremiedb:
Thanks for the quick feedback! 
I Reintegrated the lost comments in the test section. 

Also, fixes the inconsistency reported in #218. Needed to use `nextfloat` as initialization for maxpool as CuArray return minimal float value while the current NNlib was returning -Inf. Anyway, these only concern concern cases with for example a kernel 2 with padding 2. 

Below is the new behavior, consistent between cpu and cuda: 

```
julia> x = ones(Float32, 4,4,1,1) .* -1
4×4×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0

julia> p1 = maxpool(x, PoolDims(x,2,padding=1))
3×3×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0

julia> p2 = maxpool(CuArray(x), PoolDims(x,2,padding=1))
3×3×1×1 CuArray{Float32,4}:
[:, :, 1, 1] =
 -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0
```

```
julia> x = ones(Float32, 4,4,1,1) .* -1
4×4×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0
 -1.0  -1.0  -1.0  -1.0

julia> p1 = maxpool(x, PoolDims(x,2,padding=2))
4×4×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 -3.40282f38  -3.40282f38  -3.40282f38  -3.40282f38
 -3.40282f38  -1.0         -1.0         -3.40282f38
 -3.40282f38  -1.0         -1.0         -3.40282f38
 -3.40282f38  -3.40282f38  -3.40282f38  -3.40282f38

julia> p2 = maxpool(CuArray(x), PoolDims(x,2,padding=2))
4×4×1×1 CuArray{Float32,4}:
[:, :, 1, 1] =
 -3.40282f38  -3.40282f38  -3.40282f38  -3.40282f38
 -3.40282f38  -1.0         -1.0         -3.40282f38
 -3.40282f38  -1.0         -1.0         -3.40282f38
 -3.40282f38  -3.40282f38  -3.40282f38  -3.40282f38
```

I think it will be valuable to add support for the option of including/excluding padded value in the pooling operations as @denizyuret mentioned. But for now, I think having consistent behavior cpu/gpu is a first step!

On a side note, is it normal the conv tests are extremely long to run? Something to be addressed?

CarloLucibello:
> On a side note, is it normal the conv tests are extremely long to run? Something to be addressed?

yes, the fuzzing testset takes an insane amount of time, in fact it is excluded on travis. I'm not sure if it always has been like that,  maybe they are just too many. They make local testing painful, maybe we should just comment them out

PR #241, opened 2020-11-14
Improvements to `batched_mul`

mcabbott:
This adds:
* multi-threading via `Threads.@spawn`, which is disabled for very small problems
* methods `batched_mul(::Array{T,3}, ::Matrix)` which reshape to a 3-array
* unicode alias `⊠` for infix use
* a function `batched_vec(::Array{T,3}, ::Matrix)` which reshapes so that the matrix is a batch of vectors
* more detailed docstrings.

I've marked it draft as it refers to things being added here: https://github.com/JuliaMath/TensorCore.jl/pull/1. But that could be removed for now.

### DIFF
--- a/src/batched/batchedmul.jl
+++ b/src/batched/batchedmul.jl
@@ -1,26 +1,48 @@
 
-export batched_mul, batched_transpose, batched_adjoint
+export batched_mul, ⊠
+export batched_transpose, batched_adjoint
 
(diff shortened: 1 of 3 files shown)


CarloLucibello:
Is this ready?

mcabbott:
Sure, let's call it ready.

PR #242, opened 2020-11-16
port rule definitions to ChainRulesCore

simeonschaub:
The current solution of defining these in Zygote is quite unsatisfying and makes working on these unnecessarily complicated and it's also a bit weird that Zygote depends on NNlib at all. Since rules should now be defined using ChainRulesCore anyways, making use of that and moving the definitions into NNlib itself makes the most sense to me. This will require careful coordination with a new release of Zygote removing these rules, because tests will now fail because of circular dependencies. The best solution might be to add upper bounds on NNlib for all Zygote versions in General retroactively, release a new Zygote version with no dependency on NNlib and then release this. Alternatively, we could tag a breaking release for NNlib, but that seems a bit wrong to me, since these changes shouldn't actually be breaking for anything except Zygote.jl.

Rule definitions are ported from https://github.com/FluxML/Zygote.jl/blob/master/src/lib/nnlib.jl. The tests are mostly just copied from https://github.com/FluxML/Zygote.jl/blob/master/test/gradcheck.jl, eventually we probably want to port these to ChainRulesTestUtils.jl, which should be quite a bit more accurate and thorough.


DhairyaLGandhi:
It makes sense to keep NNlib fairly orthogonal and not have it depend on specific AD tooling. That's just the first thought in my head for this.

Maybe we could have the backwards functions defined here and have zygote have the "glue" adjoint/ rrule defs?

The release work should be fine either way.

simeonschaub:
> It makes sense to keep NNlib fairly orthogonal and not have it depend on specific AD tooling. That's just the first thought in my head for this.
>
> Maybe we could have the backwards functions defined here and have zygote have the "glue" adjoint/ rrule defs?

The idea behind ChainRulesCore.jl is basically that it's orthogonal to any specific AD implementation, it just allows packages to define rules in a generic way that a specific AD tool can use, so I would argue that `ChainRulesCore.rrule` is already that "glue" code. It's a fairly minimal package, so it shouldn't really have any side effects, other than that Zygote doesn't have to depend on NNlib anymore.

CarloLucibello:
> The best solution might be to add upper bounds on NNlib for all Zygote versions in General retroactively, release a new Zygote version with no dependency on NNlib and then release this. Alternatively, we could tag a breaking release for NNlib, but that seems a bit wrong to me, since these changes shouldn't actually be breaking for anything except Zygote.jl.

I would go with the easier solution, and just tag breaking releases, I'm sure no one will come after us :)
That said, I'm completely in favor of this change, Zygote should e independent from NNlib and NNlib diff rules should stay in NNlib

simeonschaub:
Ok, as @mcabbott pointed out in https://github.com/FluxML/Zygote.jl/pull/824#discussion_r524954506, I think it would make sense to use `ZygoteRules` instead of `ChainRulesCore` just for the `broadcasted` rules, because they are pretty specific to Zygote, not necessarily all AD tools. Other than that, would someone be willing to give this a review?

src/chainrulescore.jl
```diff
@@ -0,0 +1,129 @@
...
+    ∇pool = Symbol(:∇, pool)
+    pullback = Symbol(pool, :_pullback)
+    @eval function ChainRulesCore.rrule(::typeof($pool), x, pdims::PoolDims; kw...)
+        Ω = maxpool(x, pdims; kw...)
+        $pullback(Δ) = (NO_FIELDS, @thunk($∇pool(Δ, Ω, x, pdims; kw...)), DoesNotExist())
+        return Ω, $pullback
+    end
+end
+
+function ChainRulesCore.rrule(::typeof(batched_mul), A, B)
```
mcabbott, line 120: I think this should probably be restricted to `::AbstractArray{T,3}` for both. There are now also methods which accept `A::Matrix`, which is reshaped & sent to the 3-array method, and this rule applied directly to them won't work as `batched_adjoint` is then an error. 

Probably I should write some tests for their gradients since I'm to blame for the methods.

src/chainrulescore.jl
```diff
@@ -0,0 +1,129 @@
+using ChainRulesCore
+using ArrayLayouts: MemoryLayout, AbstractColumnMajor
```
mcabbott, line 2: Do we want to depend on ArrayLayouts just for this?

I believe something like this ought to work:
```
colmajor(x) = (is_strided(x) && stride(x,1) == 1) ? x : collect(x) 
```

src/chainrulescore.jl
simeonschaub: Ok, I can add those type constraints

src/chainrulescore.jl
simeonschaub: `ArrayLayouts` seems pretty minimal and should be more extensible by other packages, so I think it's fine, but I can also use your version, if people are worried about adding dependencies to NNlib.

mcabbott: I suspect the entire universe is going to end up depending on both ArrayLayouts.jl and the competing ArrayInterface.jl, directly or indirectly. There was a nice Base PR that would make `strides(A,1) === nothing` if no strides, but it gathers dust on the shelf.

simeonschaub: Ok, I just left this as-is for now, we can always revisit that in the future.

simeonschaub:
Ok, this should now be ready from my side

src/chainrulescore.jl
```diff
@@ -0,0 +1,109 @@
...
+    (:elu, :(delu.(x))),
+    (:σ, :(conj.(Ω .* (1 .- Ω)))),
+]
+    pullback = Symbol(:broadcasted_, f, :_pullback)
+    @eval @adjoint function Base.Broadcast.broadcasted(::typeof($f), x::Numeric)
+        Ω = $f.(x)
+        $pullback(Δ) = (nothing, Δ .* $df)
+        return Ω, $pullback
+    end
+end
```
CarloLucibello, line 31: let's move this to `zigoterules.jl` file

src/chainrulescore.jl
CarloLucibello: there have been some concerns in the past about too many dependencies here causing slowing down in downstream packages (e.g. CUDA.jl), so let's avoid adding another  dependency  if we could just use a oneliner

simeonschaub: Ok, I removed the dependency.

CarloLucibello:
how do we solve Compat issues? should we bump NNlib version later?

simeonschaub:
My thinking was that we first tag Zygote 0.6 and then we merge and release this. Packages like Flux that depend on both Zygote and NNlib should then bump their compat for both simultaneously..

CarloLucibello:
Zygote 0.6 is being tagged

DhairyaLGandhi:
Definitely shouldn't tag without the due diligence, I think 

If ci is a consideration, use a manifest file, not tag

CarloLucibello:
Weird, tests in the multi-threaded environment are not passing

simeonschaub:
Yes, I just checked, this reproduces for me locally as well when using two threads.

simeonschaub:
It looks like there is a race condition somewhere in the pooling code. Who would be best to take a look at that? I am not really familiar with what it is doing.

simeonschaub:
Ok, doesn't really seem related to threading, I think I misdiagnosed that. I don't think the differences here can be explained by numerical instabilities though, as they are often quite large, but am a bit unsure how to further debug this,

test/zygote.jl
```diff
@@ -0,0 +1,107 @@
...
+
+function gradcheck(f, xs...; rtol = 1e-5, atol = 1e-5)
+  grad_zygote = gradient(f, xs...)
+  grad_finite_difference = grad(central_fdm(5, 1), f, xs...)
+  #return all(isapprox.(grad_zygote, grad_finite_difference; rtol = rtol, atol = atol))
+  for (grad_zygote, grad_finite_difference) in zip(grad_zygote, grad_finite_difference)
+    @test isapprox(grad_zygote, grad_finite_difference; rtol = rtol, atol = atol)
+    if !isapprox(grad_zygote, grad_finite_difference; rtol = rtol, atol = atol)
+      display(grad_zygote - grad_finite_difference)
+      @show maximum(grad_zygote - grad_finite_difference)
```
CarloLucibello, line 15:
```suggestion
      @show maximum(abs, grad_zygote - grad_finite_difference)
```

simeonschaub:
Ok, there's something really sketchy going on with `maxpool`. I am just wondering why this wasn't caught by Zygote before. I would be really glad if someone who initially implemented this could chime in here. I can't imagine that the gradient would be that unstable, or is there some discontinuity at play here and I shouldn't be using a central difference method? I don't really see why this would be affected by threading though.

CarloLucibello:
Maybe @staticfloat or @thebhatman can help?

CarloLucibello:
I added the rules for the new softmax interface here #250 . I put the rules close to the method definition instead of using a common `chainrules.jl` file. Are there any general guidelines or common habits about where to place the rules?

CarloLucibello:
@simeonschaub since this `maxpool` is not strictly related to the porting, we could address it later

simeonschaub:
> Are there any general guidelines or common habits about where to place the rules?

Not really AFAIK. Most packages probably added them as an afterthought, so it's usually in a different file, but I see no reason not to put them in the same file if that makes more sense.

> @simeonschaub since this maxpool issue is not strictly related to the porting, we could address it later

Ok, I can try to mark these as broken for now, but it would be good to open a separate issue, since they look serious.

CarloLucibello:
> Ok, I can try to mark these as broken for now, but it would be good to open a separate issue, since they look serious.

not even worth marking them as broken, since they pass on some platforms and fail on others

CarloLucibello (approved):

CarloLucibello:
@simeonschaub merge whenever you are ready

simeonschaub:
I don't think I have commit rights here. If you merge, it would probably be good to squash merge, since I generally tend to add commits just for better reviewability with the assumption that the PR is going to be squashed anyways. I can also squash myself though, if that makes things easier.

CarloLucibello:
yes, please squash then I'll do the merge

CarloLucibello:
actually, I didn't now how easy it easy to squash-merge, I'll do that

PR #244, opened 2020-12-09
add logsumexp

CarloLucibello:
This function is available in other deep learning frameworks and would have been helpful in https://github.com/FluxML/Flux.jl/pull/1287

Also moving softmax tests to their own file


CarloLucibello:
looks like https://github.com/JuliaStats/StatsFuns.jl has `logsumexp`, and `softmax` as well. None of them gpu friendly though.

PR #246, opened 2020-12-10
github CI

CarloLucibello:
Move to gihub CI. 
Also
- disable fuzzing tests by default, since they were not run by CI and are also very slow to run locally.
- rationalize the number of tested platforms


CarloLucibello:
seems like it didn't pick up the multithreaded environment

CarloLucibello:
should be ready now

PR #247, opened 2020-12-16
added softmax!. reduced softmax's memory usage.

norci:
1. What about add a inplace version of softmax, with `dims` parameter?
2. `softmax` is updated too, this can reduce memory usage by 40%.

note:
The memory uses of `softmax!` is the same as the old `softmax`.
The performance does not change.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -27,9 +27,13 @@ julia> softmax([1, 2, 3])
 See also [`logsoftmax`](@ref).
 """
 function softmax(xs::AbstractArray; dims=1)
-    max_ = maximum(xs, dims=dims)
-    exp_ = exp.(xs .- max_)
-    exp_ ./ sum(exp_, dims=dims)
+    xs = exp.(xs .- maximum(xs, dims=dims))
+    xs ./= sum(xs, dims=dims)
+end
+
+function softmax!(xs::AbstractArray; dims=1)
+    map!(exp, xs, xs .- maximum(xs, dims=dims))
+    xs ./= sum(xs, dims=dims)
 end
 
 function softmax!(out::AbstractVecOrMat{T}, xs::AbstractVecOrMat{T}) where {T}
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -43,7 +43,7 @@ end
 ∇softmax!(Δ, x; dims = 1) = ∇softmax!(Δ, Δ, x; dims)
 function ∇softmax!(out::T, Δ::T, x::T; dims = 1) where {T<:AbstractArray}
     softmax!(out, x; dims)
-    out .= out .* (Δ .- sum(Δ .* out; dims))
+    out .*= Δ .- sum(Δ .* out; dims)
 end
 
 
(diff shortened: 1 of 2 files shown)


CarloLucibello:
I think both changes are ok. 
Since you are at it, would you mind taking care also of #249?
To make it easier, we could probably replace all those explicit loops with functional/vectorized code, 
although we have to make sure it won't impact performances.

norci:
I'm glad to help.

I have updated all the function, and refactored the test cases by removing duplicate code.

The performance has increased a little, the usage of memory has reduced a lot.

# benchmark code
```julia
using BenchmarkTools, NNlib
x = rand(1000,1000,16);
Δ = rand(1000,1000,16);
macro mb(expr)
    :(median(@benchmark $expr samples=9 evals=1))
end
@mb softmax(x)
@mb logsoftmax(x)
@mb ∇softmax(Δ, x)
@mb ∇logsoftmax(Δ, x)
```

# benchmark results
## `julia> @mb softmax(x)`
### this
```
  time:             166.028 ms
  gctime:           0.000 ns (0.00%)
  memory:           122.31 MiB
  allocs:           9
```
### original
```
  time:             174.805 ms
  gctime:           3.930 ms (2.25%)
  memory:           244.39 MiB
  allocs:           11
```
## `julia> @mb logsoftmax(x)`
### this
```
  time:             167.900 ms
  gctime:           4.940 ms (2.94%)
  memory:           244.51 MiB
  allocs:           13
```
### original
```
  time:             184.529 ms
  gctime:           3.947 ms (2.14%)
  memory:           244.51 MiB
  allocs:           13
```
## `julia> @mb ∇softmax(Δ, x)`
### this
```
  time:             213.308 ms
  gctime:           7.133 ms (3.34%)
  memory:           244.51 MiB
  allocs:           13
```
### original
```
  time:             228.071 ms
  gctime:           14.638 ms (6.42%)
  memory:           488.65 MiB
  allocs:           17
```
## `julia> @mb ∇logsoftmax(Δ, x)`
### this
```
  time:             185.306 ms
  gctime:           0.000 ns (0.00%)
  memory:           122.44 MiB
  allocs:           11
```
### original
```
  time:             217.055 ms
  gctime:           11.068 ms (5.10%)
  memory:           366.58 MiB
  allocs:           15
```

src/softmax.jl
```diff
@@ -26,49 +32,20 @@ julia> softmax([1, 2, 3])
 
 See also [`logsoftmax`](@ref).
 """
-function softmax(xs::AbstractArray; dims=1)
-    max_ = maximum(xs, dims=dims)
-    exp_ = exp.(xs .- max_)
-    exp_ ./ sum(exp_, dims=dims)
+softmax(x; dims = 1) = softmax!(similar(x), x; dims)
```
CarloLucibello, line 35: we still support julia < 1.5, so you will have to explicitly pass the keyword value

CarloLucibello: also, this pattern will error on integer array inputs, that we still want to support, so we should promote to float types

src/softmax.jl
```diff
@@ -26,49 +32,20 @@ julia> softmax([1, 2, 3])
...
+softmax!(x; dims = 1) = softmax!(x, x; dims)
+function softmax!(out::T, x::T; dims = 1) where {T<:AbstractArray}
```
CarloLucibello, line 37: we can relax the constraint of `out` and `x` being of the same type

CarloLucibello:
very nice. After addressing the comments, it will be convenient to add some tests with integer inputs

CarloLucibello:
I just noticed https://github.com/FluxML/NNlib.jl/issues/248, I'm not sure how to handle it, but thought it was mentioning here

CarloLucibello:
better do #248 in a separate PR where we also bump the minor version

src/softmax.jl
norci: I have updated the constraint for `softmax!` `logsoftmax!`.
For `∇softmax` , I guess `Δ` and `out` should be the same type. Right?

I added only one test case for integer input. is this OK?
```julia
@test softmax(Int[0, 0]) == [0.5, 0.5]
```

src/softmax.jl
norci: fixed.

CarloLucibello:
You could add a test for #249 and we should be ready

norci:
> You could add a test for #249 and we should be ready

Done.

PR #250, opened 2020-12-21
new API grad softmax

### LINKED ISSUE #248: softmax gradients inefficient, API change necessary
The `∇softmax!`, `∇softmax`, `∇logsoftmax!`, `∇logsoftmax` functions recalculate `softmax(x)` which is inefficient (the forward pass already produced the result. I implemented a new api in my [CUDNN PR](https://github.com/denizyuret/CUDA.jl/blob/dy/cudnn/lib/cudnn/nnlib.jl) where these functions get an additional `y` argument with some backward compatible methods with warnings. For backward compatibility you can make this argument optional and default to `y=(log)softmax(x)` but at least we should have the option of passing `y`.

CarloLucibello:
fix #248

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,0 +1,9 @@
+
+### v0.8 Deprecations
+@deprecate ∇softmax(Δ, x; dims=1)   ∇softmax(Δ, x, softmax(x, dims=dims); dims=dims)
+@deprecate ∇softmax!(Δ, x; dims=1)   ∇softmax!(Δ, Δ, x, softmax(x, dims=dims); dims=dims)
+@deprecate ∇softmax!(out, Δ, x; dims=1)   ∇softmax!(out, Δ, x, softmax(x, dims=dims); dims=dims)
+
+@deprecate ∇logsoftmax(Δ, x; dims=1)   ∇logsoftmax(Δ, x, logsoftmax(x, dims=dims); dims=dims)
+@deprecate ∇logsoftmax!(Δ, x; dims=1)   ∇logsoftmax!(Δ, Δ, x, logsoftmax(x, dims=dims); dims=dims)
+@deprecate ∇logsoftmax!(out, Δ, x; dims=1)   ∇logsoftmax!(out, Δ, x, logsoftmax(x, dims=dims); dims=dims)
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,9 +1,1 @@
 ### v0.8 Deprecations
-
-@deprecate ∇softmax(Δ, x; dims=1)   ∇softmax(Δ, x, softmax(x, dims=dims); dims=dims)
-@deprecate ∇softmax!(Δ, x; dims=1)   ∇softmax!(Δ, Δ, x, softmax(x, dims=dims); dims=dims)
-@deprecate ∇softmax!(out, Δ, x; dims=1)   ∇softmax!(out, Δ, x, softmax(x, dims=dims); dims=dims)
-
-@deprecate ∇logsoftmax(Δ, x; dims=1)   ∇logsoftmax(Δ, x, logsoftmax(x, dims=dims); dims=dims)
-@deprecate ∇logsoftmax!(Δ, x; dims=1)   ∇logsoftmax!(Δ, Δ, x, logsoftmax(x, dims=dims); dims=dims)
-@deprecate ∇logsoftmax!(out, Δ, x; dims=1)   ∇logsoftmax!(out, Δ, x, logsoftmax(x, dims=dims); dims=dims)
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,5 +1,4 @@
-
-### v0.8 deprecations
+### v0.8 Deprecations
 
 @deprecate ∇softmax(Δ, x; dims=1)   ∇softmax(Δ, x, softmax(x, dims=dims); dims=dims)
 @deprecate ∇softmax!(Δ, x; dims=1)   ∇softmax!(Δ, Δ, x, softmax(x, dims=dims); dims=dims)
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,5 +1,6 @@
 
-### v0.8 Deprecations
+### v0.8 deprecations
+
 @deprecate ∇softmax(Δ, x; dims=1)   ∇softmax(Δ, x, softmax(x, dims=dims); dims=dims)
 @deprecate ∇softmax!(Δ, x; dims=1)   ∇softmax!(Δ, Δ, x, softmax(x, dims=dims); dims=dims)
 @deprecate ∇softmax!(out, Δ, x; dims=1)   ∇softmax!(out, Δ, x, softmax(x, dims=dims); dims=dims)
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -22,15 +22,28 @@ with additional manipulations enhancing numerical stability.
 For a matrix input `x` it will by default (`dims=1`) treat it as a batch of vectors,
 with each column independent. Keyword `dims=2` will instead treat rows independently,
 etc...
-```julia-repl
+
+See also [`logsoftmax`](@ref).
+
+** Usage Examples**
+
+```jldoctest
 julia> softmax([1, 2, 3])
 3-element Array{Float64,1}:
   0.0900306
   0.244728
   0.665241
-```
 
-See also [`logsoftmax`](@ref).
+julia> softmax([1 2 3; 2 2 2])  # dims=1
+2×3 Array{Float64,2}:
+ 0.268941  0.5  0.731059
+ 0.731059  0.5  0.268941
+
+julia> softmax([1 2 3; 2 2 2]; dims=2)
+2×3 Array{Float64,2}:
+ 0.0900306  0.244728  0.665241
+ 0.333333   0.333333  0.333333
+```
 """
 softmax(x; dims = 1) = softmax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)
 
@@ -41,24 +54,17 @@ function softmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArr
     out ./= sum(out; dims = dims)
 end
 
-∇softmax(Δ, x, y; dims = 1) = ∇softmax!(similar(Δ), Δ, x, y; dims = dims)
+∇softmax(Δ::AbstractArray{T}, x::AbstractArray, y::AbstractArray{S}; dims = 1) where {T,S} = 
+    ∇softmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
 
 ## Can introduce at the end of deprecation cycle of ∇softmax!(out, Δ, x; dims = 1)  
 #∇softmax!(Δ, x, y; dims = 1) = ∇softmax!(Δ, Δ, x, y; dims = dims)
 
 function ∇softmax!(out::AbstractArray, Δ::AbstractArray, 
                     x::AbstractArray, y::AbstractArray; dims = 1)
-    out .*= Δ .- sum(Δ .* y; dims = dims)
+    out .= Δ .* y
+    out .= out .- y .* sum(out; dims = dims)
 end
-
-
-# ∇softmax(Δ, x; dims = 1) = ∇softmax!(similar(Δ), Δ, x; dims = dims)
-# ∇softmax!(Δ, x; dims = 1) = ∇softmax!(Δ, Δ, x; dims = dims)
-# function ∇softmax!(out::O, Δ::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
-#     softmax!(out, x; dims = dims)
-#     out .*= Δ .- sum(Δ .* out; dims = dims)
-# end
-
 
 """
     logsoftmax(x; dims=1)
@@ -80,12 +86,12 @@ logsoftmax!(x; dims = 1) = logsoftmax!(x, x; dims = dims)
 function logsoftmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
     out .= x .- maximum(x; dims = dims)
     # out .-= log.(sum(exp.(out); dims = dims))  # WARN: this will decrease performance.
-    log_ = log.(sum(exp.(out); dims = dims))
+    log_ = log.(sum(exp, out; dims = dims))
     out .-= log_
 end
 
-
-∇logsoftmax(Δ, x, y; dims = 1) = ∇logsoftmax!(similar(Δ), Δ, x, y; dims = dims)
+∇logsoftmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
+    ∇logsoftmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
 
 ## Can introduce at the end of deprecation cycle of ∇logsoftmax!(out, Δ, x; dims = 1)  
 # ∇logsoftmax!(Δ, x, y; dims = 1) = ∇logsoftmax!(Δ, Δ, x, y; dims = dims)
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -90,7 +90,7 @@ function logsoftmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:Abstract
     out .-= log_
 end
 
-∇logsoftmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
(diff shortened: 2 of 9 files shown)


src/softmax.jl
```diff
@@ -61,23 +74,25 @@ It is semantically equivalent to the following:
...
 """
 logsoftmax(x; dims = 1) = logsoftmax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)
+
 logsoftmax!(x; dims = 1) = logsoftmax!(x, x; dims = dims)
+
 function logsoftmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
     out .= x .- maximum(x; dims = dims)
     # out .-= log.(sum(exp.(out); dims = dims))  # WARN: this will decrease performance.
     log_ = log.(sum(exp.(out); dims = dims))
     out .-= log_
```
mcabbott, line 84: Maybe you can save a few allocations here?
```suggestion
    tmp_ = sum(exp, out; dims = dims)
    tmp_ .= log.(tmp_)
    out .-= tmp_
```

src/softmax.jl
```diff
@@ -33,20 +33,33 @@ julia> softmax([1, 2, 3])
 See also [`logsoftmax`](@ref).
 """
```
mcabbott, line 34: May I suggest for the docstring here:
```
```jldoctest
julia> softmax([1, 2, 3])
3-element Array{Float64,1}:
  0.0900306
  0.244728
  0.665241

julia> softmax([1 2 3; 2 2 2])  # dims=1
2×3 Array{Float64,2}:
 0.268941  0.5  0.731059
 0.731059  0.5  0.268941

julia> softmax([1 2 3; 2 2 2]; dims=2)
2×3 Array{Float64,2}:
 0.0900306  0.244728  0.665241
 0.333333   0.333333  0.333333
```

src/softmax.jl
```diff
@@ -33,20 +33,33 @@ julia> softmax([1, 2, 3])
...
-    softmax!(out, x; dims = dims)
-    out .*= Δ .- sum(Δ .* out; dims = dims)
+∇softmax(Δ, x, y; dims = 1) = ∇softmax!(similar(Δ), Δ, x, y; dims = dims)
+
+## Can introduce at the end of deprecation cycle of ∇softmax!(out, Δ, x; dims = 1)  
+#∇softmax!(Δ, x, y; dims = 1) = ∇softmax!(Δ, Δ, x, y; dims = dims)
+
+function ∇softmax!(out::AbstractArray, Δ::AbstractArray, 
+                    x::AbstractArray, y::AbstractArray; dims = 1)
+    out .*= Δ .- sum(Δ .* y; dims = dims)
```
mcabbott, line 51: Don't you need `out .= y` here too? Since `∇softmax!(similar(Δ), Δ, x, y; dims = dims)` produces garbage. 

I also wonder whether it would be safe to do `sum(y .= Δ .* y; dims = dims)` to save the allocation? By the time we are executing this, everything downstream of `y` ought to be finished.

src/softmax.jl
```diff
@@ -33,20 +33,33 @@ julia> softmax([1, 2, 3])
...
     out .= exp.(x .- maximum(x; dims = dims))
     out ./= sum(out; dims = dims)
 end
 
-∇softmax(Δ, x; dims = 1) = ∇softmax!(similar(Δ), Δ, x; dims = dims)
-∇softmax!(Δ, x; dims = 1) = ∇softmax!(Δ, Δ, x; dims = dims)
-function ∇softmax!(out::O, Δ::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
-    softmax!(out, x; dims = dims)
-    out .*= Δ .- sum(Δ .* out; dims = dims)
+∇softmax(Δ, x, y; dims = 1) = ∇softmax!(similar(Δ), Δ, x, y; dims = dims)
```
mcabbott, line 44:
```suggestion
∇softmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
    ∇softmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
```
Sometimes you might get `Δ::Fill` or `Δ::BitArray`, maybe not so common but may as well.

src/softmax.jl
```diff
@@ -61,23 +74,25 @@ It is semantically equivalent to the following:
...
-∇logsoftmax!(Δ, x; dims = 1) = ∇logsoftmax!(Δ, Δ, x; dims = dims)
-function ∇logsoftmax!(
-    out::O,
-    Δ::O,
-    x::T;
-    dims = 1,
-) where {O<:AbstractArray,T<:AbstractArray}
-    out .= Δ .- sum(Δ, dims = dims) .* softmax!(out, x; dims = dims)
+
+∇logsoftmax(Δ, x, y; dims = 1) = ∇logsoftmax!(similar(Δ), Δ, x, y; dims = dims)
```
mcabbott, line 88:
```suggestion
∇ logsoftmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
    ∇ logsoftmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
```

src/softmax.jl
```diff
@@ -61,23 +75,26 @@ It is semantically equivalent to the following:
...
-function ∇logsoftmax!(
-    out::O,
-    Δ::O,
-    x::T;
-    dims = 1,
-) where {O<:AbstractArray,T<:AbstractArray}
-    out .= Δ .- sum(Δ, dims = dims) .* softmax!(out, x; dims = dims)
+
+∇ logsoftmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
+    ∇ logsoftmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
```
mcabbott, line 90: I think the spaces are my copy-paste error, sorry!
```suggestion
∇logsoftmax(Δ:: AbstractArray{T}, x:: AbstractArray, y:: AbstractArray{S}; dims = 1) where {T,S} = 
    ∇logsoftmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
```

src/softmax.jl
```diff
@@ -61,23 +75,26 @@ It is semantically equivalent to the following:
...
 logsoftmax!(x; dims = 1) = logsoftmax!(x, x; dims = dims)
+
 function logsoftmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
     out .= x .- maximum(x; dims = dims)
-    # out .-= log.(sum(exp.(out); dims = dims))  # WARN: this will decrease performance.
-    log_ = log.(sum(exp.(out); dims = dims))
-    out .-= log_
+    red = sum(exp, out; dims = dims)
+    red .= log.(red)
+    out .-= red
```
mcabbott, line 85: The flaw in my bright idea here is that if `dims=:` then `sum(exp, rand(3), dims=:) isa Number`, so writing the `log.(red)` back into it won't work. (Although `dims=1:N` would be OK.) Sorry!

What's the least inelegant way to fix this? One way is
```suggestion
    sum_ = sum(exp, out; dims = dims)
    log_ = ndims(sum_) > 0 ? red .= log.(sum_) : log.(sum_)
    out .-= log_
```
and another is just to give up on recycling that array, and just say `log_  = log.(sum_)`.

DhairyaLGandhi:
This seems like very inelegant API, there also seem to be a few cases it needs to resolve per the reviews so far. I wouldn't want to have this in in its current state, could you explain why the API change is necessary? We shouldn't need to change the API for this particular task. It also seems to need extra computation of the softmax?

I would suggest we fix #248 first, and that would supersede issues here.

CarloLucibello:
> This seems like very inelegant API, there also seem to be a few cases it needs to resolve per the reviews so far. I wouldn't want to have this in in its current state, could you explain why the API change is necessary? We shouldn't need to change the API for this particular task. It also seems to need extra computation of the softmax?
> 
> I would suggest we fix #248 first, and that would supersede issues here.

*This* is the fix to #248. I just have to cleanup the PR. The API change is necessary because previous adjoints were very inefficient, had to recompute the output. 
I don't know why you say this is inelegant, but in any case, the API is exactly as inelegant as it was before.

src/softmax.jl
CarloLucibello: good catch

src/softmax.jl
CarloLucibello: yeah maybe better not trying to be too smart here, also the  logsum output has reduced size, so probably no big deal

DhairyaLGandhi:
We can pass in the output without recomputing it via adjoint or rrules. So it should close over the outputs.

CarloLucibello:
the gradients have to be separate functions, they are overloaded in CUDA.jl

CarloLucibello:
a Zygote's gradient test is inevitably gonna fail here, so either I remove the test or steal softmax's rules from #242  and adapt them to this PR. I'm going with the latter, plus port the tests to ChainRulesTestUtils, sorry for the conflicts this if going cause in #242  @simeonschaub

DhairyaLGandhi:
> This is the fix to #248. I just have to cleanup the PR. The API change is necessary because previous adjoints were very inefficient, had to recompute the output.

Why not close over the output. And anyway the problem is also that softmax is inefficient with our current punting to CuDNN. Deniz's PR to cuda would help here.

If you need code from #242, we can wait till that is to be merged then

CarloLucibello:
> Why not close over the output. And anyway the problem is also that softmax is inefficient with our current punting to CuDNN. Deniz's PR to cuda would help here.

As I already said, the gradients have to be separate functions, they are overloaded in CUDA.jl. 
And this already goes  in the direction requested by Deniz in #248 and already adopted in his PR (looks at the changes in lib/cudnn/nnlib.jl there)

DhairyaLGandhi:
I believe that the changes are yet to be finalized

CarloLucibello:
This how NNlib has always exported gradients, it's not just softmax, see also e.g. `conv`, `maxpool`....
Deniz's #248 is just an obvious performance fix. 
rrules will define closures containing these gradient primitives. 
If you have a radically different proposal, please open an issue and formalize it.

DhairyaLGandhi:
Is that a response to something I said?

CarloLucibello:
well about the whole discussion, using closures, inelegant API, or the last comment "I believe that the changes are yet to be finalized" (as if you want to suggest something different). I'm under the impression that you would like to adopt a different API but I don't see any concrete proposal

DhairyaLGandhi:
The pr is being worked upon, and we will converge on an api for it and while it might end up being somewhat similar to what you implemented here, but there's no reason push for this until we have converged on the specifics. My last comment in motivated by the regressions we notice with the pr in question, which are key to pursuing the correct design. I don't think I questioned exporting anything, or for performance fixes for that matter.

src/softmax.jl
mcabbott: I guess one could use `mapreduce(*, +,   Δ, y; dims)` to avoid `Δ .* y`... although right now this doesn't do anything efficient, sadly. (e.g. `mapreduce(*, +, fill(3, 2,2), fill(5, 2,2), dims=1)`.)

src/softmax.jl
mcabbott: Indeed, there are a bunch of O(N^2) allocations (for matrix input) so perhaps one O(N) doesn't matter at all.

src/deprecations.jl
```diff
@@ -0,0 +1,10 @@
+
+### v0.8 deprecations
+
+@deprecate ∇softmax(Δ, x; dims=1)   ∇softmax(Δ, x, softmax(x, dims=dims); dims=dims)
+@deprecate ∇softmax!(Δ, x; dims=1)   ∇softmax!(Δ, Δ, x, softmax(x, dims=dims); dims=dims)
+@deprecate ∇softmax!(out, Δ, x; dims=1)   ∇softmax!(out, Δ, x, softmax(x, dims=dims); dims=dims)
+
+@deprecate ∇logsoftmax(Δ, x; dims=1)   ∇logsoftmax(Δ, x, logsoftmax(x, dims=dims); dims=dims)
+@deprecate ∇logsoftmax!(Δ, x; dims=1)   ∇logsoftmax!(Δ, Δ, x, logsoftmax(x, dims=dims); dims=dims)
+@deprecate ∇logsoftmax!(out, Δ, x; dims=1)   ∇logsoftmax!(out, Δ, x, logsoftmax(x, dims=dims); dims=dims)
```
simeonschaub, line 10: I wonder whether these should really be deprecated. I could still see situations, where it's useful to just get the gradient without the forwards pass. My first thought was to maybe make `y` an optional keyword argument instead, but don't have a very strong opinion on this.

test/softmax.jl
```diff
@@ -48,8 +50,12 @@ end
...
         0.69246 0.569494 -0.503191 -0.925947 -0.0870738 -1.0697
     ]
-    @test ∇logsoftmax(ones(size(xs)), xs) ≈ ys rtol = 1e-6
-    @test ∇softmax(ones(size(xs)), xs) ≈ zeros(size(xs)) atol = 1e-6
+    
+    y = logsoftmax(xs)
+    @test ∇logsoftmax(ones(size(xs)), xs, y) ≈ ys rtol = 1e-6
+    
+    y = softmax(xs)
+    @test ∇softmax(ones(size(xs)), xs, y) ≈ zeros(size(xs)) atol = 1e-6
```
simeonschaub, line 58: It would be good to test with more random inputs as well, since just using ones can hide mistakes in the implementation. Or was your thought that this is already covered in #242?

test/softmax.jl
```diff
@@ -69,10 +75,22 @@ end
 
         map([zeros, ones]) do fn
```
simeonschaub, line 76:
```suggestion
        @testset "$fn(Float64, $(size(xs)))" for fn in [zeros, ones, rand]
```

test/test_utils.jl
```diff
@@ -0,0 +1,15 @@
+"""
+Compare numerical and automatic gradient.
+`f` has to be a scalar valued function.
+
+Use `ChainRulesTestUtils.rrule_test` instead 
+if the rrule is explicitly defined.
+"""
+function autodiff_test(f, xs...; atol=1e-9, rtol=1e-9)
```
simeonschaub, line 8: I wasn't sure at first why this was needed. Perhaps `zygote_test_gradient` might be a bit clearer as a name?

test/test_utils.jl
```diff
@@ -0,0 +1,15 @@
...
+    fdm = FiniteDifferences.central_fdm(5, 1)
+    gs_ad = Zygote.gradient(f, xs...)
+    gs_fd = FiniteDifferences.grad(fdm, f, xs...) 
+    for (g_ad, g_fd) in zip(gs_ad, gs_fd)
+        @test g_ad ≈ g_fd   atol=atol rtol=rtol
+    end
+end
```
simeonschaub, line 15:
```suggestion
end

```

simeonschaub (commented):
Overall looks sensible to me, just some comments on a few details

src/deprecations.jl
mcabbott: Since `y` is the last argument, can't it simply have a default value? I mean permanently, just leave `∇softmax` having a 2-arg and 3-arg method.

CarloLucibello: also @DhairyaLGandhi is in favor of the optional `y`, will change

simeonschaub: Yeah, sure, that's possible as well. I feel like a keyword argument might probably leave less ambiguity as to which argument specifies what, but would be fine with an optional positional argument as well.

CarloLucibello: still, I'm not totally convinced: in many situations people could be computing the forward twice without noticing. I think that the benefit of a deprecation warn in these situations outweighs the slight annoyance of having to explicitly compute and pass the forward when needing only the gradient

test/softmax.jl
CarloLucibello: I cannot change this since the r.h.s. is explicitly given, but I'am also indirectly testing this with random inputs down below when using `rrule_test`. I think this should be sufficient, even if not considering #242

src/deprecations.jl
CarloLucibello: any further comment here? If someone has a strong opinion into keeping the 2-arg method I can reinstate it, otherwise I'd leave the deprecation

CarloLucibello: I added back the 2-arg version

CarloLucibello:
@simeonschaub A few more tests are passing now, I don't know why, I did only some reorg

simeonschaub:
The question is whether that's a good thing or a bad thing 😂

CarloLucibello:
> The question is whether that's a good thing or a bad thing

aahaha, that's terrible, that's exactly the kind of shady things that make Flux unusable for serious work

PR #255, opened 2020-12-26
Add scatter operations

yuehhua:
I have generalized scatter operations to every dimensions, instead of excluding the first dimension.
Related to yuehhua/ScatterNNlib.jl#32


mcabbott:
Could these easily be combined to something like `scatter!(+, dst, indices, src)`?

yuehhua:
@mcabbott Sure! I have API like `scatter!(:add, dst, indices, src)` in my ScatterNNlib.jl. Should I move it here as well?

CarloLucibello:
> I have API like scatter!(:add, dst, indices, src) in my ScatterNNlib.jl. Should I move it here as well?

I think he is suggesting to directly pass the operator (not  a symbol), the same way you woulld use it in 
```julia
julia> reduce(+, ones(10))
10.0
```

src/utils.jl
```diff
@@ -0,0 +1,6 @@
+"""
+    save_div(x, y)
+
+Savely divde `x` by `y`. If `y` is zero, return `x` directly.
```
CarloLucibello, line 4:
```suggestion
Safely divide `x` by `y`. If `y` is zero, return `x` directly.
```

src/utils.jl
```diff
@@ -0,0 +1,6 @@
...
+"""
+save_div(x, y) = ifelse(iszero(y), x, x/y)
```
CarloLucibello, line 6:
```suggestion
safe_div(x, y) = ifelse(iszero(y), x, x/y)
```

src/scatter.jl
```diff
@@ -0,0 +1,193 @@
...
+
+The dimension of `us` must equal to dimension of `xs`. `ys`, `us` and `xs` must be
+supported array type and be the same type.`Array`, `StaticArray` and `CuArray`
+are currently supported.
+"""
+function scatter_div! end
+
+for op = [:add, :sub, :mul, :div]
+    fn = Symbol("scatter_$(op)!")
+    @eval function $fn(ys::Array{T}, us::Array{T}, xs::Array{<:IntOrTuple}) where {T<:Real}
```
CarloLucibello, line 90: Can the signatures be relaxed to AbstractArrays?

src/scatter.jl
```diff
@@ -0,0 +1,193 @@
...
+The dimension of `us` must equal to dimension of `xs`. `ys`, `us` and `xs` must be
+supported array type and be the same type.`Array`, `StaticArray` and `CuArray`
+are currently supported.
+"""
+function scatter_min!(ys::Array{T}, us::Array{T}, xs::Array{<:IntOrTuple}) where {T<:Real}
+    @simd for k = 1:length(xs)
+        k = CartesianIndices(xs)[k]
+        ys_v = view(ys, xs[k]...)
+        us_v = view(us, k)
+        @inbounds ys_v .= min.(ys_v, us_v)
```
CarloLucibello, line 152: why `max` and `min` are not included in the loop above?

src/scatter.jl
yuehhua: Done.

src/utils.jl

src/utils.jl

src/scatter.jl
```diff
@@ -0,0 +1,46 @@
...
+- `xs`: the mapping for aggregation from source (index) to destination (value).
+The index of `xs` is corresponding to the index of `us` and the value of `xs` is
+corresponding to the index of `ys`. The value of `xs` can be `Int` or `Tuple` type.
+
+The dimension of `us` must equal to dimension of `xs`. `ys`, `us` and `xs` must be
+supported array type and be the same type.`Array`, `StaticArray` and `CuArray`
+are currently supported.
+"""
+function scatter!(op, ys::AbstractArray{T}, us::AbstractArray{T}, xs::AbstractArray{<:IntOrTuple}) where {T<:Real}
+    @simd for k = 1:length(xs)
```
CarloLucibello, line 25:
```suggestion
    @simd for k in CartesianIndices(xs)
```

src/scatter.jl
```diff
@@ -0,0 +1,46 @@
...
+
+"""
+    scatter!(op, ys, us, xs)
+
+Scatter operation.
+
+    ys[xs[k]...] = (op).(ys[xs[k]...], us[k...])
+
+# Arguments
+- `op`: operations to be applied on `ys` and `us`, e.g. `+`, `-`, `*`, `/`, `max`, `min` and `mean`.
```
CarloLucibello, line 13: since `mean` will be doing something different from the line of code, we should mention it separately

src/scatter.jl
```diff
@@ -0,0 +1,46 @@
+export scatter!
+
+const IntOrTuple = Union{Integer,Tuple}
+
+"""
+    scatter!(op, ys, us, xs)
+
+Scatter operation.
```
CarloLucibello, line 8:
```suggestion
Scatter operation. For each index `k` in `xs`, accumulate values in `ys` according to
```

src/scatter.jl
```diff
@@ -0,0 +1,46 @@
...
+    end
+    ys
+end
+
+function scatter!(op::typeof(mean), ys::AbstractArray{T}, us::AbstractArray{T}, xs::AbstractArray{<:IntOrTuple}) where {T<:Real}
+    Ns = zero(ys)
+    ys_ = zero(ys)
+    scatter!(+, Ns, one.(us), xs)
+    scatter!(+, ys_, us, xs)
+    ys .+= safe_div.(ys_, Ns)
```
CarloLucibello, line 39: Should this add to `ys` or just assign?

src/utils.jl
```diff
@@ -0,0 +1,6 @@
+"""
+    safe_div(x, y)
+
+Safely divde `x` by `y`. If `y` is zero, return `x` directly.
```
CarloLucibello, line 4:
```suggestion
Safely divide `x` by `y`. If `y` is zero, return `x` directly.
```

src/scatter.jl
```diff
@@ -0,0 +1,64 @@
...
+
+The dimension of `src` must equal to dimension of `idx`. `dst`, `src` and `idx` must be
+supported array type and be the same type.`Array`, `StaticArray` and `CuArray`
+are currently supported.
+"""
+function scatter!(op, dst::AbstractArray{T}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple}) where {T<:Real}
+    @simd for k in CartesianIndices(idx)
+        dst_v = view(dst, idx[k]...)
+        us_v = view(src, k)
+        @inbounds dst_v .= (op).(dst_v, us_v)
```
mcabbott, line 29: What is broadcasting doing here? If `idx` must contain integers or tuples of integers, and then this is scalar, and could just be `setindex!`.

yuehhua: It is still array, not scalar. If I take off broadcasting here, tests will not pass.

mcabbott: But it's always a one-element, zero-dimensional array? If so, does this have advantages over dealing with the scalars?

src/scatter.jl
```diff
@@ -0,0 +1,65 @@
...
+corresponding to the index of `dst`. The value of `idx` can be `Int` or `Tuple` type.
+
+The dimension of `src` must equal to dimension of `idx`. `dst`, `src` and `idx` must be
+supported array type and be the same type.
+"""
+function scatter!(op, dst::AbstractArray{T}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple}) where {T<:Real}
+    @simd for k in CartesianIndices(idx)
+        dst_v = view(dst, idx[k]...)
+        src_v = view(src, k)
+        @inbounds dst_v .= (op).(dst_v, src_v)
```
chengchingwen, line 29:
```suggestion
@inbounds begin
  id = idx[k]
  dst_v = dst[id...]
  dst_v[id...] = op(dst_v, src[k])
end
```

chengchingwen: I forget the indentation.

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+# Arguments
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather(src::AbstractArray{T}, idx::AbstractArray{<:Integer}) where {T}
+    dst = Array{T}(undef, size(idx)...)
+    gather!(dst, src, idx)
+end
+
+function gather_indices(X::Array{T}) where T
```
chengchingwen, line 43: What is this for?

src/scatter.jl
```diff
@@ -0,0 +1,65 @@
...
+# Arguments
+- `op`: operations to be applied on `dst` and `src`, e.g. `+`, `-`, `*`, `/`, `max` and `min`.
+- `dst`: the destination for `src` to aggregate to. This argument will be mutated.
+- `src`: the source data for aggregating.
+- `idx`: the mapping for aggregation from source (index) to destination (value).
+The index of `idx` is corresponding to the index of `src` and the value of `idx` is
+corresponding to the index of `dst`. The value of `idx` can be `Int` or `Tuple` type.
+
+The dimension of `src` must equal to dimension of `idx`. `dst`, `src` and `idx` must be
+supported array type and be the same type.
```
chengchingwen, line 23: Should add some check

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+
+# Arguments
+- `dst`: the destination where data would be assigned to.
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather!(dst::AbstractArray{T,N}, src::AbstractArray{T}, idx::AbstractArray{<:Integer,N}) where {T,N}
+    @assert size(dst) == size(idx) "dst and idx must have the same size."
+    @simd for k = CartesianIndices(idx)
+        @inbounds view(dst, k) .= view(src, idx[k]...)
```
chengchingwen, line 19: `k` is only `Integer` here, so no `...` needed.

```suggestion
        @inbounds dst[k] = src[idx[k]]
```

mcabbott: Shouldn't gather accept `idx::AbstractArray{<:IntOrTuple}` for symmetry? In which case the splat is needed. 

But the views seem strange, see my comment somewhere... unless the intention is to allow e.g. `idx = [(1,:), (3,:)]`?

chengchingwen:
Maybe we should specified how our `scatter`/`gather` different from TF or Torch?

src/gather.jl
chengchingwen: You are right, this should accept `IntOrTuple`, then the splat is actually needed. I'll update the suggestion.

I don't think something like `idx = [(1, :), (3,:)]` would be allowed. The semantic is unclear.

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+For each index `k` in `idx`, assign values to `dst` according to
+
+    dst[k...] = src[idx[k]...]
+
+# Arguments
+- `dst`: the destination where data would be assigned to.
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather!(dst::AbstractArray{T,N}, src::AbstractArray{T}, idx::AbstractArray{<:Integer,N}) where {T,N}
```
chengchingwen, line 16:
```suggestion
function gather!(dst::AbstractArray{T,N}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple,N}) where {T,N}
```

yuehhua:
I discussed with @chengchingwen yesterday. We hope to introduce some new features to generalize scatter/gather operations. New features are listed in the following:

- [ ]  Add `dims` argument for specifying which dimension to start scattering and the dimensions will be contiguous block, like [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/scatter_nd) does.
- [ ]  The operational unit of scatter/gather could be a scalar or (at least) a vector. Both of them will be provided. The scalar unit will be designed by assigning `dims=:`, while the vector unit will be assigning with integer `dims=1`.
- [ ] Following above, `dims=1` will be a default option.

src/scatter.jl
yuehhua: I try to figure out what is happening here. It should be scalar.

src/gather.jl
yuehhua: `CartesianIndex` is not tuple, so it doesn't need unpacking `k...`.

src/gather.jl
yuehhua: Check new commit please.

src/gather.jl
yuehhua: Check new commit please.

src/scatter.jl
```diff
@@ -0,0 +1,94 @@
...
+    @inbounds for k in CartesianIndices(idx)
+        dst[idx[k]...] = op(dst[idx[k]...], src[k])
+    end
+    dst
+end
+
+function scatter_vec!(op, dst::AbstractArray{T}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple},
+                      dims::Integer) where {T<:Real}
+    @simd for k in CartesianIndices(idx)
+        dst_v = view(dst, colons(dims)..., idx[k]...)
```
chengchingwen, line 49:
```suggestion
        dst_v = view(dst, Base.ntuple(_->Colon(), dims)..., idx[k]...)
```
better with tuple

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+For each index `k` in `idx`, assign values to `dst` according to
+
+    dst[k...] = src[idx[k]...]
+
+# Arguments
+- `dst`: the destination where data would be assigned to.
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather!(dst::AbstractArray{T,N}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple,N}) where {T,N}
```
chengchingwen, line 16: also need `dims` here

src/utils.jl
```diff
@@ -0,0 +1,19 @@
...
+Safely divide `x` by `y`. If `y` is zero, return `x` directly.
+"""
+safe_div(x, y) = ifelse(iszero(y), x, x/y)
+
+"""
+    least_dims(idxs)
+
+Compute the least dimensions, of which array can be accessed by the indices `idxs`.
+"""
+least_dims(idxs::AbstractArray{<:Integer}) = (maximum(idxs), )
```
chengchingwen, line 13: why do we need this?

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+    @assert size(dst) == size(idx) "dst and idx must have the same size."
+    for k = CartesianIndices(idx)
+        @inbounds dst[k] = src[idx[k]]
```
mcabbott, line 19: Before `@inbounds`, should we check? Possibly like this you can still say `@inbounds gather(...` to disable it:
```
@boundscheck _checkin(idx, src)

_checkin(idx::AbstractArray{<:Integer}, arr) = 
    minimum(idx) in LinearIndices(arr) && maximum(idx) in LinearIndices(arr) || throw("some bounds error")
    
function _checkin(idx::AbstractArray{Tuple}, arr)
    for d in 1:ndims(arr)
        maximum(I->I[d], idx) in axes(arr,d) || error()
    end
end 
``` 
Surely there's a nicer way of treating the tuple case than my sketch here.

src/gather.jl
```diff
@@ -0,0 +1,49 @@
...
+    dst[k...] = src[idx[k]...]
+
+where destination `dst` is created according to `idx`.
+
+# Arguments
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather(src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple}) where {T}
+    dst = Array{T}(undef, size(idx)...)
```
mcabbott, line 39: I think this should be
```suggestion
    dst = similar(src, axes(idx)...)
```
which will allow for both CuArrays and OffsetArrays. 

It might be worth changing all other `size` to `axes` too, no reason not to allow offsets when it's easy.

src/scatter.jl
```diff
@@ -0,0 +1,94 @@
...
+Examples for dims are lists here:
+
+- `dims=0`: dst[idx[k]...] = op(dst[idx[k]...], src[k...])
+- `dims=1`: dst[:, idx[k]...] = (op).(dst[:, idx[k]...], src[k...])
+- `dims=2`: dst[:, :, idx[k]...] = (op).(dst[:, :, idx[k]...], src[k...])
+"""
+function scatter!(op, dst::AbstractArray{T}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple},
+                  dims::Integer=1) where {T<:Real}
+    if dims > 0
+        scatter_vec!(op, dst, src, idx, dims)
```
mcabbott, line 31:
```suggestion
        scatter_vec!(op, dst, src, idx, Val(dims))
```
Since `dims` affects the types, maybe it should be passed as Val.  Then `scatter_vec!` which actually does the work can be type-stable.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.8.0"
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
```
CarloLucibello, line 8: we are trying to stay lean, so we should avoid adding this dependency if possible

yuehhua: Sure! I will try to avoid it.

src/scatter.jl
mcabbott: Note that `scatter_vec!` will also have to accept the argument like `, ::Val{dim}) where {dim}` instead of `, dim::Int)` for this to work.

src/gather.jl
yuehhua: Check new commits please.

src/scatter.jl
yuehhua: Check new commits please.

src/gather.jl
yuehhua: Check new commits please.

src/utils.jl
```diff
@@ -0,0 +1,32 @@
...
+function least_dims(idxs::AbstractArray{<:Tuple})
+    Tuple(maximum(xs) for xs in zip(idxs...))
+end
+
+_check_input(idx::AbstractArray{<:Integer}, arr) = checkbounds(arr, minimum(idx):maximum(idx))
+
+function _check_input(idx::AbstractArray{Tuple}, arr)
+    pairs = map(xs -> Base.OneTo(maximum(xs)), zip(idx...))
+    checkbounds(arr, pairs...)
+end
```
mcabbott, line 24: Does this work?

I think you want something like this, I haven't tried but believe it ought to work on the GPU, someone should check though:
```
idx = [(1,2), (3,4), (2,100), (-100,3)]

lows, highs = reduce(idx; init=(first(idx), first(idx))) do (lows, highs), xs
    map(min, xs, lows), map(max, xs, highs)
end

checkbounds(arr, CartesianIndex(lows))
checkbounds(arr, CartesianIndex(highs))
```
Also, with `dims` this won't be right, as the check should be only against some of the axes of the array. So perhaps the function should be more like `_check_bounds(axes(arr)[dims:end], idx)` --- pass the axes not the whole array.

src/gather.jl
```diff
@@ -0,0 +1,86 @@
+export gather, gather!
+
+"""
+    gather!(dst, src, idx, dims)
```
mcabbott, line 4: Should `dims` be made a keyword arg, on user-facing functions?
```suggestion
    gather!(dst, src, idx; dims=0)
```

yuehhua: Check new commits please.

src/utils.jl
yuehhua: Could you elaborate this clearly? Maybe I just cannot get your point.

If the point is questioning about the functionality of `_check_input` function, we do have test on this.
It should work well or throw exceptions such that the test case should catch exceptions correctly.

If the point is the implementation of checking, not passing the whole array but their axes.
I still not understand the reason passing the whole array would be expensive. Maybe there are some other considerations.

src/utils.jl
yuehhua: Check `scatter` functions please.

DhairyaLGandhi:
bumping on this

PiotrSokol:
@yuehhua , any plans on including CUDA dispatches for scatter ?

yuehhua:
@PiotrSokol Sure! Just let me finish this first. That way, the API can be determined. The CUDA dispatch will be made in CUDA.jl. Sorry for letting everyone waiting so long. I almost finished other tasks, then I can focus on this in a couple of days.

src/gather.jl
```diff
@@ -0,0 +1,78 @@
...
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+- `dims`: the number of dimensions to be view as an operational unit from the beginning.
+If dims=0 is given, operational unit is scalar. Default dims is 1.
+
+Examples for dims are lists here:
+
+- `dims=0`: dst[k...] = src[idx[k]...]
+- `dims=1`: dst[:, k...] .= src[:, idx[k]...]
+- `dims=2`: dst[:, :, k...] .= src[:, :, idx[k]...]
```
CarloLucibello, line 22: do we have to pass a `dims` argument? can't the proper indexing be inferred from the dimensions of `dst` and `src` and from the eltype of `idx`?

src/utils.jl
```diff
@@ -0,0 +1,56 @@
...
+
+Compute the least dimensions, of which array can be accessed by the indices `idxs`.
+"""
+least_dims(idxs::AbstractArray{<:Integer}) = (maximum(idxs), )
+
+function least_dims(idxs::AbstractArray{<:Tuple})
+    Tuple(maximum(xs) for xs in zip(idxs...))
+end
+
+_check_input(idx::AbstractArray{<:Integer}, arr) = checkbounds(arr, minimum(idx):maximum(idx))
```
CarloLucibello, line 19: these check functions are not general utilities, they shouldn't stay in`src/utils.jl` but where they are needed. 
Also, they should have more specific names, like `_check_gather_input`

CarloLucibello:
@yuehhua I took the liberty to rewrite the `gather` implementation and push it here in order to fix some issues, add some tests, speed things up.

I think that `gather` is a first piece of the implementation that is ready, we can merge it even without gradient rules, so that we decompose this huge work into multiple PRs and make it easier to review. I'm going to file a separate PR with just `gather`.

src/gather.jl
CarloLucibello: I implemented my own suggestion and removed `dims` :)

src/utils.jl
CarloLucibello: fixed this

yuehhua:
@CarloLucibello Thank you so much! I was busy then and have time now working on this PR.
Feel free to complete the functionality of `gather`.

src/utils.jl
yuehhua: I need this in checking dimensions in `scatter` as well! I move them back in new coming commits.

CarloLucibello:
> @CarloLucibello I want to separate rrules in another PR avoiding taking too much time on this PR.
So, this PR just focus on the implementation of scatter functions and support mutating and non-mutating version.
That way, we move forward step by step.

This makes perfect sense, makes the process much easier. So now this PR and #280 are basically unrelated, and after merging both we can look into the diff rules

src/scatter.jl
```diff
@@ -0,0 +1,179 @@
...
+If dims=0 is given, operational unit is scalar. Default dims is 1.
+
+Examples for dims are lists here:
+
+- `dims=0`: dst[idx[k]...] = op(dst[idx[k]...], src[k...])
+- `dims=1`: dst[:, idx[k]...] = (op).(dst[:, idx[k]...], src[:, k...])
+- `dims=2`: dst[:, :, idx[k]...] = (op).(dst[:, :, idx[k]...], src[:, :, k...])
+"""
+function scatter!(op, dst::AbstractArray{T}, src::AbstractArray{T}, idx::AbstractArray{<:IntOrTuple};
+                  dims::Integer=1) where {T<:Real}
```
CarloLucibello, line 55: unless I'm missing something, I think we can remove `dims` entirely because essentially it is determined by the other arrays' shapes and eltypes, right? Look at how this is done in #280

Project.toml
```diff
@@ -4,13 +4,15 @@ version = "0.7.11"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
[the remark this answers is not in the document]
yuehhua: OK

src/scatter.jl
yuehhua: I spend couples of days checking the generality of current API is good. I have new commit pushed.

yuehhua:
I give some benchmark about using the performance annotations or not.

With annotations of `@simd` and `@inbounds`,

```julia
using NNlib
using BenchmarkTools

d = 50
nbins = 20
l = 2^20

hist = zeros(Float32, d, nbins)
δ = rand(Float32, d, l)
idx = rand(1:nbins, l)

scatter!(+, hist, δ, idx)
@benchmark scatter!(+, $hist, $δ, $idx)
```

I got

```julia
BenchmarkTools.Trial: 
  memory estimate:  336.00 MiB
  allocs estimate:  7340038
  --------------
  minimum time:     1.191 s (3.17% GC)
  median time:      1.202 s (3.14% GC)
  mean time:        1.205 s (3.12% GC)
  maximum time:     1.220 s (3.03% GC)
  --------------
  samples:          5
  evals/sample:     1
```

Again, I take off the annotations with the same configurations.

```julia
julia> @benchmark scatter!(+, $hist, $δ, $idx)
BenchmarkTools.Trial: 
  memory estimate:  336.00 MiB
  allocs estimate:  7340038
  --------------
  minimum time:     1.209 s (2.99% GC)
  median time:      1.231 s (2.94% GC)
  mean time:        1.258 s (2.97% GC)
  maximum time:     1.361 s (3.03% GC)
  --------------
  samples:          4
  evals/sample:     1
```

Apparently, without annotations, it become slower. However, the difference is not huge. For mean time, it saves about (1.258 - 1.205)/1.258 * 100% = 4.21% in relative time.
I think that it is still worth introducing performance annotations.

src/scatter.jl
```diff
@@ -0,0 +1,189 @@
...
+- `idx`: the mapping for aggregation from source (index) to destination (value).
+The index of `idx` is corresponding to the index of `src` and the value of `idx` is
+corresponding to the index of `dst`. The value of `idx` can be `Int` or `Tuple` type.
+
+    dst[:, ..., idx[k]...] = (op).(dst[:, ..., idx[k]...], src[:, ..., k...])
+"""
+function scatter!(op,
+                  dst::AbstractArray{Tdst,Ndst},
+                  src::AbstractArray{Tsrc,Nsrc},
+                  idx::AbstractArray{<:Integer,Nidx}) where {Tdst<:Real,Tsrc<:Real,Ndst,Nsrc,Nidx}
```
CarloLucibello, line 59:
```suggestion
                  idx::AbstractArray{<:Integer,Nidx}) where {Tdst,Tsrc,Ndst,Nsrc,Nidx}
```
this restriction is not needed, we can support scatter acting on non-numerical arrays

src/scatter.jl
```diff
@@ -0,0 +1,189 @@
...
+
+function scatter!(op,
+                  dst::AbstractArray{Tdst,Ndst},
+                  src::AbstractArray{Tsrc,Nsrc},
+                  idx::AbstractArray{NTuple{N,Int},Nidx}) where {Tdst<:Real,Tsrc<:Real,Ndst,Nsrc,N,Nidx}
+    dims = _check_dims(Ndst, Nsrc, N, Nidx)
+    @boundscheck _check_output(idx, dst, src, dims)
+    @boundscheck _check_input(idx, src)
+    T = promote_type(Tdst, Tsrc)
+    scatter!(op, T.(dst), src, idx, Val(dims))
```
CarloLucibello, line 75:
```suggestion
    scatter!(op, dst, src, idx, Val(dims))
```
scatter! is supposed to modify the `dst`, while this was creating a copy of `dst`. We don't need any promotion here.

src/scatter.jl
```diff
@@ -0,0 +1,189 @@
...
+end
+
+function scatter!(op, dst::AbstractArray{Tdst}, src::AbstractArray{Tsrc}, idx::AbstractArray{<:IntOrTuple},
+                  dims::Val{N}) where {Tdst<:Real,Tsrc<:Real,N}
+    colons = Base.ntuple(_->Colon(), N)
+    @simd for k in CartesianIndices(idx)
+        dst_v = view(dst, colons..., idx[k]...)
+        src_v = view(src, colons..., k)
+        @inbounds dst_v .= (op).(dst_v, src_v)
```
CarloLucibello, line 84: let's remove all inbounds and simd annotations, otherwise will encounter friction in reviews

src/scatter.jl
```diff
@@ -0,0 +1,189 @@
...
+                  idx::AbstractArray{NTuple{N,Int},Nidx}) where {Tdst<:Real,Tsrc<:Real,Ndst,Nsrc,N,Nidx}
+    dims = _check_dims(Ndst, Nsrc, N, Nidx)
+    @boundscheck _check_output(idx, dst, src, dims)
+    @boundscheck _check_input(idx, src)
+    T = promote_type(Tdst, Tsrc)
+    scatter!(op, T.(dst), src, idx, Val(dims))
+end
+
+function scatter!(op, dst::AbstractArray{Tdst}, src::AbstractArray{Tsrc}, idx::AbstractArray{<:IntOrTuple},
+                  dims::Val{N}) where {Tdst<:Real,Tsrc<:Real,N}
```
CarloLucibello, line 79:
```suggestion
function scatter!(op, dst::AbstractArray{Tdst}, src::AbstractArray{Tsrc}, idx::AbstractArray{<:IntOrTuple},
                  dims::Val{N}) where {Tdst,Tsrc,N}
```
Also, probably we don't need an extra method with the additional `dims` argument. `dims` can be calculated inside. In #280 that didn't seem to have  a performance impact, but maybe you can benchmark it by yourself here.

src/scatter.jl
yuehhua: Done.

src/scatter.jl
DhairyaLGandhi: Good catch, either the promotion needs to handled earlier, or enforce the eltype at the API level.

yuehhua: Done.

src/scatter.jl
yuehhua: Well.......fine. Done.

src/scatter.jl
yuehhua: It may provide pre-compilation information for compiler to compile specialized function for `dims` and dispatch on it. Constant propagation helps a lot.

CarloLucibello: I think the compiler is smart enough to reason with `N = Ndst - Nidx`, and the compilation is already specialized on `dst` and `idx` which univocally determine `N`. Anyways, it's an implementation detail not exposed to the user,  mostly a matter of conciseness, it's fine if you keep it as it is

src/scatter.jl
```diff
@@ -0,0 +1,187 @@
...
+
+    dst[:, ..., idx[k]...] = (op).(dst[:, ..., idx[k]...], src[:, ..., k...])
+"""
+function scatter!(op,
+                  dst::AbstractArray{Tdst,Ndst},
+                  src::AbstractArray{Tsrc,Nsrc},
+                  idx::AbstractArray{<:Integer,Nidx}) where {Tdst,Tsrc,Ndst,Nsrc,Nidx}
+    dims = _check_dims(Ndst, Nsrc, 1, Nidx)
+    @boundscheck _check_output(idx, dst, src, dims)
+    @boundscheck _check_input(idx, src)
```
CarloLucibello, line 62:
```suggestion
```
can be removed since inbounds is no longer used

yuehhua: Sure.

src/scatter.jl
```diff
@@ -0,0 +1,187 @@
...
+end
+
+function scatter(op::typeof(mean),
+                 src::AbstractArray{T,Nsrc},
+                 idx::AbstractArray{<:IntOrTuple,Nidx}) where {T,Nsrc,Nidx}
+    FT = float(T)
+    dims = Nsrc - Nidx
+    dst = zeros(FT, size(src)[1:dims]..., maximum_dims(idx)...)
+    scatter!(op, dst, FT.(src), idx)
+end
```
CarloLucibello, line 187: Instead of all these methods, one can use just this single one (or somthing similar)
```julia
function scatter(op, 
                src::AbstractArray{Tsrc, Nsrc}, 
                idx::AbstractArray{Tidx, Nidx}) where 
                    {Tsrc, Nsrc, Nidx, Tidx<:IntOrIntTuple}

    M = typelength(Tidx) 
    dstsize = (size(src)[1:Nsrc-M]..., size(idx)...)
    dst = similar(src, Tsrc, dstsize)
    fill!(dst, Base.reduce_empty(op, Tsrc))
    return scatter!(op, dst, src, idx)
end

typelength(::Type{<:Number}) = 1
typelength(::Type{<:NTuple{M}}) where M = M
```

CarloLucibello: actually not a good idea, `Base.reduce_empty` doesn't support min/max. 
But please use `similar` and `fill!` instead of `zeros` and `ones` so that we support generic arrays.

src/scatter.jl
```diff
@@ -0,0 +1,187 @@
...
+    scatter!(op, dst, src, idx)
+end
+
+function scatter(op::typeof(mean),
+                 src::AbstractArray{T,Nsrc},
+                 idx::AbstractArray{<:IntOrTuple,Nidx}) where {T,Nsrc,Nidx}
+    FT = float(T)
+    dims = Nsrc - Nidx
+    dst = zeros(FT, size(src)[1:dims]..., maximum_dims(idx)...)
+    scatter!(op, dst, FT.(src), idx)
```
CarloLucibello, line 186:
```suggestion
    scatter!(op, dst, src, idx)
```
no need to cast src

src/scatter.jl
yuehhua: Oh! Doesn't `zeros` and `ones` support generic arrays?

DhairyaLGandhi: It does afaik

It needs the identities for the eltypes defined.

CarloLucibello: no they don't, they will always return `Array` type, to support generic arrays (e.g. StaticArray, CuArray,...) the correct thing to do is `similar` and then `fill!`

src/scatter.jl
yuehhua: Done.

src/scatter.jl
```diff
@@ -0,0 +1,187 @@
...
+    scatter!(op, dst, src, idx, Val(dims))
+end
+
+function scatter!(op,
+                  dst::AbstractArray{Tdst,Ndst},
+                  src::AbstractArray{Tsrc,Nsrc},
+                  idx::AbstractArray{NTuple{N,Int},Nidx}) where {Tdst,Tsrc,Ndst,Nsrc,N,Nidx}
+    dims = _check_dims(Ndst, Nsrc, N, Nidx)
+    _check_output(idx, dst, src, dims)
+    _check_input(idx, src)
```
CarloLucibello, line 72:
```suggestion
```
bound checking is automatically performed in the loop, we don't need these functions anymore

src/scatter.jl
DhairyaLGandhi:
```julia
julia> struct M{T,S}
         a::T
         b::S
       end

julia> mr = reshape([M(rand(), rand()) for _ = 1:9], 3,3)
3×3 Array{M{Float64,Float64},2}:
 M{Float64,Float64}(0.33762, 0.0495362)  …  M{Float64,Float64}(0.260097, 0.922337)
 M{Float64,Float64}(0.282951, 0.18587)      M{Float64,Float64}(0.487388, 0.81491)
 M{Float64,Float64}(0.808664, 0.79451)      M{Float64,Float64}(0.0642862, 0.964739)

julia> Base.zero(::Type{<:M{T,S}}) where {T,S} = M(zero(T), zero(S))

julia> zero(mr)
3×3 Array{M{Float64,Float64},2}:
 M{Float64,Float64}(0.0, 0.0)  …  M{Float64,Float64}(0.0, 0.0)
 M{Float64,Float64}(0.0, 0.0)     M{Float64,Float64}(0.0, 0.0)
 M{Float64,Float64}(0.0, 0.0)     M{Float64,Float64}(0.0, 0.0)

julia> @code_warntype zero(mr)
Variables
  #self#::Core.Compiler.Const(zero, false)
  x::Array{M{Float64,Float64},2}

Body::Array{M{Float64,Float64},2}
1 ─ %1 = Base.similar(x)::Array{M{Float64,Float64},2}
│   %2 = Base.zero($(Expr(:static_parameter, 1)))::Core.Compiler.Const(M{Float64,Float64}(0.0, 0.0), false)
│   %3 = Base.fill!(%1, %2)::Array{M{Float64,Float64},2}
└──      return %3

```

Seems like it works as expected

yuehhua: For `maximum_dims`, I want the dimensions of the value in `idx`, not the dimensions of `idx` itself. So, `size(idx)` may not work for my case.

src/scatter.jl
DhairyaLGandhi: Why check bounds manually anyway!

src/scatter.jl
DhairyaLGandhi: With the automatic lowering for the different types of arrays

yuehhua: Done.

src/scatter.jl
yuehhua: Done.

src/scatter.jl
yuehhua: I commit a version for optimizing the process of generating colons.

```julia
julia> @code_typed ntuple(_->Colon(), Val(1))
CodeInfo(
1 ─     return (Colon(),)
) => Tuple{Colon}

julia> @code_typed ntuple(_->Colon(), Val(2))
CodeInfo(
1 ─     return (Colon(), Colon())
) => Tuple{Colon,Colon}

julia> @code_typed ntuple(_->Colon(), 3)
CodeInfo(
1 ── %1  = (n === 0)::Bool
└───       goto #3 if not %1
2 ──       goto #34
3 ── %4  = (n === 1)::Bool
└───       goto #5 if not %4
4 ──       goto #33
5 ── %7  = (n === 2)::Bool
└───       goto #7 if not %7
6 ──       goto #32
7 ── %10 = (n === 3)::Bool
└───       goto #9 if not %10
8 ──       goto #31
9 ── %13 = (n === 4)::Bool
└───       goto #11 if not %13
10 ─       goto #30
11 ─ %16 = (n === 5)::Bool
└───       goto #13 if not %16
12 ─       goto #29
13 ─ %19 = (n === 6)::Bool
└───       goto #15 if not %19
14 ─       goto #28
15 ─ %22 = (n === 7)::Bool
└───       goto #17 if not %22
16 ─       goto #27
17 ─ %25 = (n === 8)::Bool
└───       goto #19 if not %25
18 ─       goto #26
19 ─ %28 = (n === 9)::Bool
└───       goto #21 if not %28
20 ─       goto #25
21 ─ %31 = (n === 10)::Bool
└───       goto #23 if not %31
22 ─       goto #24
23 ─ %34 = invoke Base._ntuple(_2::Function, _3::Int64)::Tuple{Vararg{Colon,N} where N}
24 ┄ %35 = φ (#22 => (Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon()), #23 => %34)::Tuple{Vararg{Colon,N} where N}
25 ┄ %36 = φ (#20 => (Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon()), #24 => %35)::Tuple{Vararg{Colon,N} where N}
26 ┄ %37 = φ (#18 => (Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon()), #25 => %36)::Tuple{Vararg{Colon,N} where N}
27 ┄ %38 = φ (#16 => (Colon(), Colon(), Colon(), Colon(), Colon(), Colon(), Colon()), #26 => %37)::Tuple{Vararg{Colon,N} where N}
28 ┄ %39 = φ (#14 => (Colon(), Colon(), Colon(), Colon(), Colon(), Colon()), #27 => %38)::Tuple{Vararg{Colon,N} where N}
29 ┄ %40 = φ (#12 => (Colon(), Colon(), Colon(), Colon(), Colon()), #28 => %39)::Tuple{Vararg{Colon,N} where N}
30 ┄ %41 = φ (#10 => (Colon(), Colon(), Colon(), Colon()), #29 => %40)::Tuple{Vararg{Colon,N} where N}
31 ┄ %42 = φ (#8 => (Colon(), Colon(), Colon()), #30 => %41)::Tuple{Vararg{Colon,N} where N}
32 ┄ %43 = φ (#6 => (Colon(), Colon()), #31 => %42)::Tuple{Vararg{Colon,N} where N}
33 ┄ %44 = φ (#4 => (Colon(),), #32 => %43)::Tuple{Vararg{Colon,N} where N}
34 ┄ %45 = φ (#2 => (), #33 => %44)::Tuple{Vararg{Colon,N} where N}
└───       return %45
) => Tuple{Vararg{Colon,N} where N}
```

`ntuple(_->Colon(), Val(N))` returns directly, while `ntuple(_->Colon(), N)` encounter several branches.

src/scatter.jl
```diff
@@ -0,0 +1,165 @@
...
+typelength(::Type{<:NTuple{M}}) where M = M
+
+"""
+    scatter!(op, dst, src, idx)
+
+Scatter operation, which scatters data in `src` and assigns to `dst` according to `idx`.
+With the data going to the same place, specified operation is applied on to reduce data.
+For each index `k` in `idx`, accumulate values in `dst` according to
+
+    dst[idx[k]...] = (op).(dst[idx[k]...], src[k...])
```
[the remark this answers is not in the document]
yuehhua: Done.

src/scatter.jl
```diff
@@ -0,0 +1,165 @@
...
+
+# Arguments
+- `op`: operations to be applied on `dst` and `src`, e.g. `+`, `-`, `*`, `/`, `max` and `min`.
+- `dst`: the destination for `src` to aggregate to. This argument will be mutated.
+- `src`: the source data for aggregating.
+- `idx`: the mapping for aggregation from source (index) to destination (value).
+The index of `idx` is corresponding to the index of `src` and the value of `idx` is
+corresponding to the index of `dst`. The value of `idx` can be `Int` or `Tuple` type.
+
+    dst[:, ..., idx[k]...] = (op).(dst[:, ..., idx[k]...], src[:, ..., k...])
```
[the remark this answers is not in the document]
yuehhua: Done.

src/scatter.jl
```diff
@@ -0,0 +1,165 @@
...
+    dst[idx[k]...] = dst[idx[k]...] + mean.(src[k...])
+
+# Arguments
+- `dst`: the destination for `src` to aggregate to. This argument will be mutated.
+- `src`: the source data for aggregating.
+- `idx`: the mapping for aggregation from source (index) to destination (value).
+The index of `idx` is corresponding to the index of `src` and the value of `idx` is
+corresponding to the index of `dst`. The value of `idx` can be `Int` or `Tuple` type.
+
+    dst[:, ..., idx[k]...] = (op).(dst[:, ..., idx[k]...], src[:, ..., k...])
```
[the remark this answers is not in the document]
yuehhua: Done.

src/utils.jl
```diff
@@ -0,0 +1,42 @@
...
+    end
+    counts
+end
+
+function divide_by_counts!(xs, idx::AbstractArray, N)
+    counts = count_indices(idx, N)
+    @inbounds for ind = CartesianIndices(counts)
+        view(xs, :, ind) ./= counts[ind]
+    end
+end
```
CarloLucibello, line 42:
```suggestion
```

not used in this PR

src/scatter.jl
```diff
@@ -0,0 +1,149 @@
...
+    scatter!(op, dst, src, idx)
+end
+
+function scatter(op::typeof(min),
+                 src::AbstractArray{T,Nsrc},
+                 idx::AbstractArray{<:IntOrTuple,Nidx}) where {T,Nsrc,Nidx}
+    dims = Nsrc - Nidx
+    dstsize = (size(src)[1:dims]..., maximum_dims(idx)...)
+    dst = similar(src, T, dstsize)
+    fill!(dst, typemax(T))
```
CarloLucibello, line 136: I wonder if  `typemax` is the right thing to do here. The problem is if there are positions in `dst` which receive no contributions for `src` they will end up holding `typemax`, which doesn't seem meaningful. Maybe we should error out in such cases, but doing this check may have a performance impact

src/utils.jl

src/scatter.jl
yuehhua: Yeah, I thought this issue before. Checking the position of `dst` is properly covered by `idx` is the way to avoid holding `typemax`. But still, it is necessary to check values in `src` is smaller than the value we assigned, either `typemax` or `similar`. `similar` gives the value existing in bare memory, so we have no idea knowing if the values are smaller enough.

yuehhua: What if we give the `maximum` of `src`? Thus, the value is at least smaller or equals to the `maximum` of `src`.

mcabbott: I think `maximum(src)` would be more surprising, in un-visited entries. `typemax` seems OK to me.

yuehhua:
If it is ready to go, just let it go.

CarloLucibello:
ok, I'll merge this so that work can proceed, I'll open a pr to revisit the docstrings

PR #257, opened 2020-12-28
remove ZygoteRules

CarloLucibello:
With this change, Flux's tests pass with both Zygote v0.5 and v0.6,
redundancy or ChainRule's rrules here with the ones defined in Zygote v0.5 doesn't seem to be a problem. 
ZygoteRules' adjoints instead were giving redefinition warnings.

With this change we can tag a v0.7.10 release instead of v0.8, which would relieve us of a lot of pain:
CUDA depends on NNlib, and Flux depends on CUDA,
but CUDA master is only compatible with julia 1.6, so it would take a while to see CUDA release with the NNlib's compat bound update. Moreover, we would have to drop julia < 1.6 support in Flux. 

@simeonschaub do you think this change is fine?


src/zygoterules.jl
```diff
@@ -1,17 +1,35 @@
...
+# # This is a performance hack specifically for Zygote, because it doesn't handle fused
+# # broadcasts well
+# for (f, df) in [
+#     (:relu, :(x .> 0)),
+#     (:selu, :(dselu.(x))),
+#     (:elu, :(delu.(x))),
+#     (:σ, :(conj.(Ω .* (1 .- Ω)))),
+# ]
+#     pullback = Symbol(:broadcasted_, f, :_pullback)
+#     @eval @adjoint function Base.Broadcast.broadcasted(::typeof($f), x::Numeric)
```
simeonschaub, line 10: I am not sure this is a good idea, since IIUC, loading NNlib will now always error if Zygote is not loaded first. Perhaps we can use Requires.jl for this instead?

simeonschaub (changes requested):

src/zygoterules.jl
CarloLucibello: I don't understand, when should it error? I can do `using NNlib`  just fine

CarloLucibello: Using requires is not an option I think, because Zygote v0.5 precompilation triggers NNlib's one, NNlib define the rules, then Zygote tries to define the same rules and gives the warning

simeonschaub: > I don't understand, when should it error? I can do `using NNlib`  just fine

I can't try it ATM, but since it doesn't load `ZygoteRules` now, I don't know where it should get the definition of `@adjoint` from. It might just work for the tests, since those already load Zygote. Did you do `using NNlib` before loading any other packages?

simeonschaub: > Using requires is not an option I think, because Zygote v0.5 precompilation triggers NNlib's one, NNlib define the rules, then Zygote tries to define the same rules and gives the warning

I think we could use requires for checking if Zygote is loaded and then only define those rules if it's the appropriate version.

simeonschaub (commented):
_

src/zygoterules.jl
CarloLucibello: > I don't know where it should get the definition of @adjoint from

ah, maybe you didn't see the new rules down below?

> I think we could use requires for checking if Zygote is loaded and then only define those rules if it's the appropriate version.

I think this won't work because Zygote v0.5 depends on NNlib, so NNlib is always loaded first

CarloLucibello: > I don't know where it should get the definition of `@adjoint` from

maybe there is been a misunderstanding, I replaced ZygoteRules' adjoint with ChainRules' rrules. I don't know why it wasn't done before, maybe because is not something any AD engine would want to do

simeonschaub: 🤦I didn't see this was commented out for some reason, sorry! If this is just a temporary workaround, using ChainRulesCore for these seems ok to me, but we really want to avoid this once people are actually using other ADs than Zygote. https://github.com/JuliaDiff/ChainRulesCore.jl/issues/270 might be helpful here.

simeonschaub:
@oxinabox Will this cause problems for Nabla.jl?

oxinabox:
No, this should be great with Nabla

src/activations.jl
```diff
@@ -222,7 +222,28 @@ See [Softshrink Activation Function](https://www.gabormelli.com/RKB/Softshrink_A
...
-for f in (:σ, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
+for f in (:σ, :hardσ, :logσ, :hardtanh, :relu, :leakyrelu, 
+        :relu6, :rrelu, :elu, :gelu, :swish, :lisht, :selu, 
+        :celu, :trelu, :softsign, :softplus, :logcosh, :mish, :tanhshrink, :softshrink)
     @eval $(f)(x::AbstractArray, args...) =
       error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
 end
+
+# # This is a performance hack specifically for Zygote, because it doesn't handle fused
+# # broadcasts well
```
simeonschaub, line 233:
```suggestion
# # broadcasts well
# TODO: don't define these for ADs other than Zygote
```

simeonschaub (approved):

PR #258, opened 2020-12-28
Remove TODO about not defining broadcasted for ADs other than Zygote

oxinabox:
Follow up from #257 I think we should do this for all ADs,
and so the TODO shouldn't be there.

AFAICT, this code is basically just hand-unrolling what the AD would ideally do on its own.
So at worst it has basically no effect except maybe to reduce compile-time slightly.

We are talking about doing something like this for a whole bunch of scalar functions in ChainRules.
https://github.com/JuliaDiff/ChainRules.jl/issues/222

@willtebbutt what do you think?


willtebbutt:
The only scenario in which I can imagine this being a bad idea is if the AD in question wants to fuse the reverse-pass. Otherwise, I agree that it's pretty harmless.

CarloLucibello (approved):

PR #259, opened 2020-12-30
more extensive tests and some reorg

CarloLucibello:
- move AD rules close to function definitions
- more robust and extensive AD tests for the activations
- extend more systematically AD rules with their broadcasted counterpart (also see https://github.com/JuliaDiff/ChainRulesCore.jl/issues/271)


src/gemm.jl
```diff
@@ -95,31 +93,27 @@ for (gemm, elt) in gemm_datatype_mappings
...
-                              transA, transB, m, n,
-                              ka, alpha, ptrAk, max(1,Base.stride(A,2)),
-                              ptrBk, max(1,Base.stride(B,2)), beta, ptrCk,
-                              max(1,Base.stride(C,2)))
-                    end
-                end
-                set_num_threads(old_threads)
-
-            else # small problem, no threads
+            # TODO fix multithreading problems
```
mcabbott, line 96: What's the reason for this?

CarloLucibello: it was producing wrong results in the multithreaded CI environment. Didn't investigated the causes

mcabbott: Got a link? I can't reproduce locally.

CarloLucibello: can't find it. You could comment in again this branch and file a PR to trigger CI

mcabbott: The commit "fix compat" comments these out. But at that point, `using Compat` has already been deleted, so it will fail, not with the wrong results, but because the function from Compat.jl is not being imported. 

That was deleted in "fix tests and cleanup", the only change to this file in that commit.

PR #261, opened 2020-12-30
fix hardsigmoid and use float(x) instead of x/1 

CarloLucibello:
~~We do a lot of wasteful divisions in the effort of being type stable with integer inputs, something which we don't really care about. 
With this PR we still support integer inputs but we focus on Float32 and Float64 performances.
I remove all those `x/1` divisions and  `oftype(x/1, someconstant)` conversions.
As a result, the code is also much nicer. 
This PR is breaking for integer inputs and also disrespectful of Float16, since all constants are now Float32. 
We should benchmark  before merging (if we want to merge)
This PR makes some of the function type-unstable with integer inputs~~
 
- Fix the hardsigmoid definition, not matching the cited paper, so that now it matches the pytorch definition.
- Replace the occurrencies of `x / 1`  with `float(x)` for clarity

### DIFF
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -17,7 +17,8 @@ end
 # Aliases
 export sigmoid, hardsigmoid, logsigmoid, thresholdrelu
 
-
+# of type float
+oftf(x, y) = oftype(float(x), y)
 
 """
     σ(x) = 1 / (1 + exp(-x))
@@ -33,12 +34,13 @@ end
 const sigmoid = σ
 
 """
-    hardσ(x) = max(0, min(1, (x + 1)/2)
+    hardσ(x) = max(0, min(1, (x + 3) / 6)
 
-Segment-wise linear approximation of sigmoid.
-See [BinaryConnect: Training Deep Neural Networks withbinary weights during propagations](https://arxiv.org/abs/1511.00363).
+Piecewise linear approximation of sigmoid.
 """
-hardσ(x) = max(zero(x), min(one(x), (x + 1) / 2))
+hardσ(x) = max(0, min(1, (x + 3) / 6))
+
+# https://pytorch.org/docs/stable/generated/torch.nn.Hardsigmoid.html
 
 const hardsigmoid = hardσ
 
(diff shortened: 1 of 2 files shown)


ToucheSir:
Do you have any benchmarks handy showing how much of a performance improvement this offers? Also I'm generally sympathetic to simplification, but getting rid of Float16 support does not feel like a good compromise.

mcabbott:
It was my impression that these divisions compiled away, although surely calling `float(x)` etc would be less myserious:
```
julia> f(x,y) = oftype(x/1, y)
f (generic function with 1 method)

julia> @code_typed f(1, 2)
CodeInfo(
1 ─ %1 = Base.sitofp(Float64, y)::Float64
└──      return %1
) => Float64

julia> @code_typed Float64(2)
CodeInfo(
1 ─ %1 = Base.sitofp(Float64, x)::Float64
└──      return %1
) => Float64
```

Quite a lot of the other things changed here do seem like unnecessary clutter, e.g. because `min` promotes its arguments anyway.

src/activations.jl
```diff
@@ -33,13 +33,13 @@ end
...
 """
-    hardσ(x, a=0.2) = max(0, min(1, a * x + 0.5))
+    hardσ(x) = max(0, min(1, (x + 1)/2)
 
 Segment-wise linear approximation of sigmoid.
 See [BinaryConnect: Training Deep Neural Networks withbinary weights during propagations](https://arxiv.org/abs/1511.00363).
 """
-hardσ(x, a=0.2) = oftype(x/1, max(zero(x/1), min(one(x/1), oftype(x/1,a) * x + oftype(x/1,0.5))))
-    
+hardσ(x) = max(zero(x), min(one(x), (x + 1) / 2))
```
CarloLucibello, line 41: the default value of `a` was in disagreement with the one in the paper. Also, the paper doesn't mention any tunable parameter `a`

AriMKatz:
+1 for keeping Float16 support, if possible

CarloLucibello:
According to @mcabbott 's suggestion, I removed some `one(x)` and `zero(x)` clutter in `min`s and `max`s.
I'm also casting constants to the appropriate types to avoid Float16 promotions as requested.
As @mcabbott showed, the changes here are purely aesthetic and have no performance impact, but they do make the code much more clean. 
The only thing we lose on some functions is type stability with integer inputs, wich doesn't seem important

CarloLucibello:
I changed my mind, It could be the case that in some weird applications integer inputs are used, so better be fast also in that case sacrificing a readability a little

PR #265, opened 2021-01-05
Add SineReLU to the activation list

djthegr8:
See [here](https://medium.com/@wilder.rodrigues/sinerelu-an-alternative-to-the-relu-activation-function-e46a6199997d) for the reference.

### DIFF
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -97,7 +97,15 @@ function rrelu(x, l = 1 / 8.0, u = 1 / 3.0)
     a = oftype(x / 1, (u - l) * rand() + l)
     return leakyrelu(x, a)
 end
-
+"""
+   sinerelu(x, epsilon = 0.0025) = x > 0 ? x : epsilon * (sin(x) - cos(x))
+   
+Sine Rectified Linear Unit activation function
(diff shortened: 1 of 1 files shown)


CarloLucibello:
we don't add a new activation just because someone made medium article about it

djthegr8:
It isn't plainly a medium article. Anyways, it is included in the
keras.contrib functions...

ToucheSir:
Though I don't believe there's a formal threshold for what should and shouldn't belong in NNlib, I think it's fair to expect the following from any new activation function:
1. Some kind of publication (a pre-print is fine) with background motivation and reasonably comprehensive testing.
2. Enough name recognition + common usage in publications, libraries or frameworks beyond the original author's work.

In this case, SineReLU was added to keras-contrib via a PR by _the same author_ of the medium article. It also does not appear in any other libraries or model implementations with a sizeable following (i.e. GitHub stars). I couldn't find a formal paper describing it either, so unless the author forgot to link one in their medium post the grounds for this are even shakier.

Lastly, I'm not sure what keras-contrib's grounds for contributions are, but it seems like they're very much a "kitchen sink" model where all functionality is _caveat emptor_ and may not ever make it upstream. On those terms, Flux/NNlib are probably closer to TF/Keras proper, PyTorch or JAX in that we require a somewhat higher burden of proof before adding new features.

djthegr8:
Cool, thanks!

PR #266, opened 2021-01-11
improve bilinear upsampling

maxfreu:
Sorry to bother you again! But this stuff didn't let me sleep, so here is a CPU implementation. See [GPU PR](https://github.com/JuliaGPU/CUDA.jl/pull/636) here.

- [x] parallelized
- [x] faster than current implementation
- [x] open-cv and pytorch compliant
- [x] tests pass <- breaking tests seem unrelated (?)

I reviewed the tests of the current implementation and found it a bit strange, actually. So this one comes with a bit different tests.

Mean benchmark times in ms, upsampling by a factor of 2,
tested on 12 threads @3.7Ghz, julia 1.7

32x32x1024x1
|   | before  | after  |
|---|---|---|
| forward  | 53  | 2.7  |
| backward  | 94  |  0.8 |

196x196x128x1
|   | before  | after  |
|---|---|---|
| forward  | 318  | 21  |
| backward  | 64  |  3.6 |

single threaded:

32x32x1024x1
|   | before  | after  |
|---|---|---|
| forward  | 50  | 27  |
| backward  | 71  |  7.5 |

196x196x128x1
|   | before  | after  |
|---|---|---|
| forward  | 300  | 150  |
| backward  | 56  |  35 |

Single core would rock with cwhn tensor layout, but parallelized they are more or less the same.

Kind regards! :)

### DIFF
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -29,9 +29,6 @@ The interpolation grid is identical to the one used by open-cv.
 
 Currently only 2d upsampling is supported.
 """
-function upsample_bilinear end
-
-# this is the user-facing part
 function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
     w,h,c,n = size(x)
     if outsize===nothing
@@ -150,8 +147,6 @@ end
 # Outputs
 - `dx`: downsampled version of `Δ`
 """
-function ∇upsample_bilinear end
-
 function ∇upsample_bilinear(Δ::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
     w,h,c,n = size(Δ)
     if outsize===nothing
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -171,22 +171,13 @@ end
 # Outputs
 - `dx`: downsampled version of `Δ`
 """
-function ∇upsample_bilinear(Δ::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
-    w,h,c,n = Base.size(Δ)
-    if scale != (1,1) && size !== nothing
-        error("Please provide either scale or size, not both. Got scale=$scale and size=$size.")
-    end
-    if size===nothing
-        out_w = ceil(Int, w/scale[1])
-        out_h = ceil(Int, h/scale[2])
-    else
-        out_w, out_h = size
+function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
+    out_w, out_h = size
     end
     dx = zeros(T, out_w, out_h, c, n)
     return ∇upsample_bilinear_whcn!(Δ, dx)
 end
 
-∇upsample_bilinear(Δ, scale::Real; size=nothing) = ∇upsample_bilinear(Δ, (scale, scale); size=size)
 
 function ∇upsample_bilinear_whcn!(Δ::AbstractArray{T,4}, grad_input::AbstractArray{T,4}) where T
     size(grad_input)[3:4] == size(Δ)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -222,7 +222,7 @@ end
 function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, scale; size=nothing)
     Ω = upsample_bilinear(x, scale; size=size)
     function upsample_bilinear_pullback(Δ)
-        (NO_FIELDS, ∇upsample_bilinear(Δ, scale; size=(Base.size(x,1),Base.size(x,2))), DoesNotExist())
+        (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))), DoesNotExist())
     end
     return Ω, upsample_bilinear_pullback
 end
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -20,13 +20,13 @@ end
     upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing)
 
 Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `scale`,
-using bilinear interpolation. The interpolation is identical to the one used by pytorch.
+using bilinear interpolation.
 
 The size of the output is equal to
 `(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
 
 Examples:
-```
+```julia
 upsample_bilinear(x, (2, pi)) # real scaling factors are allowed
 upsample_bilinear(x; outsize=(64,64)) # note the semicolon, outsize is a keyword argument
 ```
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -20,13 +20,16 @@ end
     upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing)
 
 Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `scale`,
-using bilinear interpolation.
+using bilinear interpolation. The interpolation is identical to the one used by pytorch.
 
 The size of the output is equal to
 `(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
 
-The interpolation grid is identical to the one used by open-cv.
-
+Examples:
+```
+upsample_bilinear(x, (2, pi)) # real scaling factors are allowed
+upsample_bilinear(x; outsize=(64,64)) # note the semicolon, outsize is a keyword argument
+```
 Currently only 2d upsampling is supported.
 """
 function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -107,8 +107,8 @@ function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); s
     else
         out_w, out_h = size
     end
-    y = Array{T,4}(undef, out_w, out_h, c, n)
-    return upsample_bilinear_whcn_kernel!(y, x)
+    y = similar(x, T, out_w, out_h, c, n)
+    return upsample_bilinear!(y, x)
 end
 
 upsample_bilinear(x, scale::Real; size=nothing) = upsample_bilinear(x, (scale,scale); size=size)
@@ -118,6 +118,8 @@ function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); s
     res = upsample_bilinear(y, scale; size=size)
     return round.(T, res)
 end
+
+upsample_bilinear!(y::AbstractArray{<:Any,4}, x::AbstractArray{<:Any,4}) = upsample_bilinear_whcn_kernel!(y,x)
 
 # this is the core function which works on arrays of arbitrary size
 # the implementation is a translation of https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/UpSampleMoreKernel.cpp
@@ -175,15 +177,17 @@ end
 function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
     _, _, c, n = Base.size(Δ)
     out_w, out_h = size
-    dx = zeros(T, out_w, out_h, c, n)
-    return ∇upsample_bilinear_whcn_kernel!(dx, Δ)
+    dx = zero(similar(Δ, T, out_w, out_h, c, n))
+    return ∇upsample_bilinear!(dx, Δ)
 end
+
+∇upsample_bilinear!(dx::AbstractArray{<:Any,4}, Δ::AbstractArray{<:Any,4}) = ∇upsample_bilinear_whcn_kernel!(dx, Δ)
 
 function ∇upsample_bilinear_whcn_kernel!(dx::AbstractArray{T,4}, Δ::AbstractArray{T,4}) where T
     if size(dx) == size(Δ)
         return Δ
     end
-    
+
     size(dx)[3:4] == size(Δ)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
     in_w, in_h, channels, batches = size(dx)
 
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -219,7 +219,7 @@ function ∇upsample_bilinear_whcn!(Δ::AbstractArray{T,4}, grad_input::Abstract
     return grad_input
 end
 
-function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, scale; size=nothing)
+function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, scale=(1,1); size=nothing)
     Ω = upsample_bilinear(x, scale; size=size)
     function upsample_bilinear_pullback(Δ)
         (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))), DoesNotExist())
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -213,7 +213,7 @@ end
 function ChainRulesCore.rrule(::typeof(upsample_bilinear), x; size)
     Ω = upsample_bilinear(x; size=size)
     function upsample_bilinear_pullback(Δ)
-        (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))), DoesNotExist())
+        (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))))
     end
     return Ω, upsample_bilinear_pullback
 end
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -34,7 +34,10 @@ Currently only 2d upsampling is supported.
 """
 function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
     w,h,c,n = Base.Base.size(x)
-    if size===nothing
+    if scale != (1,1) && size !== nothing
+        error("Please provide either scale or size, not both. Got scale=$scale and size=$size.")
+    end
+    if size === nothing
         out_w = floor(Int, scale[1]*w)
         out_h = floor(Int, scale[2]*h)
     else
@@ -100,6 +103,9 @@ end
 """
 function ∇upsample_bilinear(Δ::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
     w,h,c,n = Base.size(Δ)
+    if scale != (1,1) && size !== nothing
+        error("Please provide either scale or size, not both. Got scale=$scale and size=$size.")
+    end
     if size===nothing
         out_w = ceil(Int, w/scale[1])
         out_h = ceil(Int, h/scale[2])
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -107,8 +107,11 @@ function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); s
     else
         out_w, out_h = size
     end
+    if (w,h) == (out_w, out_h)
+        return x
+    end
     y = similar(x, T, out_w, out_h, c, n)
-    return upsample_bilinear!(y, x)
+    return upsample_bilinear_whcn!(y, x)
 end
 
 upsample_bilinear(x, scale::Real; size=nothing) = upsample_bilinear(x, (scale,scale); size=size)
@@ -119,8 +122,6 @@ function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); s
     return round.(T, res)
 end
 
-upsample_bilinear!(y::AbstractArray{<:Any,4}, x::AbstractArray{<:Any,4}) = upsample_bilinear_whcn_kernel!(y,x)
-
 # this is the core function which works on arrays of arbitrary size
 # the implementation is a translation of https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/UpSampleMoreKernel.cpp
 # which implements open-cv style linear interpolation / upsampling
@@ -130,10 +131,7 @@ upsample_bilinear!(y::AbstractArray{<:Any,4}, x::AbstractArray{<:Any,4}) = upsam
 # - RGB types could be supported via reinterpreting
 # - integer types need to be converted to Float and back
 # - rationals work, but are slow
-function upsample_bilinear_whcn_kernel!(output::AbstractArray{T,4}, input::AbstractArray{T,4}) where T
-    if size(input) == size(output)
-        return input
-    end
+function upsample_bilinear_whcn!(output::AbstractArray{T,4}, input::AbstractArray{T,4}) where T
     size(input)[3:4] == size(output)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
     in_w, in_h, channels, batches = size(input)
     # treat batch and channel dimension as one for better parallelization granularity
@@ -175,19 +173,16 @@ end
 - `dx`: Downsampled version of `Δ`
 """
 function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
-    _, _, c, n = Base.size(Δ)
+    w, h, c, n = Base.size(Δ)
     out_w, out_h = size
-    dx = zero(similar(Δ, T, out_w, out_h, c, n))
-    return ∇upsample_bilinear!(dx, Δ)
-end
-
-∇upsample_bilinear!(dx::AbstractArray{<:Any,4}, Δ::AbstractArray{<:Any,4}) = ∇upsample_bilinear_whcn_kernel!(dx, Δ)
-
-function ∇upsample_bilinear_whcn_kernel!(dx::AbstractArray{T,4}, Δ::AbstractArray{T,4}) where T
-    if size(dx) == size(Δ)
+    if (w,h) == (out_w, out_h)
         return Δ
     end
+    dx = zero(similar(Δ, T, out_w, out_h, c, n))
+    return ∇upsample_bilinear_whcn!(dx, Δ)
+end
 
+function ∇upsample_bilinear_whcn!(dx::AbstractArray{T,4}, Δ::AbstractArray{T,4}) where T
     size(dx)[3:4] == size(Δ)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
     in_w, in_h, channels, batches = size(dx)
 
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -44,6 +44,8 @@ function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); o
     return upsample_bilinear_whcn!(y, x)
 end
 
+upsample_bilinear(x, scale::Real; outsize=nothing) = upsample_bilinear(x, (scale,scale); outsize=outsize)
+
 # this is the core function which works on arrays of arbitrary size
 # the implementation is a translation of https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/UpSampleMoreKernel.cpp
 # which implements open-cv style linear interpolation / upsampling
@@ -107,6 +109,8 @@ function ∇upsample_bilinear(Δ::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1
     dx = zeros(T, out_w, out_h, c, n)
     return ∇upsample_bilinear_whcn!(Δ, dx)
 end
+
+∇upsample_bilinear(Δ, scale::Real; outsize=nothing) = ∇upsample_bilinear(Δ, (scale, scale); outsize=outsize)
 
 function ∇upsample_bilinear_whcn!(Δ::AbstractArray{T,4}, grad_input::AbstractArray{T,4}) where T
     size(grad_input)[3:4] == size(Δ)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -1,234 +1,214 @@
 export upsample_bilinear, ∇upsample_bilinear, pixel_shuffle
 
+# utility function
+@inline function compute_source_index_and_lambda(
+    ratio, # 0 < ratio < 1
+    output_index,
+    input_size,
+    output_size
+)
+    real_input_index = ratio*output_index
+    input_index0 = floor(Int, real_input_index) # typecast to int was here in C++
+    offset = (input_index0 < input_size - 1) ? 1 : 0
+    input_index1 = input_index0 + offset
+    lambda1 = real_input_index - input_index0
+    lambda0 = 1 - lambda1
+    return input_index0, input_index1, lambda0, lambda1
+end
+
 """
-    upsample_bilinear(x::AbstractArray{<:Number,4}, k::NTuple{2,Int})
+    upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing)
 
-Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `k`,
-using bilinear interpolation. 
+Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `scale`,
+using bilinear interpolation.
 
-The size of the output is equal to 
-`(k[1]*S1, k[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
+The size of the output is equal to
+`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
 
-The interpolation grid is identical to the one used by `imresize` from `Images.jl`.
+The interpolation grid is identical to the one used by open-cv.
 
 Currently only 2d upsampling is supported.
 """
-function upsample_bilinear(x::AbstractArray{T,4}, k::NTuple{2,Int}) where T
-    # This function is gpu friendly
-    
-    imgsize = size(x)
-    newsize = get_newsize(imgsize, k)
+function upsample_bilinear end
 
-    # Get linear interpolation lower- and upper index, and weights
-    ilow1, ihigh1, wdiff1 = get_inds_and_ws(x, imgsize[1], newsize[1], 1)
-    ilow2, ihigh2, wdiff2 = get_inds_and_ws(x, imgsize[2], newsize[2], 2)
-
-    # Adjust the upper interpolation indices of the second dimension
-    ihigh2_r = adjoint_of_idx(ilow2)[ihigh2]
-
-    @inbounds y = @view(x[ilow1,ilow2,:,:]) .* (1 .- wdiff1) .+ @view(x[ihigh1,ilow2,:,:]) .* wdiff1
-    @inbounds y .= y .* (1 .- wdiff2) .+ y[:,ihigh2_r,:,:] .* wdiff2
-    # @inbounds y = y .* (1 .- wdiff2) .+ @view(y[:,ihigh2_r,:,:]) .* wdiff2 # equivalent to line above
-    return y
+# this is the user-facing part
+function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = size(x)
+    if outsize===nothing
+        out_w = floor(Int, scale[1]*w)
+        out_h = floor(Int, scale[2]*h)
+    else
+        out_w, out_h = outsize
+    end
+    y = Array{T,4}(undef, out_w, out_h, c, n)
+    return upsample_bilinear_whcn!(y, x)
 end
 
-function get_inds_and_ws(x::T, n::Int, m::Int, dim::Int) where T <: AbstractArray
-    # Creates interpolation grid for resampling. 
-    # Creates the same grid as used in Image.jl `imresize`.
-    step = n // m
-    offset = (n + 1)//2 - step//2 - step * (m//2 - 1)
-    xq = clamp.(range(offset, step=step, length=m), 1, n)
+# this is the core function which works on arrays of arbitrary size
+# the implementation is a translation of https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/UpSampleMoreKernel.cpp
+# which implements open-cv style linear interpolation / upsampling
+# for simplicity, corners are aligned and all logic for other behaviour has been stripped
+# - whcn because there is also a cwhn implementation
+# - the function is parallelized using @threads
+# - RGB types could be supported via reinterpreting
+# - integer types need to be converted to Float and back
+# - rationals work, but are slow
+function upsample_bilinear_whcn!(output::AbstractArray{T,4}, input::AbstractArray{T,4}) where T
+    if size(input) == size(output)
+        return input
+    end
+    size(input)[3:4] == size(output)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
+    in_w, in_h, channels, batches = size(input)
+    # treat batch and channel dimension as one for better parallelization granularity
+    channels *= batches
+    out_w, out_h, _, _ = size(output)
+    output_slice_size = out_h * out_w
 
-    # Creates interpolation lower and upper indices, and broadcastable weights
-    ilow = floor.(Int, xq)
-    ihigh = ceil.(Int, xq)
-    sizew = ntuple(i-> i == dim ? length(xq) : 1, ndims(x))
-    wdiff = convert(T, reshape(xq .- ilow, sizew)) # wdiff possibly lives on gpu
-    return ilow, ihigh, wdiff
+    # T() and // so that we can handle rationals (super slow)
+    width_scale  = T((in_w - 1) // (out_w - 1))
+    height_scale = T((in_h - 1) // (out_h - 1))
+
+    @inline idx(c, h, w) = c * in_h * in_w + h * in_w + w + 1
+
+    @inbounds Threads.@threads for c in 0:channels-1
+        for oh in 0:out_h-1
+            ih0, ih1, h0lambda, h1lambda = compute_source_index_and_lambda(height_scale, oh, in_h, out_h)
+            for ow in 0:out_w-1
+                iw0, iw1, w0lambda, w1lambda = compute_source_index_and_lambda(width_scale, ow, in_w, out_w)
+                output_offset = c * output_slice_size + oh * out_w + ow + 1
+                output[output_offset] =
+                    (h0lambda * w0lambda * input[idx(c, ih0, iw0)] + # h0 * w0 * i00
+                     h0lambda * w1lambda * input[idx(c, ih0, iw1)] + # h0 * w1 * i01
+                     h1lambda * w0lambda * input[idx(c, ih1, iw0)] + # h1 * w0 * i10
+                     h1lambda * w1lambda * input[idx(c, ih1, iw1)])  # h1 * w1 * i11
+            end
+        end
+    end
+    return output
+end
+
+# this is just for convenience, to support matrices
+
+# function upsample_bilinear_whcn!(output::AbstractMatrix, input::AbstractMatrix)
+#     y = reshape(output, (size(output)..., 1, 1))
+#     x = reshape(input,  (size(input)..., 1, 1))
+#     return dropdims(upsample_bilinear_whcn!(y, x); dims=(3,4))
+# end
+
+# function upsample_bilinear(x::AbstractArray{T,2}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+#     w,h = size(x)
+#     if outsize===nothing
+#         out_w = floor(Int, scale[1]*w)
+#         out_h = floor(Int, scale[2]*h)
+#     else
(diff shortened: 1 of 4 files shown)


CarloLucibello:
Well, that's an awesome improvement! 
Can you benchmark single-threaded performance against current master as well? 
Maybe we should also start adding benchmark scripts to some `perf/`  folder, even if they are not very polished.
Can you annotate with a comment tests where the outputs matches the pytorch one, so that we avoid deleting them in the future?

src/upsample.jl
```diff
@@ -20,13 +20,16 @@ end
     upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing)
 
 Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `scale`,
-using bilinear interpolation.
+using bilinear interpolation. The interpolation is identical to the one used by pytorch.
```
CarloLucibello, line 23: let's not mention other DL frameworks in the docs, doesn't look very professional

src/upsample.jl
```diff
@@ -1,234 +1,158 @@
...
-```julia
-x = [1, 2, 3, 4, 5]
-idx = [1, 2, 2, 3, 4, 4, 5]
-idx_adjoint = adjoint_of_idx(idx)
-@assert x[idx][idx_adjoint] == x
+The size of the output is equal to
+`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
+
+Examples:
+```
```
CarloLucibello, line 29:
```suggestion
```julia
```

maxfreu:
In case you didn't see: I posted the single core benchmark in the original post.

CarloLucibello:
ufff, maxpool tests are passing again, it's incredible how unstable they are. Can you change `@test_broken` to `@test` there?

maxfreu:
I couldn't find any occurences of `@test_broken` in my pooling.jl test file. Did you mean I should remove `broken=xy` in `gradtest`? I just got a glimpse of what you mean by unstable; sometimes the `AutoDiff: spatial_rank` test passes (locally), sometimes not.

CarloLucibello:
Yes, maybe change
```julia
gradtest(x -> maxpool(x, pdims), x; broken=spatial_rank <= 2)
```
to
```julia
gradtest(x -> maxpool(x, pdims), x; broken=spatial_rank <= 0)
```
since I have the impression we will have to change it back soon

test/pooling.jl
```diff
@@ -813,15 +813,15 @@ end
 @testset "AutoDiff: spatial_rank=$spatial_rank" for spatial_rank in (1, 2)
   x = rand(rng, repeat([10], spatial_rank)..., 3, 2)
   pdims = PoolDims(x, 2)
-  gradtest(x -> maxpool(x, pdims), x; broken=spatial_rank <= 2)
+  gradtest(x -> maxpool(x, pdims), x, broken=spatial_rank <= 0) # was <= 2 before
```
maxfreu, line 816: Is this what you meant? I think it's still messed up.

CarloLucibello: that's fine. This should be properly fixed somewhere else, but we need to have tests passing in this PR

CarloLucibello:
Can you add a test with non-integer scale?

It's a bit sad we lose CuArray support until https://github.com/JuliaGPU/CUDA.jl/pull/636 is merged, but I cannot think of a way around it. 

Now that you have thought a bit more about this, would you be able to extend the code to support the 1d and 3d case (in a later PR)? 

Following the discussion in https://github.com/FluxML/Flux.jl/pull/1468, can you add support for integer scale?
You are welcome to chime in the discussion on the interface btw

maxfreu:
Supporting 1D and 3D is easy; just one less or more for-loop. Nearest neighbour is also easy to implement this way, as only the source index calculation is changed.

CarloLucibello:
> Supporting 1D and 3D is easy; just one less or more for-loop.

a PR would be much appreciated

src/upsample.jl
```diff
@@ -1,234 +1,162 @@
...
-    return idx_adjoint
+function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = size(x)
+    if outsize===nothing
+        out_w = floor(Int, scale[1]*w)
+        out_h = floor(Int, scale[2]*h)
+    else
+        out_w, out_h = outsize
+    end
+    y = Array{T,4}(undef, out_w, out_h, c, n)
```
mcabbott, line 43: I have not checked with this implementation, but the previous one gave errors on an array of integers. I think something like this might be the simplest way to allow that:
```suggestion
    y = Array{float(T),4}(undef, out_w, out_h, c, n)
```
although I hope it won't break anything else?

mcabbott: Oh this won't allow rationals, I guess.

It also won't work at all unless `upsample_bilinear_whcn!(output::AbstractArray{T,4}, input::AbstractArray{T,4})` is relaxed not to use the same `T`.

maxfreu: Yes, the current implementation only works for reals and rationals. Integers can only be supported by converting forth and back I think, which could be supported by a short wrapper. Hmm and yes, maybe it makes sense to relax the type constraint. That way one would save one allocation for converting to float type and one could directly read the ints.

maxfreu: Hmmm, does it really make sense to support ints here? What should the gradients be? Also int?

mcabbott: One reason is to allow doc examples like `[1 2; 3 4]` to work, for which just `float(x)` would be fine. But perhaps it's worth handling better, aren't images stored as `Array{UInt8}` fairly common? 

I think gradients should not be locked to `Int`, I would have said `float(eltype(Δ))` but not if you want rationals to work.

The reason to support rationals is to allow some pretty tests?

src/upsample.jl
```diff
@@ -1,234 +1,162 @@
...
-The above holds as long as `idx` contains every index in `x`.
+Currently only 2d upsampling is supported.
 """
-function adjoint_of_idx(idx::Vector{Int})
-    d = trues(length(idx))
-    d[2:end] .= diff(idx)
-    idx_adjoint = findall(d)
-    return idx_adjoint
+function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); outsize::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = size(x)
```
mcabbott, line 36: May I put in a vote for `size=` as a keyword name? Along the lines of `range(1, 10, length=100)`, which likewise re-uses the name of a function. 
```suggestion
function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
    w,h,c,n = Base.size(x)
```

maxfreu: In principle no problem, but maybe we should we wait what https://github.com/FluxML/Flux.jl/pull/1468 results in.

mcabbott: Indeed, they should agree. And we should use `size=(10,10)` there too.

I guess the one reservation is that this is only the "spatial" size, not literally the whole `size(y)` of the output. But still seems OK.

CarloLucibello:
can rebase and remove the changes to the pooling tests now

maxfreu:
> can rebase and remove the changes to the pooling tests now

Sorry I don't know what rebase is. Do you mean I should simply undo the changes to the pooling tests? Probably I shouldn't have commited to the master branch of my fork. (?)

mcabbott:
You might be able to merge with this website, maybe the button says "resolve conflicts". On master there is now an alternative fix to the problem solved by "broken=spatial_rank == 0) # was == 2 before" etc.

src/upsample.jl
maxfreu: No, I think rationals are useless except for exact tests. I'm currently focusing on the GPU implementation and don't have the muse to code int support now, but feel free :) But we can also add it later, no problem.

mcabbott: Sure, happy to fiddle once this is merged.

src/upsample.jl
```diff
@@ -64,235 +64,163 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    idx_adjoint = findall(d)
-    return idx_adjoint
+function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = Base.Base.size(x)
+    if size===nothing
+        out_w = floor(Int, scale[1]*w)
+        out_h = floor(Int, scale[2]*h)
+    else
+        out_w, out_h = size
+    end
```
mcabbott, line 106: If you supply a size and a scale, perhaps it should check them for agreement? Or at least complain about being given both, rather than silently pick one? (I agree that favouring size over scale makes sense.)

maxfreu: Oh these peculiarities of writing real world code 🤣 let me fix it before you merge

CarloLucibello:
Schermata del 2021-01-21 23-52-02

@DhairyaLGandhi can you upgrade @mcabbott 's role?

DhairyaLGandhi:
Seems like we still need to make changes based on https://github.com/FluxML/NNlib.jl/pull/266#discussion_r562007525

CarloLucibello:
> Seems like we still need to make changes based on [#266 (comment)](https://github.com/FluxML/NNlib.jl/pull/266#discussion_r562007525)

that's orthogonal to my request

src/upsample.jl
mcabbott: Not critical, but might save someone a real WTF moment someday...

I don't have a merge button, in fact, but Carlo does. Could also merge now and tweak later. (I had a commit with docstrings etc. which will need to become another PR.)

CarloLucibello: > I don't have a merge button, 

@DhairyaLGandhi

DhairyaLGandhi:
Well, if it prevented premature merging then that's alright

Either way, I added the repo explicitly just in case, can you confirm?

CarloLucibello:
> Well, if it prevented premature merging then that's alright

It didn't prevent anything. The role should be org wide, or at the very minimum you should also add Flux and Zygote

DhairyaLGandhi:
Again, it's the same team.

CarloLucibello:
Is this set on Write?
https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/setting-base-permissions-for-an-organization

DhairyaLGandhi:
Base permissions would not be set to write, some teams do have write access to some repos.

maxfreu:
I need another git lesson. I implemented erroring and simple integer support, then merged the remote changes due to the resolve into my local branch and pushed (next time pull first...). The fork repo shows the correct files with error and int support. But the commits are not listed here, no CI triggered. What to do?

src/upsample.jl
```diff
@@ -64,235 +64,163 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+Currently only 2d upsampling is supported.
 """
-function adjoint_of_idx(idx::Vector{Int})
-    d = trues(length(idx))
-    d[2:end] .= diff(idx)
-    idx_adjoint = findall(d)
-    return idx_adjoint
+function upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = Base.Base.size(x)
+    if size===nothing
```
DhairyaLGandhi, line 101: Spaces

src/upsample.jl
```diff
@@ -64,235 +64,163 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
 
+Examples:
 ```julia
-x = [1, 2, 3, 4, 5]
-idx = [1, 2, 2, 3, 4, 4, 5]
-idx_adjoint = adjoint_of_idx(idx)
-@assert x[idx][idx_adjoint] == x
+upsample_bilinear(x, (2, pi)) # real scaling factors are allowed
+upsample_bilinear(x; size=(64,64)) # note the semicolon, size is a keyword argument
```
DhairyaLGandhi, line 95: Good chance to rename `size` in this PR

src/upsample.jl
```diff
@@ -64,235 +64,163 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    return ntuple(i -> i <= length(k) ? sz[i]*k[i] : sz[i], length(sz))
+upsample_bilinear(x, scale::Real; size=nothing) = upsample_bilinear(x, (scale,scale); size=size)
+
+# this is the core function which works on arrays of arbitrary size
+# the implementation is a translation of https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/UpSampleMoreKernel.cpp
+# which implements open-cv style linear interpolation / upsampling
+# for simplicity, corners are aligned and all logic for other behaviour has been stripped
+# - whcn because there is also a cwhn implementation
+# - the function is parallelized using @threads
+# - RGB types could be supported via reinterpreting
```
DhairyaLGandhi, line 119: Can we make them work out of the box

src/upsample.jl
```diff
@@ -64,235 +64,163 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+# - integer types need to be converted to Float and back
+# - rationals work, but are slow
+function upsample_bilinear_whcn!(output::AbstractArray{T,4}, input::AbstractArray{T,4}) where T
+    if size(input) == size(output)
+        return input
+    end
+    size(input)[3:4] == size(output)[3:4] || error("Number of input and output channels and batches must match. Got input $(size(input)) and output $(size(output))")
```
DhairyaLGandhi, line 126: When would this be violated?

maxfreu: When someone calls this lower level function without thinking about it.

src/upsample.jl
maxfreu: should be fixed in latest version which is not present here yet

src/upsample.jl
maxfreu: Boys, I need a final vote for outsize or size :D
Dhairya: outsize
Carlo: ?
Michael: size
Me: I'm lazy, leave size

src/upsample.jl
maxfreu: Well... yes and no. Leaving the batch dim out, assuming the data comes in form `Matrix{RGB}` we could reinterpret to `Array{UInt8, 3}` with size (3,W,H). Then we would have to permute the dims to WHC, call the function with that and convert back; rather inefficient. There's also a lazy path using PermutedDimsView, I don't know how that is performance-wise. `@avx` could *maybe* be used to rearrange the loops in the compiled code automatically, but it's also super easy to provide a function which works on cwhn directly; just switch loop order and `getindex` calls, done.

I would like to leave this for later, as this is probably no deep learning use case.

mcabbott: It is a pity that RGB types don't support scalar multiplication, otherwise this could be very simple:
```
julia> using ImageCore

julia> 0.5 * RGB(0.2, 0.3, 0.4) + 0.5 * RGB(0.6, 0.5, 0.4)
ERROR: MethodError: no method matching *(::Float64, ::RGB{Float64})
```
With reinterpret, as you say, they would end up as the first dimension not the 3rd. PermutedDimsArray could move this, without a copy, might not be inefficient... seems fine to leave until someone needs it, though. Nowhere else does this package build anything specially for RGB.

src/upsample.jl
DhairyaLGandhi: I would want this to be changed to something that isn't also an existing function to reduce ambiguity :

src/upsample.jl
DhairyaLGandhi: Well, we intend to take a differentiable programming view. Many times we support RGB implicitly. Maybe we should be adding specific tests for this? Thats somewhat separate though

maxfreu: > It is a pity that RGB types don't support scalar multiplication, otherwise this could be very simple

Aaah wait, one could in theory reinterpret to an SVector, then it should work with setting `C=1`.

> And allowing the user to specify which dimensions are channel/batch-like would add complexity.

That would maybe require implementing sth like [NamedDims.jl](https://github.com/invenia/NamedDims.jl) and dispatch on the dimension names or similar, so quite a big overhaul. I would love to see sth like that for the whole DL ecosystem so that I get get rid of all transposes during data preparation. Maybe when I retire :rofl:

src/upsample.jl
mcabbott: When in doubt, follow Base? `range(1, 10, length=100)` isn't written `range(1, 10, outlength=100)`, for good reason.

DhairyaLGandhi: Packages should do what makes sense for their users and developers. This would be better for more people who contribute and not matter for others.

mcabbott:
> The fork repo shows the correct files with error and int support. But the commits are not listed here

Ah, looks like you fixed it? I was trying to see how to cherry-pick those across, and failing... seemed like a really weird state.

maxfreu:
> Ah, looks like you fixed it?

Seems so, but I destroyed sth in the making.

src/upsample.jl
```diff
@@ -64,235 +64,175 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+    w,h,c,n = Base.size(x)
+    if scale != (1,1) && size !== nothing
+        error("Please provide either scale or size, not both. Got scale=$scale and size=$size.")
+    end
+    if size === nothing
+        out_w = floor(Int, scale[1]*w)
+        out_h = floor(Int, scale[2]*h)
+    else
+        out_w, out_h = size
+    end
```
mcabbott, line 109: I don't see why the test fails. But one comment here is that using `(1,1)` as a sentinel value might be confusing, I mean no human will deliberately write that but some program might. Perhaps they should both be `nothing`?

```suggestion
function upsample_bilinear(x::AbstractArray{T,4}, scale::Union{Nothing,NTuple{2,Real}}=nothing; size::Union{Nothing,NTuple{2,Integer}}=nothing) where {T}
    w,h,c,n = Base.size(x)
    if size === nothing && scale !== nothing
        out_w = floor(Int, scale[1]*w)
        out_h = floor(Int, scale[2]*h)
    elseif size !== nothing && scale === nothing
        out_w, out_h = size
    else
        throw(ArgumentError("Please provide either scale or size, not both. Got scale=$scale and size=$size."))
    end
```

src/upsample.jl
maxfreu: I'd like to leave this for later, therefore resolving.

src/upsample.jl
maxfreu: I think if some program sets `scale=(1,1)` it wouldn't change anything either way, or am I missing sth? Having `(1,1)` as default is just shorter and omits the union, so I'd leave it as is.

src/upsample.jl
```diff
@@ -64,235 +64,175 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-        ## Conv with views is not dispatched to CUDA.conv
-        # yy = conv(@view(Δ[:,1:nextras[2],:,:]), weight2, pad=pad2, stride=stride2)
-        # dx[:,[1],:,:] .+= conv(@view(Δ[:,1:nextras[2],:,:]), weight2, pad=pad2, stride=stride2)
-        # weight2 .= @view(weight2[:,end:-1:1,:,:])
-        # dx[:,[end],:,:] .+= conv(@view(Δ[:,end-nextras[2]+1:end,:,:]), weight2, pad=pad2, stride=stride2)
+function ∇upsample_bilinear(Δ::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
+    w,h,c,n = Base.size(Δ)
+    if scale != (1,1) && size !== nothing
+        error("Please provide either scale or size, not both. Got scale=$scale and size=$size.")
     end
```
mcabbott, line 178: And here, perhaps simplest would be never to accept `scale` at all here? It's more of an internal function, only called within the gradient rule, and at that point the desired size is known exactly. No need to think hard about whether rounding errors up & down with non-integer scale are going to work out correctly.

src/upsample.jl
mcabbott: Ah you are right, it should run fine with `(1,1)`. The only thing it won't do is throw an error with both (1,1) and a keyword, it will silently choose the keyword, but that's a pretty far-out edge case, no problem.

src/upsample.jl
maxfreu: Yes. But @CarloLucibello  implemented having the same call signature for the earlier upsampling methods, forwards and backwards and so did I in consequence. But in principle with the newer implementation, (out)size is enough. Feel free to remove it.

src/upsample.jl
```diff
@@ -64,235 +64,175 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    # TODO there must be a more convenient way to send to gpu 
-    kernel = convert(typeof(Δ), reshape(kernel, length(kernel), 1, 1, 1))
-    kernel = dropdims(kernel, dims=(2,3,4))
-    return kernel
+    return grad_input
 end
 
-function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, k)
-    Ω = upsample_bilinear(x, k)
+function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, scale; size=nothing)
```
mcabbott, line 222: I am not 100% sure, but this might also need the default scale:
```suggestion
function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, scale=(1,1); size=nothing)
```

src/upsample.jl
mcabbott: OK, I can't make one suggestion for the whole function, but step 1:
```suggestion
function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::Union{Nothing,NTuple{2,Integer}}) where T
```

src/upsample.jl
```diff
@@ -64,235 +64,175 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    if (n1 > 0) & (n2 > 0)
-        dx[1,1,:,:] .+= sum(kern[1:n1,1:n2] .* @view(Δ[1:n1,1:n2,:,:]), dims=(1,2))[1,1,:,:]
-        dx[1,end,:,:] .+= sum(kern[1:n1,end-n2+1:end] .* @view(Δ[1:n1,end-n2+1:end,:,:]), dims=(1,2))[1,1,:,:]
-        dx[end,end,:,:] .+= sum(kern[end-n1+1:end,end-n2+1:end] .* @view(Δ[end-n1+1:end,end-n2+1:end,:,:]), dims=(1,2))[1,1,:,:]
-        dx[end,1,:,:] .+= sum(kern[end-n1+1:end,1:n2] .* @view(Δ[end-n1+1:end,1:n2,:,:]), dims=(1,2))[1,1,:,:]
+    if size===nothing
+        out_w = ceil(Int, w/scale[1])
+        out_h = ceil(Int, h/scale[2])
+    else
+        out_w, out_h = size
```
mcabbott, line 183: Step 2:
```suggestion
    out_w, out_h = size
```

src/upsample.jl
```diff
@@ -64,235 +64,175 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    step = 1//n
-    if n % 2 == 0
-        start = step//2
-        upward = collect(start:step:1//1)
-        kernel = [upward; reverse(upward)]
-    else
-        start = step
-        upward = collect(start:step:1//1)
-        kernel = [upward; reverse(upward[1:end-1])]
+∇upsample_bilinear(Δ, scale::Real; size=nothing) = ∇upsample_bilinear(Δ, (scale, scale); size=size)
```
mcabbott, line 189: Step 3:
```suggestion
```
I think the one-scale methods are going in the Flux PR

src/upsample.jl
maxfreu: AAhh sorry man, maybe I have wasted your time here. I think it might be better to leave it for now. Otherwise you also have to fix the tests. And size would need to be positional. And the rrule needs to be changed. My heady is already dizzy.

mcabbott: I think the RRule is OK, it always calls with one keyword. It's OK to have a mandatory keyword (less confusing than switching to positional I think).

test/upsample.jl
```diff
@@ -13,32 +13,65 @@
...
-        g_cpu = Zygote.gradient(x -> sum(sin.(upsample_bilinear(x, (3, 2))))
-                                , x)[1]
-        @test Array(g_cpu) ≈ g_cpu  atol=1e-4
-    end
+    gradtest(x->upsample_bilinear(x, (3, 2)), x, atol=1e-3) # works to higher precision for Float64
+
+    # additional grad check, also compliant with pytorch
+    o = ones(Float32,6,4,2,1)
+    grad_true = 6*ones(Float32,2,2,2,1)
+    @test ∇upsample_bilinear(o, (3,2)) ≈ grad_true
```
mcabbott, line 42: Step N.... I think this is the only test which needs changing? 
```suggestion
    @test ∇upsample_bilinear(o; size = (2,2)) ≈ grad_true
```

src/upsample.jl
CarloLucibello: Given the precedent of range's length I vote size, its clear, ts shorter, it sounds better, makes sense for both users and developers (I'm sure no one is going to be tripped by this)

src/upsample.jl
DhairyaLGandhi: We could have dispatch rules too,

maxfreu: Dispatch in which sense? I dont know what you mean here.

src/upsample.jl
maxfreu: I‘d like to leave them in so that one can use it in a functional way in case it’s needed.

mcabbott: But do you mean the `upsample_bilinear(x, scale::Real; size=nothing)` method, or the gradient?

maxfreu: Sry I meant the forward pass. I’m on phone and didn’t look close enough. I’ll go through all this next week. Have a nice weekend! :)

src/upsample.jl
maxfreu: Ok, I'll change the GPU code to `size` as well then.

src/upsample.jl
maxfreu: @CarloLucibello I would like to hear your opinion on the gradient signature, thinking of the discussion in https://github.com/FluxML/NNlib.jl/pull/262

src/upsample.jl
```diff
@@ -64,235 +64,166 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-    end
-    if (size(Δ, 1) == 1) && (size(Δ, 2) == 1)
-        dx = Δ
-        return dx
+function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
+    out_w, out_h = size
     end
+    dx = zeros(T, out_w, out_h, c, n)
+    return ∇upsample_bilinear_whcn!(Δ, dx)
+end
```
mcabbott, line 179:
```suggestion
    dx = zeros(T, out_w, out_h, c, n)
    return ∇upsample_bilinear_whcn!(Δ, dx)
end
```
Probably one `end` is enough, now that I see all the pieces together...

maxfreu:
Ok, grad API change as suggested/requested by Michael is in, tests are passing. If there are no more objections, we (you) can merge. I will update the corresponding CUDA PR and report back after it has been merged. In the meantime bilinear upsampling wont work on GPU, but I think that's no problem at this early stage.

DhairyaLGandhi:
Cool, let me do a final pass before we make this into an api to support

src/upsample.jl
```diff
@@ -64,235 +64,166 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
 
-    return dx
+# Outputs
+- `dx`: Downsampled version of `Δ`
+"""
+function ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
+    _, _, c, n = Base.size(Δ)
+    out_w, out_h = size
+    dx = zeros(T, out_w, out_h, c, n)
+    return ∇upsample_bilinear_whcn!(Δ, dx)
```
mcabbott, line 179: My last complaint is that I'd prefer here `∇upsample_bilinear_whcn!(dx, dy)`, i.e. that the mutating function mutate its first argument, not its second. I don't swear this package is completely consistent on that but this would be the standard pattern. 

It is just an internal function, though, since I see your Cu PR doesn't overload this, but instead overloads the allocating version. 

The alternative pattern would have been to write `dx = similar(x, T, ints...)` which should automatically be the same Array/CuArray type as `x`, and then have the mutating `∇upsample!` dispatch on that. That (and the same pattern for the forward pass) might reduce the need to duplicate logic about `floor(Int, width_scale*w)` etc. But I think it's also OK to leave it.

DhairyaLGandhi:
We should really have the order of gradient related function arguments be `(dy, dx)` since it's more descriptive about the intention of the function. That is the standard pattern since the chain rule operates on the output of the previous transform which the gradient functions accept as the first args

maxfreu:
Ok now I'm confused. Is `dy = Δ`? Also to be honest, I'm not completely familiar with the differentiable programming terminology. I think julia guidelines tell us to have the mutated argument first, right? So in the case of `∇upsample_bilinear_whcn_kernel!(dx, Δ)` (after last commit), Δ is the gradient backpropagated from downstream layers and dx is the the function output (the "wobble" to produce Δ). `dx` is mutated by the function, `Δ` is not - so this would comply with the guidelines. So if I should change it back, please bump me once again.

As a side note: I created another layer of indirection, wrapping the kernel in a `(∇)upsample_bilinear!` function. This allows me to remove all size related code from the CUDA PR and overload only the mutating functions (which calculate thread count and block size then).

mcabbott:
Yes, that looks fine. I like `dx` for the perturbation of input `x`, and `dy` for perturbation of output `y`. For reverse mode AD, we get `dy` and make `dx`. Some people like adding bars to things, and mostly `Δ` seems to mean `dy`.

Do `∇upsample_bilinear_whcn_kernel!` and `∇upsample_bilinear!` need to be distinct functions at all?

I think the `if size(dx) == size(Δ)` check (trivial case) can move earlier, to be done before allocating `dx` at all.

DhairyaLGandhi:
Again, the way this API was set was for syntactic consistency. This is not to say that we shouldn't follow the "mutated argument first" anywhere, but Base does not *dictate* how libraries can use the flexibility of the language. I would keep the consistency here. What is mutated and what isn't can be discussed as a separate issue.

src/upsample.jl
```diff
@@ -64,235 +64,174 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-        dx[:,[1],:,:] .+= conv(Δ[:,1:nextras[2],:,:], weight2, pad=pad2, stride=stride2)
-        weight2 .= weight2[:,end:-1:1,:,:]
-        dx[:,[end],:,:] .+= conv(Δ[:,end-nextras[2]+1:end,:,:], weight2, pad=pad2, stride=stride2)
-
-        ## Conv with views is not dispatched to CUDA.conv
-        # yy = conv(@view(Δ[:,1:nextras[2],:,:]), weight2, pad=pad2, stride=stride2)
-        # dx[:,[1],:,:] .+= conv(@view(Δ[:,1:nextras[2],:,:]), weight2, pad=pad2, stride=stride2)
-        # weight2 .= @view(weight2[:,end:-1:1,:,:])
-        # dx[:,[end],:,:] .+= conv(@view(Δ[:,end-nextras[2]+1:end,:,:]), weight2, pad=pad2, stride=stride2)
+function ∇upsample_bilinear_whcn_kernel!(dx::AbstractArray{T,4}, Δ::AbstractArray{T,4}) where T
```
DhairyaLGandhi, line 186:
```suggestion
function ∇upsample_bilinear_whcn_kernel!(Δ::AbstractArray{T,4}, dx::AbstractArray{T,4}) where T
```

DhairyaLGandhi (changes requested):

src/upsample.jl
maxfreu: Ok, I'll change it back then :)

mcabbott:
> Again, the way this API was set was for syntactic consistency
> I would keep the consistency here. 

Which other functions do you have in mind? These place the mutated `dx` first:
```
methods(NNlib.∇maxpool!)
methods(NNlib.∇softmax!)
methods(NNlib.∇conv_data!) # clearer in comments than in variable names
```

src/upsample.jl
mcabbott: I think the package elsewhere puts mutated first like this.

DhairyaLGandhi:
I'm pretty certain that is context dependent? And maybe some of those functions need to be updated accordingly then.

maxfreu:
> I think the if size(dx) == size(Δ) check (trivial case) can move earlier, to be done before allocating dx at all.

I was only thinking about the overhead of the comparison itself, not the allocation, right.

> Do ∇upsample_bilinear_whcn_kernel! and ∇upsample_bilinear! need to be distinct functions at all? Even without this step, the other PR could overload ∇upsample_bilinear!(::CuArray, ...) and be sure to get the answer.

The arguments to the GPU kernel and this one are a bit different. I could bring both in line, but this would require to hoist some of the logic out of the CPU kernel, which would make things less clear maybe, but it works. My thoughts about the API go like this:

```julia
const NDA = NamedDimsArray
# these two could be fused into one, yes. The parent() call would have to go into the kernel then.
upsample_bilinear!(y::NDA{(:w,:h,:c,:n)}, x::...) = upsample_bilinear_whcn_kernel!(parent(y), parent(x))
upsample_bilinear!(y::NDA{(:c,:w,:h,:n)}, x)      = upsample_bilinear_cwhn_kernel!(parent(y), parent(x))

function upsample_bilinear!(y::NDA{(:w,:h,:c,:n),T,N,A}, x::...) where {T, N, A<:CuArray}
  a,b,c = ...
  threads = ...
  blocks = ...
  @cuda threads blocks upsample_bilinear_whcn_kernel!(a,b,c, parent(x), parent(y))  # <- the GPU kernel args are a bit different
  return y
end

upsample_bilinear!(y::CWHN_GPU_TYPE, x) = ...
```

mcabbott:
OK. My suggestion here would look more like this:
```julia
upsample_bilinear!(y, x) = upsample_bilinear_whcn!(y, x)  # default setting

function upsample_bilinear_whcn!(y::AbstractArray, x:: AbstractArray)
  # direct implementation as in this PR
end

# This worker has one job, very simply dispatch, will never change:
function upsample_bilinear_whcn!(y::CuArray, x::...)
  a,b,c = ...
  threads = ...
  blocks = ...
  @cuda threads blocks upsample_bilinear_whcn_kernel!(a,b,c, parent(x), parent(y))  # the real GPU kernel
  return y
end

# These two workers can be added later, without breaking anything:
upsample_bilinear_cwhn!(y::AbstractArray, x:: AbstractArray)
upsample_bilinear_cwhn!(y::CuArray, x::...) = ...

const NDA = NamedDimsArray
# Only one function dispatches on NDA, and it does not need to load CUDA:
upsample_bilinear!(y::NDA{(:w,:h,:c,:n)}, x::...) = upsample_bilinear_whcn!(parent(y), parent(x))
upsample_bilinear!(y::NDA{(:c,:w,:h,:n)}, x)      = upsample_bilinear_cwhn!(parent(y), parent(x))
# Although perhaps in reality it would be the non-mutating one, not sure `similar` will otherwise produce names...
```

Re argument order, it looks fine I think, Dhairya got me worried that we were all over the map in this package, but the examples I can find seem pretty consistent. So this PR should match those, and it does.

src/upsample.jl
maxfreu: I think it's ok to keep it as, is. Look at e.g.:

https://github.com/FluxML/NNlib.jl/blob/d9aaaf7ac4df683f1d3361efb3cbaef9aa8112a1/src/conv.jl#L101-L103

maxfreu:
> OK. My suggestion here would look more like this:

Aah yes, molto bene :) Will massage it tomorrow. The rest should maybe be discussed on slack or zulip or so - where are you?

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-            weight2[:,:,i,i] .= kern2
-        end
-        # weight2 = cat(fill(kern2, n_chan)..., dims=(3,4)) # slow
-    
-        yy = conv(Δ[:,1:nextras[2],:,:], weight2, pad=pad2, stride=stride2)
-        dx[:,[1],:,:] .+= conv(Δ[:,1:nextras[2],:,:], weight2, pad=pad2, stride=stride2)
-        weight2 .= weight2[:,end:-1:1,:,:]
-        dx[:,[end],:,:] .+= conv(Δ[:,end-nextras[2]+1:end,:,:], weight2, pad=pad2, stride=stride2)
+"""
+    ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::Union{Nothing,NTuple{2,Integer}}=nothing) where T
```
CarloLucibello, line 166:
```suggestion
    ∇upsample_bilinear(Δ::AbstractArray{T,4}; size::NTuple{2,Integer}) where T
```

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
 
+Examples:
 ```julia
-x = [1, 2, 3, 4, 5]
-idx = [1, 2, 2, 3, 4, 4, 5]
-idx_adjoint = adjoint_of_idx(idx)
-@assert x[idx][idx_adjoint] == x
+upsample_bilinear(x, (2, pi)) # real scaling factors are allowed
+upsample_bilinear(x; size=(64,64)) # note the semicolon, size is a keyword argument
```
CarloLucibello, line 95:
```suggestion
upsample_bilinear(x; size=(64,64)) # specify ouput size
```

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+    lambda0 = 1 - lambda1
+    return input_index0, input_index1, lambda0, lambda1
 end
 
 """
-    adjoint_of_idx(idx::Vector{<:Integer})
-
-# Arguments
-- `idx`: a vector of indices from which you want the adjoint.
+    upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real}=(1,1); size::Union{Nothing,NTuple{2,Integer}}=nothing)
```
CarloLucibello, line 84:
```suggestion
    upsample_bilinear(x::AbstractArray{T,4}, scale::NTuple{2,Real})
    upsample_bilinear(x::AbstractArray{T,4}; size::NTuple{2,Integer}) 
```

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
 
-# Outputs
--`idx_adjoint`: index that inverses the operation `x[idx]`.
+Upsamples the first 2 dimensions of the array `x` by the upsample factors stored in `scale`,
+using bilinear interpolation.
```
CarloLucibello, line 87:
```suggestion
using bilinear interpolation.

As an alternative to using `scale`, the resulting image `size` can be directly specified with a keyword argument.
```

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
 
-# Explanation
-Determines the adjoint of the vector of indices `idx`, based on the following assumptions:
-* `idx[1] == 1`
-* `all(d in [0,1] for d in diff(idx))`
-The adjoint of `idx` can be seen as an inverse operation such that:
+The size of the output is equal to
+`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.
```
CarloLucibello, line 90:
```suggestion
`(scale[1]*S1, scale[2]*S2, S3, S4)`, where `S1, S2, S3, S4 = size(x)`.

As an alternative to using `scale`, the resulting image `size` can be directly specified with a keyword argument.
```

src/upsample.jl
```diff
@@ -64,235 +64,169 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
-- `dx`: downsampled version of `Δ`
-
-# Explanation
-
-Custom adjoint for [`upsample_bilinear`](@ref). 
-The adjoint of upsampling is a downsampling operation, which
-in this implementation is performed using `NNlib.conv` in combination with a downsampling kernel based on the
-upsampling factors. Because of the zero-padding during convolution, the values at the boundary are polluted by edge-effects,
-which have been corrected for manually.
+Currently only 2d upsampling is supported.
```
CarloLucibello, line 97:
```suggestion
```

src/upsample.jl
mcabbott: xref https://github.com/FluxML/NNlib.jl/pull/266#discussion_r562678254  above, a similar suggestion.

CarloLucibello: Maybe this is more clear, even if not completely faithful to the actual definition. Actually, this could become the new definition, what do you say @maxfreu

CarloLucibello:
Tried to commit some of the suggestions but don't have the rights

src/upsample.jl
maxfreu: Yes, that looks more clear. The python programmer in the back of my mind crammed everything into a single function call.

mcabbott: If you make two methods, and one figures stuff out then calls the other one, I think it will work to attach the `rrule` only to the last one -- the one which is always called and does the work. Which here would be the `size` option.

src/upsample.jl
```diff
@@ -64,235 +64,166 @@ function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::T
...
+    return dx
 end
 
-function ChainRulesCore.rrule(::typeof(upsample_bilinear), x, k)
-    Ω = upsample_bilinear(x, k)
+function ChainRulesCore.rrule(::typeof(upsample_bilinear), x; size)
+    Ω = upsample_bilinear(x; size=size)
     function upsample_bilinear_pullback(Δ)
-        (NO_FIELDS, ∇upsample_bilinear(Δ, k), DoesNotExist())
+        (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))), DoesNotExist())
```
CarloLucibello, line 216:
```suggestion
        (NO_FIELDS, ∇upsample_bilinear(Δ; size=(Base.size(x,1),Base.size(x,2))))
```

CarloLucibello:
@DhairyaLGandhi I lost write access

DhairyaLGandhi:
Haven't changed anything, what does it say?

src/upsample.jl
DhairyaLGandhi: Is it this that's meant to be committed?

mcabbott: Good catch. Should we be disturbed that CI didn't catch this?

DhairyaLGandhi: If this is needed, to be correct, it should be positional no?

CarloLucibello:
I don't know what's happening, I lost write access  yesterday, couldn't see the Merge Pull Request button, but it reappeared right now

src/upsample.jl
CarloLucibello: this is a consequence of the method split in last commit

DhairyaLGandhi: The api should reflect that then

CarloLucibello (approved):

src/upsample.jl
CarloLucibello: I don't understand

DhairyaLGandhi:
Please address the comment inhe api before merging

DhairyaLGandhi:
Please address the comment inhe api before merging

src/upsample.jl
CarloLucibello: @DhairyaLGandhi can you be more explicit?

DhairyaLGandhi: If the methods were split, then there should be corresponding functions that should be able to completely specify the expected outcome with those new split methods.

mcabbott: Is it concerning that CI didn't complain? 

There is a test which runs this gradient. Maybe charinrules etc. just takes the first N elements of the tuple, without checking its length, and so just ignores any trailing elements.

Seems far removed from API concerns, not sure what you are thinking. Nor why you suggest it should be positional? Of course things could be designed other ways, but at present the size is always a keyword.

DhairyaLGandhi: So you could have a method which take the dimensions as positional in this case.

mcabbott: Yea, other designs are possible.

DhairyaLGandhi: It may not have been caught since we can't exactly guarantee the number of partials any such call might have. But yes, we should do the correct thing else we'll get the incorrect values, so no question on the change.

The split methods may be an artifact of refactoring, but the downstream functions should be consistent and if they need the size given explicitly in some cases, then they need to accept them, and the need aspect is determined by it being positional.

mcabbott: > The split methods may be an artifact of refactoring, but the downstream functions should be consistent and if they need the size given explicitly in some cases, then they need to accept them, and the need aspect is determined by it being positional.

This is pretty hard to parse. Can you write a code sample showing us what you have in mind?

By "split" Carlo just means that there was a change from one function body with if-statements, to using multiple dispatch. He doesn't mean "broken" or something. Obviously things must work, if you think there are problems with "downstream functions" not getting what they need, then I think you ought to say exactly which function you mean. But revisiting the whole design is well beyond the scope of this one-line correction.

maxfreu:
Hi, I'm a bit late to the party: There have been many comments on the API - which changes do you refer to? This one https://github.com/FluxML/NNlib.jl/pull/266#issuecomment-768208127?

CarloLucibello:
> Hi, I'm a bit late to the party: There have been many comments on the API - which changes do you refer to? This one [#266 (comment)](https://github.com/FluxML/NNlib.jl/pull/266#issuecomment-768208127)?

The comment is in this thread https://github.com/FluxML/NNlib.jl/pull/266#discussion_r568376383, but I don't think there is anything to address. I'll merge in 1 day if no objections arise.

DhairyaLGandhi:
Well, since it's an api change, I'd be careful not to merge without proper checks.

maxfreu:
Thanks everybody for your time and efforts in making this better! :) I'll try to finish the GPU PR next week, depending on Tim's occupancy.

maxfreu:
https://github.com/JuliaGPU/CUDA.jl/pull/636 has been merged. After the next release tag we'll be at warp speed :)

A quick test with (32,32,1024,1) on my GTX980 shows 3.3us for bilinear upsampling vs 4.4us for nearest, so I recommend the former for now (some day nearest will be faster). On CPU single threaded they are about the same, but bilinear can take advantage of more cores.

CarloLucibello:
That was some great work!

roflmaostc:
Reading all these several issues/PRs about the bilinear upsampling layer is like reading a good book. It was quite exciting to follow the whole discussions and seeing the final performance and outcome :laughing: 
Great work!

PR #269, opened 2021-01-17
Add `upsample_nearest`

mcabbott:


maxfreu:
Hi! Does this support GPU?

mcabbott:
I haven't tried it, but I can't see why it wouldn't.

mcabbott:
I will remove the `upsample_x(::Array, ::Int)` methods, I think, as there is no need for NNlib to be too strongly attached to what the last two dimensions mean. 

If #266 is going to be done soon, perhaps it's easiest to fix up these doc changes after that. (I know it's poor style to mix things slightly, but I'm no good at juggling too many branches locally...)

maxfreu:
Looks neat now! :)

> but I'm no good at juggling too many branches locally.

And I'm not good at merging intertwined stuff, that's why I asked :rofl:

src/upsample.jl
```diff
@@ -1,4 +1,68 @@
...
+        s = scales[cld(d,2)]
+        isodd(d) ? s : div(size(x, cld(d,2)),s)
+    end
+    mid = sum(reshape(x, tempsize), dims=ntuple(d -> 2d-1, S))
+    reshape(mid, outsize)
+end
+
+function ChainRulesCore.rrule(::typeof(upsample_nearest), x::AbstractArray, s::Tuple)
+    Ω = upsample_nearest(x, s)
+    upsample_nearest_pullback(Δ) = (NO_FIELDS, ∇upsample_nearest(Δ, k), DoesNotExist())
```
CarloLucibello, line 63:
```suggestion
    upsample_nearest_pullback(Δ) = (NO_FIELDS, ∇upsample_nearest(Δ, s), DoesNotExist())
```

src/upsample.jl
```diff
@@ -1,4 +1,68 @@
...
+    end
+    readsize = ntuple(N+S) do d
+        d > 2S && return size(x, d-S)
+        isodd(d) ? 1 : size(x, cld(d,2))
+    end
+    reshape(out, writesize) .= reshape(x, readsize)
+    out
+end
+
+function ∇upsample_nearest(x::AbstractArray{T,N}, scales::NTuple{S, <:Integer}) where {T,N,S}
```
CarloLucibello, line 46: x  is supposed to be the input or the incoming sensitivities? If the latter, calling it Δ could be clearer

test/upsample.jl
```diff
@@ -1,3 +1,15 @@
+@testset "upsample_nearest, integer scale via reshape" begin
+    x = reshape(Float32[1. 2.; 3. 4.], (2,2,1,1))
+    @test upsample_nearest(x, (3,3))[1,:] == [1,1,1, 2,2,2]
+
+    y = upsample_nearest(x, (2,3))
+    @test size(y) == (4,6,1,1)
+    ∇upsample_nearest(y, (2,3)) == [6 12; 18 24]
+
```
CarloLucibello, line 8:
```suggestion
gradtest(x -> upsample_nearest(x, (2,3)), rand(2,2,1,1), check_rrule=true)
```

src/upsample.jl
mcabbott: Good catch, thanks

test/upsample.jl
```diff
@@ -1,3 +1,15 @@
+@testset "upsample_nearest, integer scale via reshape" begin
+    x = reshape(Float32[1. 2.; 3. 4.], (2,2,1,1))
+    @test upsample_nearest(x, (3,3))[1,:] == [1,1,1, 2,2,2]
+
+    y = upsample_nearest(x, (2,3))
+    @test size(y) == (4,6,1,1)
+    ∇upsample_nearest(y, (2,3)) == [6 12; 18 24]
+gradtest(x -> upsample_nearest(x, (2,3)), rand(2,2,1,1), check_rrule=true)
```
mcabbott, line 8:
```suggestion

    gradtest(x -> upsample_nearest(x, (2,3)), rand(2,2,1,1), check_rrule=false)

```

CarloLucibello: ah right, cannot check the closure, doesn't matter anyway, zygote testing is enough

src/upsample.jl
mcabbott: Was going to call it `downsample_nearest` until I realised that would imply `mean` while the gradient wants `sum`. Can change `x` if you think it too confusing.

CarloLucibello: I don't see a precedent for downsample nearest in other libraries, so I think this is fine as it is

PR #270, opened 2021-01-18
Partially revert #259

mcabbott:
and add more comments about where Compat.jl is actually being used.


CarloLucibello:
so I guess I confused myself in #259 . Thanks for catching this

mcabbott:
No problem. Did not look into these maxpool gradients, but it appeared that fewer tests were failing than previously marked in the tests.

PR #273, opened 2021-01-20
Make softmax understand `Inf`

### LINKED ISSUE #206: softmax([Inf, 1.0]) gives NaNs
```julia
julia> NNlib.softmax([Inf, 1.0])
2-element Array{Float64,1}:
 NaN
 NaN
```

The result should be `[1.0, 0.0]`.

mcabbott:
Closes #206

When I try to time it, this check costs about 1% when everything is finite, which seems completely fine.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -103,14 +103,11 @@ function logsoftmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T
     if all(isfinite, max_)
         out .= x .- max_
     else
-        # @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 0, -Inf), out - log_)
-        @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 0, _lowestlog(T)), x - max_)
+        @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 0, -Inf), x - max_)
     end
     log_ = log.(sum(exp, out; dims = dims))
     out .-= log_
 end
-
-_lowestlog(::Type{T}) where {T} = log(nextfloat(zero(float(T))))
 
 ∇logsoftmax(Δ::AbstractArray{T}, x::AbstractArray, y::AbstractArray{S}; dims = 1) where {T,S} =
     ∇logsoftmax!(similar(y, promote_type(T, S)), Δ, x, y; dims = dims)
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -51,11 +51,12 @@ softmax!(x; dims = 1) = softmax!(x, x; dims = dims)
 
 function softmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
     max_ = maximum(x; dims = dims)
-    out .= exp.(x .- max_)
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
Inf doesn't seem that common a use case to drop performance on, the current behaviour is to error iirc

mcabbott:
> Inf doesn't seem that common a use case to drop performance on

Let us remember that #248 discovered a *factor of 2* speed problem in *exactly this function* less than a month ago. To have some context for balancing correctness against speed, here.

> the current behaviour is to error 

Issue #206, linked above, shows the current behaviour. 
```
julia> NNlib.softmax([Inf, 1.0])
2-element Array{Float64,1}:
 NaN
 NaN
 ```

File: src/softmax.jl [hunk 1]
```diff
@@ -94,14 +98,18 @@ logsoftmax(x; dims = 1) = logsoftmax!(similar(x, (float ∘ eltype)(x)), x; dims
...
-function logsoftmax!(out::O, x::T; dims = 1) where {O<:AbstractArray,T<:AbstractArray}
-    out .= x .- maximum(x; dims = dims)
-    # out .-= log.(sum(exp.(out); dims = dims))  # WARN: this will decrease performance.
+function logsoftmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
+    max_ = maximum(x; dims = dims)
+    if all(isfinite, max_)
+        out .= x .- max_
+    else
+        @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 0, -Inf), x - max_)
+    end
```
CarloLucibello, line 107: why you are using a slightly different pattern here compared to the softmax! change?

CarloLucibello:
Maybe we could apply this change only to softmax, since its more likely to be used only after training when out in the wild, and leave logsoftmax be to squeeze every tiny performance fraction for training? I'm not sure this suggestion makes sense :) 
I don't think that 1% is a big deal, at least right now, but also it's hard to gauge how much this change would be relevant. Maybe @cossio could help here

src/softmax.jl
mcabbott: I don't remember exactly. In the case which does have `Inf`, this version does one pass over `out`, while the other does two. Will check that making `softmax!` do the one-pass version doesn't hurt anything.

mcabbott:
1% was a quick test, one could try to be more systematic. It's one cheap pass over a length-N array, compared to `log` or `exp` over an N^2 matrix, so it's difficult to imagine it being important. And elsewhere we have, for instance, chosen to *allocate* another length-N array, rather than mutate an existing one, again on the grounds that it can't matter much (and the mutation would need avoid some edge cases).

What would matter a lot is changing the `log` / `exp` to be vectorised versions, that's a factor 2 to 4 effect. When last I checked the plan was to split this package into a Base version and an accelerated CPU version. But from where I stand it's pretty difficult to picture getting that off the ground.

CarloLucibello (approved):

CarloLucibello:
@DhairyaLGandhi what's going on here, why I don't have write access anymore?

src/softmax.jl
mcabbott: Now changed to match.

Issue #274, opened 2021-01-27
DenseConvDims not always type stable


joostveenema:
The following code example shows that the type of `DenseConvDims` cannot be properly inferred when it's input types are all NTuples. When changing one of the inputs to an array it can for some reason be correctly inferred. 

This seems to be the source of https://github.com/FluxML/Flux.jl/issues/1350. Maybe linked to https://github.com/FluxML/NNlib.jl/issues/125?

```
using NNlib
using InteractiveUtils

function main()
    array = [1,1,1,1]::Array{Int64, 1}
    ntuple = (1,1,1,1)::NTuple{4, Int64}

    # type stable
    @code_warntype NNlib.DenseConvDims(array, ntuple)
    @code_warntype NNlib.DenseConvDims(ntuple, array)
    
    # type unstable
    @code_warntype NNlib.DenseConvDims(ntuple, ntuple)
end
main()
```
```
Variables
  #self#::Type{DenseConvDims}
  x_size::NTuple{4,Int64}
  w_size::NTuple{4,Int64}

Body::DenseConvDims{2,_A,_B,_C,_D,_E,_F,_G} where _G where _F where _E where _D where _C where _B where _A
1 ─ %1 = NNlib.:(var"#DenseConvDims#6")(1, 0, 1, false, #self#, x_size, w_size)::DenseConvDims{2,_A,_B,_C,_D,_E,_F,_G} where _G where _F where _E where _D where _C where _B where _A
└──      return %1
```

CarloLucibello:
glad you found this, should be easy to fix

CarloLucibello:
Actually I don't think the constructor can be made type stable, since the type parameter is a runtime input value:
```julia
julia> struct A{N}; x::Int; end

julia> A(i) = A{i}(10)
A

julia> @code_warntype A(1)
Variables
  #self#::Type{A}
  i::Int64

Body::A{_A} where _A
1 ─ %1 = Core.apply_type(Main.A, i)::Type{A{_A}} where _A
│   %2 = (%1)(10)::A{_A} where _A
└──      return %2
```
@joostveenema why do you think this is related to FluxML/Flux.jl#1350?

CarloLucibello:
> When changing one of the inputs to an array it can for some reason be correctly inferred.

they are not inferred, the call just throws an error

joostveenema:
You are right, the code example does not make sense. I thought I had simplified the problem. Let's try again.

[...]

PR #275, opened 2021-02-04
Force return type of channels_in and channels_out

joostveenema:
This change fixes the errors from my examples in https://github.com/FluxML/Flux.jl/issues/1350#issuecomment-765348942.

Solution taken from https://github.com/FluxML/NNlib.jl/issues/274#issuecomment-773129949.

### DIFF
--- a/src/dim_helpers/DenseConvDims.jl
+++ b/src/dim_helpers/DenseConvDims.jl
@@ -12,8 +12,8 @@ end
 # Getters for the fields
 input_size(c::DenseConvDims) = c.I
 kernel_size(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = K
-channels_in(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_in
-channels_out(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_out
+channels_in(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_in::Int
+channels_out(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_out::Int
 
 # Convenience wrapper to create DenseConvDims objects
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
I think you should never need to do that (also I think the better syntax would be `f(...)::Int = ...`

Cc @staticfloat

CarloLucibello:
> the better syntax would be f(...)::Int 

I tried that but didn't understand how to combine it with the `where` statement. In any case, this is entirely equivalent to what is being done here

src/dim_helpers/DenseConvDims.jl
```diff
@@ -12,8 +12,8 @@ end
 # Getters for the fields
 input_size(c::DenseConvDims) = c.I
 kernel_size(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = K
-channels_in(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_in
-channels_out(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_out
+channels_in(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_in::Int
+channels_out(c::DenseConvDims{N,K,C_in,C_out,S,P,D,F}) where {N,K,C_in,C_out,S,P,D,F} = C_out::Int
```
CarloLucibello, line 16:
```suggestion
channels_in(c::DenseConvDims{N,K,C_in,C_out}) where {N,K,C_in,C_out} = C_in::Int
channels_out(c::DenseConvDims{N,K,C_in,C_out}) where {N,K,C_in,C_out} = C_out::Int
```
a good chance to avoid some eye bleeding

CarloLucibello:
Could you add `@inferred` tests?

staticfloat:
> I tried that but didn't understand how to combine it with the where statement.

Looks like this works in long-form but not short-form.  That may be a parsing bug that gets fixed eventually, I agree that simply annotating the return statement is equivalent here, so that's fine.

> I think you should never need to do that

In an ideal world, I agree.  Unfortunately, our type inferrence algorithms really struggle with some NNlib code, so if we can do this to help it out, I say let's go for it.

CarloLucibello:
thanks!

PR #278, opened 2021-02-06
Upsample docstring improvements

mcabbott:
This was a commit of #269 which got reverted so as not to cause chaos.

Should have no functional changes. Code for `pixel_shuffle` is altered to remove `@assert`.

### DIFF
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -261,11 +261,12 @@ end
     pixel_shuffle(x, r::Integer)
 
 Pixel shuffling operation, upscaling by a factor `r`.
-The operation converts an input of size [W,H,r²C,N] to size [rW,rH,C,N].
 
-Used extensively in super-resolution networks to upsample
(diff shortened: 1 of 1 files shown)


src/upsample.jl
```diff
@@ -234,25 +258,72 @@ function ChainRulesCore.rrule(::typeof(upsample_bilinear), x; size)
 end
 
 """
-    pixel_shuffle(x, r)
+    pixel_shuffle(x, r::Integer)
+
+Pixel shuffling operation, upscaling by a factor `r`.
+The operation converts an input of size [W,H,r²C,N] to size [rW,rH,C,N].
```
CarloLucibello, line 264: Can we state somehow that pixel_shuffle supports arbitrary dimensions?

mcabbott: I put an example of a "1D image" , but perhaps these words should change too.

PR #279, opened 2021-02-21
add pad_constant, pad_repeat, pad_reflect, pad_zeros

CarloLucibello:
Maybe not the most optimized implementation, but it is simple, generic, AD and gpu friendly.

Other padding modes should not be too hard to implement similarly 
cfr. https://pytorch.org/docs/stable/nn.html#padding-layers

### DIFF
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -1,0 +1,209 @@
+export pad_constant, pad_repeat, pad_reflect, pad_zeros
+
+"""
+    pad_zeros(x, pad::Tuple; [dims])
+    pad_zeros(x, pad::Int; [dims])
+
+Pad the array `x` with zeros.
+Equivalent to [`pad_constant`](@ref) with the constant equal to 0. 
+"""
+pad_zeros(x::AbstractArray, pad::NTuple{M,Int}; dims=1:M÷2) where M =
+  pad_constant(x, pad, 0; dims=dims)
+
+"""
+    pad_constant(x, pad::Tuple, val=0; [dims])
+    pad_constant(x, pad::Int, val=0; [dims])
+
+Pad the array `x` with the constant value `val`.
+
+`pad` can a tuple of integers `(l1, r1, ..., ln, rn)`
+of some length `2n` that specifies the left and right padding size
+for each of the dimensions in `dims`. If `dims` is not given, 
+it defaults to the first `n` dimensions.
+
+For integer `pad` input instead, it is applied on both sides
+on every dimension in `dims`. In this case, `dims` 
+defaults to the first `ndims(x)-2` dimension 
+(i.e. excludes the channel and batch dimension).
+
+See also [`pad_reflect`](@ref) and [`pad_repeat`](@ref).
+
+```jldoctest
+julia> pad_constant(reshape(1:4, 2, 2), (1, 2, 3, 4), 8)
+5×9 Matrix{Int64}:
+ 8  8  8  8  8  8  8  8  8
+ 8  8  8  1  3  8  8  8  8
+ 8  8  8  2  4  8  8  8  8
+ 8  8  8  8  8  8  8  8  8
+ 8  8  8  8  8  8  8  8  8
+````
+"""
+function pad_constant(x::AbstractArray, pad::NTuple{M,Int}, val=0; 
+                    dims=1:M÷2) where M
+  length(dims) == M÷2 ||
+    throw(ArgumentError("The number of dims should be equal to the number of padding dimensions"))
+  outsize, center = pad_outsize_and_center(size(x), pad, dims)
+  y = fill!(similar(x, eltype(x), outsize), val)
+  y[center...] .= x
+  return y
+end
+
+function pad_outsize_and_center(sz::NTuple{N,Int}, pad, dims) where {N}
+  leftpad, rightpad = pad[1:2:end], pad[2:2:end]
+  
+  outsize = ntuple(N) do i
+              k = findfirst(==(i), dims)
+              if k === nothing
+                return sz[i]
+              else
+                return sz[i] + leftpad[k] + rightpad[k]
+              end
+            end::NTuple{N,Int}
+  
+  leftcorner = ones(Int, N)
+  rightcorner = collect(outsize)
+  for (i, d) in enumerate(dims)
+    leftcorner[d] += leftpad[i]
+    rightcorner[d] -= rightpad[i]
+  end
+  center = ntuple(N) do i
+            leftcorner[i]:rightcorner[i]
+          end 
+  return outsize, center
+end
+
+function rrule(::typeof(pad_constant), x::AbstractArray, pad::NTuple{M,Int}, val=0; 
+              dims=1:M÷2) where M
+  szx = size(x)
+  y = pad_constant(x, pad, val; dims=dims)
+  
+  function pad_constant_pullback(Δ)
+    outsize, center = pad_outsize_and_center(szx, pad, dims)
+    (NO_FIELDS, 
+    @thunk(Δ[center...]), 
+    DoesNotExist(),
+    @thunk(sum(Δ) - sum(Δ[center...])), 
+    )
+  end
+  
+  return y, pad_constant_pullback
+end
+
+
+"""
+    pad_repeat(x, pad::Tuple; [dims])
+    pad_repeat(x, pad::Int; [dims])
+ 
+Pad the array `x` repeating the values on the border.
+
+`pad` can a tuple of integers `(l1, r1, ..., ln, rn)`
+of some length `2n` that specifies the left and right padding size
+for each of the dimensions in `dims`. If `dims` is not given, 
+it defaults to the first `n` dimensions.
+
+For integer `pad` input instead, it is applied on both sides
+on every dimension in `dims`. In this case, `dims` 
+defaults to the first `ndims(x)-2` dimensions 
+(i.e. excludes the channel and batch dimension). 
+
+See also [`pad_reflect`](@ref) and [`pad_constant`](@ref).
+
+```jldoctest
+julia> pad_repeat(reshape(1:9, 3, 3), (1,2,3,4))
+6×10 Matrix{Int64}:
+ 1  1  1  1  4  7  7  7  7  7
+ 1  1  1  1  4  7  7  7  7  7
+ 2  2  2  2  5  8  8  8  8  8
+ 3  3  3  3  6  9  9  9  9  9
+ 3  3  3  3  6  9  9  9  9  9
+ 3  3  3  3  6  9  9  9  9  9
+```
+"""
+function pad_repeat(x::AbstractArray, pad::NTuple{M,Int}; 
+                    dims=1:M÷2) where M
+  length(dims) == M÷2 ||
+    throw(ArgumentError("The number of dims should be equal to the number of padding dimensions"))
+  for (i, d) in enumerate(dims)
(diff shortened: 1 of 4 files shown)


DhairyaLGandhi:
How about calculating the size, filling in the val in the padded region and copying in the array? Sounds cleaner

CarloLucibello:
right, it wouldn't be AD friendly but the adjoint is easy to write, I'll do it

DhairyaLGandhi:
We can already pad arbitrary dimensions in Conv, so it would be better to add a setting to pad with a user defined value in there rather than an extra kernel.

CarloLucibello:
Added also pad_reflect and pad_repeat.

> We can already pad arbitrary dimensions in Conv, so it would be better to add a setting to pad with a user defined value in there rather than an extra kernel.

I tried a bit but didn't understand how to hook in there, I'm not even sure `conv` materializes a padded input. So I think merging with the conv machinery requires some major reengineering, probably not worth it, not stuff for this PR in any case

DhairyaLGandhi:
Will go through the PR, but before that so I understand it, is the intention to hook into the conv infrastructure (and reduce the API noise) before releasing any code?

CarloLucibello:
> Will go through the PR, but before that so I understand it, is the intention to hook into the conv infrastructure (and reduce the API noise) before releasing any code?

no. As I said

> I tried a bit but didn't understand how to hook in there, I'm not even sure conv materializes a padded input. So I think merging with the conv machinery requires some major reengineering, probably not worth it, not stuff for this PR in any case

Having these functions is of independent interest in any case

DhairyaLGandhi:
Since Conv is the biggest consumer of this by a wide margin, I would expect that such padding standards to be wrapped into the Conv machinery better. For example, I don't want to release a breaking version just to deprecate these methods.

darsnack:
You mean deprecate in the future? I think padding functions are useful beyond just consumption by conv.

Currently, the conv implementations don't actually materialize any of the padded regions. If we wanted to allow people to specify what gets padded in conv using these functions, then it would be a major overhaul of the entire convolution machinery. Either way, I don't see how that would be that these functions get deprecated. There isn't any functionality in the current conv machinery that could provide these functions instead. There might be tweaks to these functions that end up being breaking, but seeing as we would be drastically changing the conv implementations, we'd have a breaking change anyways.

DhairyaLGandhi:
> There might be tweaks to these functions that end up being breaking, but seeing as we would be drastically changing the conv implementations, we'd have a breaking change anyways.

Pretty sure we change internals without breaking api and behaviour frequently in the ecosystem, so 🤷

darsnack:
Yes, but the conv APIs are not designed to specify anything other than the padding dimension (assumes non-zero padding). To support consumption of such functions by conv, we would need to change either the `ConvDims` structs or function interfaces need to allow specifying the padding type. My first look assessment tells me that if anything is a breaking change, it is going to be the conv interfaces/structures. I don't think this code will need to change at all (certainly no need to deprecate).

DhairyaLGandhi:
> add a setting to pad with a user defined value in there rather than an extra kernel.

This isn't going to need any breaking change.

> To support consumption of such functions by conv, 

We explicitly don't need to support consumption of these functions here since the difference is really in how these dims are specified, which `conv` takes care of already.

To be clear adding to the API is non-breaking, removing from it is.

CarloLucibello:
Is this good to go?

DhairyaLGandhi:
Not yet, will take a look in the day

DhairyaLGandhi:
will get on this tomorrow, I have some important engagements come up

DhairyaLGandhi:
This seems to be touching a lot of unexpected things from the name of the PR, can those be factored out

CarloLucibello:
> This seems to be touching a lot of unexpected things from the name of the PR, can those be factored out

will do

Why are you unexporting? There is no reason why this set of functionality should be unexported while all the rest is exported

DhairyaLGandhi:
This shouldn't be allowed right, we are padding in the batch dimension? I guess we should allow the rest of the dimensions, but checking.

```julia
julia> NNlib.pad_constant(r, (1,2, 3, 4), dims = 3:4)
3×4×6×8 Array{Float32,4}:
[:, :, 1, 1] =
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
[...]
```

src/padding.jl
```diff
@@ -0,0 +1,205 @@
...
+  
+  leftcorner = ones(Int, N)
+  rightcorner = collect(outsize)
+  for (i, d) in enumerate(dims)
+    leftcorner[d] += leftpad[i]
+    rightcorner[d] -= rightpad[i]
+  end
+  center = ntuple(N) do i
+            leftcorner[i]:rightcorner[i]
+          end
```
DhairyaLGandhi, line 69: We could guarantee this function sees `length(pad) == 2 * N` and then do
```julia
center = broadcast((x,y) -> x .+ y, axes(x), pad[2:2:end])
```
or similar which would be cleaner and nice to guarantee we have padding information for all the dims in one place.

src/padding.jl
```diff
@@ -0,0 +1,205 @@
...
+6×10 Matrix{Int64}:
+ 1  1  1  1  4  7  7  7  7  7
+ 1  1  1  1  4  7  7  7  7  7
+ 2  2  2  2  5  8  8  8  8  8
+ 3  3  3  3  6  9  9  9  9  9
+ 3  3  3  3  6  9  9  9  9  9
+ 3  3  3  3  6  9  9  9  9  9
+```
+"""
+function pad_repeat(x::AbstractArray, pad::NTuple{M,Int};
```
DhairyaLGandhi, line 118: I'm inclined not to define this and instead just tell users to use `repeat`. Is there an advantage to defining it here?

src/conv.jl
```diff
@@ -211,11 +211,7 @@ for front_name in (:conv, :∇conv_data, :∇conv_filter,
 end
 
 for Dims in [:DenseConvDims, :DepthwiseConvDims, :PoolDims]
-    pullback = Symbol(Dims, :_pullback)
-    @eval function ChainRulesCore.rrule(::Type{$Dims}, args...; kwargs...)
-        $pullback(Δ) = (NO_FIELDS, ntuple(_ -> DoesNotExist(), length(args))...)
-        return $Dims(args...; kwargs...), $pullback
-    end
+    @eval @non_differentiable $Dims(args...)
```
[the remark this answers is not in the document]
CarloLucibello: done

src/padding.jl
CarloLucibello: I get the gist, but I don't think that code is going to work (besides the fact that there is no variable named `x`)
and I can't think of a way to make it right. Also, moving away from ntuple could cause type instability, I don't advise going in that direction

src/padding.jl
CarloLucibello: as you can see by yourself the code is not trivial, you can't tell users "use repeat". So yes, we should have that here. 
Another point is that sometimes we provide some functionality even if the implementation is simple (think about the `Dense` layer) because it is still convenient to have and makes for a unified and consistent experience.

src/padding.jl
DhairyaLGandhi: `x` is the array as is common practice in so much code. We can still make use of the `NTuple`, in fact, `pad` would be one.

DhairyaLGandhi:
```Julia
pad_constant(x::AbstractArray{T,N}, pad::Int, val = 0; kwargs...) where {T,N} =
  pad_constant(x, ntuple(_ -> pad, 2N), val)
pad_constant(x::AbstractArray{T,N}, pad::Tuple, val = 0; dims = :) where {T,N} =
  pad_constant(x, gen_pad(pad, dims, N), val)

function pad_idx(pad, dims, N)
  is = zip( (2 .* dims) .- 1, (2 .* dims))
end

function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
  if P == 2N
    return pad
  elseif P == D
   is = pad_idx(pad, dims, N)
   ps = zeros(Int, 2N)
   for (p,i) in zip(pad, is)
     ps[collect(i)] .= p
   end
   ntuple(i -> ps[i], 2N)
  end
end

# Expects length(pad) == 2M
function pad_constant(x::AbstractArray{T,M}, pad::NTuple{N,Int}, val = 0) where {T,M, N}
  sz, c = size_and_center(x, pad)
  res = fill!(similar(x, sz...), val)
  res[c..., ntuple(_ -> Colon(), M-2)...] = x
  res
end

# pad = (l, r, t, b)
function size_and_center(x, pad::NTuple{N,Int}) where N
  ds = 1:2:N
  dst = ntuple(i -> ds[i], N÷2)
  sz = map(i -> pad[i] + pad[i+1], dst) .+ size(x)
  center = broadcast((x,y) -> x .+ y, axes(x), pad[2:2:end]::NTuple{N÷2,Int})
  sz, center
end
```

Excluding the bit with `gen_pad`, which is not optimised at all (but not too bad either), but other than that, this runs 3-6x faster for small arrays, and discounting allocation times, faster for larger arrays as well (I assume because of inference but unsure).

DhairyaLGandhi: I think we can either replace it now, or I will PR it later

DhairyaLGandhi:
We should separate out the array building from the effect of the padding in the docstrings

test/padding.jl
```diff
@@ -0,0 +1,78 @@
...
+  y = @inferred pad_constant(x, (3, 2, 4, 5))
+  @test size(y) == (7, 11, 2)
+  @test y[4:5, 5:6, :] ≈ x
+  y[4:5, 5:6, :] .= 0
+  @test all(y .== 0)
+
+  @test pad_constant(x, (3, 2, 4, 5)) ≈ pad_zeros(x, (3, 2, 4, 5))
+  @test pad_zeros(x, 2) ≈ pad_zeros(x, (2,2)) 
+  
+  y = @inferred pad_constant(x, (3, 2, 4, 5), 1.2, dims=(1,3))
```
DhairyaLGandhi, line 13: I think `inferred` comes up as a problematic test often times, should be remove this?

CarloLucibello: I don't know of such issues. I know of packages using inferred so much that their tests become very annoying to read https://github.com/JuliaGraphs/LightGraphs.jl/blob/ce533729381a6c0b0d8a1eb5f0c3e058833cb55b/test/core.jl
I think it is ok to have inferred when you worry the function could be type-unstable. However, in most code you are confident enough in type stability that it's totally ok to not check it with inferred. Since I was in doubt about this particular function, I preferer to keep the `@inferred` check.

src/padding.jl
CarloLucibello: can you add it later? because your size_and_center doesn't support dims (e.g. `dims=(1,3)`)  and there are also a few things I don't understand

CarloLucibello:
nightly fail unrelated. This should be ready to go

DhairyaLGandhi:
I'm a bit uncomfortable creating so many copies by dealing with dims like in this PR; and since we have a better version for constant padding already available, I would like to switch to that instead anyway, so let's get the ball rolling on that.

src/padding.jl
DhairyaLGandhi: Its a different API entirely, and any "non-trivial"-ness in the implementation comes from forcing the same kwarg handling that has come up before. You definitely can tell users, and it would probably be better supported if you did.

With Dense there is the convenience factor, here the API call syntax might be different, but you can get a more complete implementation, which is clearly better since becoming a DSL is clearly the wrong way forward anyway.

PR #280, opened 2021-02-24
add gather

CarloLucibello:
easiest part of #255 

@yuehhua

### DIFF
--- a/src/gather.jl
+++ b/src/gather.jl
@@ -3,29 +3,24 @@ export gather, gather!
 """
     gather!(dst, src, idx)
 
-Reverse operation of scatter. Gathers data from source `src` 
-and writes it in destination `dst` according to the index
-array `idx`.
+Reverse operation of [`scatter!`](@ref). Gathers data from source `src` 
+and writes it in destination `dst` according to the index array `idx`.
 For each `k` in `CartesianIndices(idx)`, assign values to `dst` according to
 
     dst[:, ... , k] .= src[:, ... , idx[k]...]
 
 Notice that if `idx` is a vector containing integers,
-and both `dst` and `src` are matrices, previous 
-expression simplifies to
+and both `dst` and `src` are matrices, previous expression simplifies to
 
     dst[:, k] .= src[:, idx[k]]
 
-and `k` will range over `1:length(idx)`. 
+and `k` will run over `1:length(idx)`. 
 
-The elements of `idx` may be repeated. A single `src` column
-can end up being copied into zero, one, or multiple `dst` columns.
+The elements of `idx` can be integers or integer tuples and may be repeated. 
+A single `src` column can end up being copied into zero, one, 
+or multiple `dst` columns.
 
-# Arguments
-
-- `dst`: the destination where data would be assigned to.
-- `src`: the source of the data to be assigned.
-- `idx`: the mapping from source to destination.
+See [`gather`](@ref) for an allocating version.
 """
 function gather!(dst::AbstractArray{Tdst,Ndst}, 
                  src::AbstractArray{Tsrc,Nsrc}, 
@@ -33,22 +28,41 @@ function gather!(dst::AbstractArray{Tdst,Ndst},
                     {Tdst, Tsrc, Ndst, Nsrc, Nidx, Tidx <: IntOrIntTuple}
 
     M = typelength(Tidx)
-    Ndst - Nidx == Nsrc - M || throw(ArgumentError("Incompatible input shapes."))
-    size(dst)[1:Ndst-Nidx] == size(src)[1:Ndst-Nidx] || throw(ArgumentError("Incompatible input shapes."))
-    size(dst)[Ndst-Nidx+1:end] == size(idx) || throw(ArgumentError("Incompatible input shapes."))
+    d = Ndst - Nidx 
+    d == Nsrc - M || throw(ArgumentError("Incompatible input shapes."))
+    size(dst)[1:d] == size(src)[1:d] || throw(ArgumentError("Incompatible input shapes."))
+    size(dst)[d+1:end] == size(idx) || throw(ArgumentError("Incompatible input shapes."))
 
-    coldst = ntuple(i -> Colon(), Ndst - Nidx)
-    colsrc = ntuple(i -> Colon(), Nsrc - M)
+    colons = ntuple(i -> Colon(), d)
     for k in CartesianIndices(idx)
-        view(dst, coldst..., k) .= view(src, colsrc..., idx[k]...)
+        view(dst, colons..., k) .= view(src, colons..., idx[k]...)
     end
     return dst
 end
 
 """
-    gather(src, idx)
+    gather(src, idx) -> dst
 
-Non-mutating version of [`gather!`](@ref).
+Reverse operation of [`scatter`](@ref). Gathers data from source `src` 
+and writes it in a destination `dst` according to the index
+array `idx`.
+For each `k` in `CartesianIndices(idx)`, assign values to `dst` 
+according to
+
+    dst[:, ... , k] .= src[:, ... , idx[k]...]
+
+Notice that if `idx` is a vector containing integers
+and `src` is a matrix, previous expression simplifies to
+
+    dst[:, k] .= src[:, idx[k]]
+
+and `k` will run over `1:length(idx)`. 
+
+The elements of `idx` can be integers or integer tuples and may be repeated. 
+A single `src` column can end up being copied into zero, one, 
+or multiple `dst` columns.
+
+See [`gather!`](@ref) for an in-place version.
 """
 function gather(src::AbstractArray{Tsrc, Nsrc}, 
                 idx::AbstractArray{Tidx, Nidx}) where 
@@ -59,6 +73,3 @@ function gather(src::AbstractArray{Tsrc, Nsrc},
     dst = similar(src, Tsrc, dstsize)
     return gather!(dst, src, idx)
 end
-
-typelength(::Type{<:Number}) = 1
-typelength(::Type{<:NTuple{M}}) where M = M
--- a/src/gather.jl
+++ b/src/gather.jl
@@ -1,0 +1,64 @@
+export gather, gather!
+
+"""
+    gather!(dst, src, idx)
+
+Reverse operation of scatter. Gathers data from source `src` 
+and writes it in destination `dst` according to the index
+array `idx`.
+For each `k` in `CartesianIndices(idx)`, assign values to `dst` according to
+
+    dst[:, ... , k] .= src[:, ... , idx[k]...]
+
+Notice that if `idx` is a vector containing integers,
+and both `dst` and `src` are matrices, previous 
+expression simplifies to
+
+    dst[:, k] .= src[:, idx[k]]
+
+and `k` will range over `1:length(idx)`. 
+
+The elements of `idx` may be repeated. A single `src` column
+can end up being copied into zero, one, or multiple `dst` columns.
+
+# Arguments
+
+- `dst`: the destination where data would be assigned to.
+- `src`: the source of the data to be assigned.
+- `idx`: the mapping from source to destination.
+"""
+function gather!(dst::AbstractArray{Tdst,Ndst}, 
+                 src::AbstractArray{Tsrc,Nsrc}, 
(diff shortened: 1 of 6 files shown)


src/gather.jl
```diff
@@ -0,0 +1,82 @@
...
+  # TODO: use M = _length(eltype(idx)) to merge the integer method into this?
+  @boundscheck _gather_checkbounds(src, idx)
+  Ndst - Nidx == Nsrc - M  || throw(ArgumentError(""))
+  size(dst)[1:Ndst-Nidx] ==  size(src)[1:Ndst-Nidx] || throw(ArgumentError(""))
+  size(dst)[Ndst-Nidx+1:end] ==  size(idx) || throw(ArgumentError(""))
+  
+  coldst = ntuple(i -> Colon(), Ndst - Nidx)
+  colsrc = ntuple(i -> Colon(), Nsrc - M)
+  @simd for k in CartesianIndices(idx)
+    @inbounds view(dst, coldst..., Tuple(k)...) .= view(src, colsrc..., idx[k]...)
```
DhairyaLGandhi, line 28: We usually don't use performance annotations in FluxML

src/gather.jl
```diff
@@ -0,0 +1,82 @@
...
+# Arguments
+- `dst`: the destination where data would be assigned to.
+- `src`: the source data to be assigned.
+- `idx`: the mapping for assignment from source to destination.
+"""
+function gather!(dst::AbstractArray{Tdst,Ndst}, 
+                 src::AbstractArray{Tsrc,Nsrc}, 
+                 idx::AbstractArray{NTuple{M,Int}, Nidx}) where {Tdst,Tsrc,Ndst,Nsrc,Nidx,M}
+  # TODO: use M = _length(eltype(idx)) to merge the integer method into this?
+  @boundscheck _gather_checkbounds(src, idx)
```
DhairyaLGandhi, line 20: It would error if it's out of bounds anyway right?

src/gather.jl
CarloLucibello: we do now

src/gather.jl
CarloLucibello: no because the loop is marked as inbounds

DhairyaLGandhi: Yes and it shouldnt

src/gather.jl
DhairyaLGandhi: There's typically no difference in ml routines, so we still don't

yuehhua:
@chengchingwen You may want to check it.

src/gather.jl
CarloLucibello: just to give some context, this is a port (+generalization) of code from https://github.com/yuehhua/ScatterNNlib.jl
that has been around for a while, it's battle-tested, perfomant, and has been benchmarked against similar frameworks. 
The port will be time-consuming and nontrivial, so please let's not make it more painful than needed and let's not undo other people's work.
Bounds checking outside sensitive loops and going inbounds inside is perfectly reasonable, done all over the place in Base, and it's here for good reason.
So the answer to "We usually don't use performance annotations in FluxML" is "So what?" (and a secondary answer is that's the reason we are slow and using loopvectorization or similars has been discussed and will happen at some point).
Please make focused and clear comments, like
"we shouldn't be doing *a specific thing* because *a specific issue* could arise in a *specific situation*" 
and keep in mind the context I just mentioned, otherwise, this is going to be a nightmare.

yuehhua: For performance annotations, @DhairyaLGandhi may want to check it [here](https://github.com/FluxML/NNlib.jl/pull/255#issuecomment-787475179). I have done a small benchmark on `scatter!` w/o annotations.

DhairyaLGandhi: Cool, seems like it's not doing much in terms of uptick. Let's leave it out. For folks who need it, they can add those to the function call anyway, and doing it here leaves us open to cases where custom arrays don't check bounds appropriately (or array of structs which may not simd cleanly) for one reason or another. Re vectorisation etc: we would rather consult experts who can give us an idea of better memory accesses etc and that work is ongoing, so let's not jump the gun.

CarloLucibello: I tested without annotations as well in this pr and the perfomance impact is negligible, so I remove them

CarloLucibello:
this should be ready to go

CarloLucibello:
I'm going to merge this as soon as tests pass so that @yuehhua can continue his work (adding AD rules, adding cuda support).
I'll open a PR for a final api discussion on both gather and scatter

DhairyaLGandhi:
I'd expect to do a review of the implementation before merging these operations into NNlib.

CarloLucibello:
> I'd expect to do a review of the implementation before merging these operations into NNlib.

I'll add fake comments in the new pr so that the implementation can be easily reviewed

DhairyaLGandhi:
As I see it, the implementation is not following the TF semantics, is that right?


I'd rather have reviews before merging.

DhairyaLGandhi:
There seems to be a lot of allocating going on, we could alternatively let Julia figure out the indexing for us;

```julia
function gather(arr, inds)
  s = size(arr, 1)
  g1 = similar(arr, 2s, size(inds, 1))
  g2 = similar(g1)
  for k ∈ 1:2, j ∈ axes(inds,1)
    ii = inds[j, k]
    o1 = (k-1)*s
    o2 = (2-k)*s
    for i ∈ axes(arr,1)
      g = arr[i,ii]
      g1[i + o1, j] = g
      g2[i + o2, j] = g
    end
  end
  g1, g2
end
```

So this version would not allocate quite as much, and we can use column copies instead of row copies which would be more Julian.

Also leads to a sizeable perf difference for small cases (and maybe for large ones). This is only generating the first half, but we can make it do the other as well of course.

```Julia
julia> src = [3, 4, 5, 6, 7];

julia> index = [1 2 3 4;
                    4 2 1 3;
                    3 5 5 3][:,:,1:1]

julia> @btime gather($(src), $index);
  2.905 μs (37 allocations: 1.11 KiB)

# this one
julia> @btime gather($(src'), $index);
  131.193 ns (2 allocations: 256 bytes)
```

CarloLucibello:
> As I see it, the implementation is not following the TF semantics, is that right?

I think it does follow TF semantics
https://www.tensorflow.org/api_docs/python/tf/gather
although our implementation is less general (we just have their defaults for batch_dims and axis) 

> I'd rather have reviews before merging.

yeah, ok, but as usual... please be quicker and avoid blocking other people work all over the place

CarloLucibello:
I don't understand what your implementation is doing (and why it is returning 2 things).
and I cannot reproduce you benchmark
```julia
julia> function gather2(arr, inds)
  s = size(arr, 1)
  g1 = similar(arr, 2s, size(inds, 1))
  g2 = similar(g1)
  for k ∈ 1:2, j ∈ axes(inds,1)
    ii = inds[j, k]
    o1 = (k-1)*s
    o2 = (2-k)*s
    for i ∈ axes(arr,1)
      g = arr[i,ii]
      g1[i + o1, j] = g
      g2[i + o2, j] = g
    end
  end
  g1, g2
end

julia> src = [3, 4, 5, 6, 7];

julia> index = [1 2 3 4;
                    4 2 1 3;
                    3 5 5 3][:,:,1:1];

julia> @btime gather($(src), $index);
  60.446 ns (1 allocation: 176 bytes)

julia> @btime gather2($(src), $index);
ERROR: BoundsError: attempt to access 5-element Vector{Int64} at index [1, 4]
Stacktrace:
  [1] getindex
    @ ./array.jl:802 [inlined]
  [2] gather2(arr::Vector{Int64}, inds::Array{Int64, 3})
    @ Main ./REPL[3]:10
  [3] var"##core#313"(src#310::Vector{Int64}, index#311::Array{Int64, 3})
    @ Main ~/.julia/packages/BenchmarkTools/eCEpo/src/execution.jl:371
```
I see just 1 allocation for the implementation, are you benchmarking the right thing?
Also, your code doesn't appear to work on the general case of d-dimensional src and indx.
If you have anything working and actionable in the short term I can try to add it to this pr, otherwise performance optimization can come later.

src/gather.jl
```diff
@@ -0,0 +1,64 @@
...
+    for k in CartesianIndices(idx)
+        view(dst, coldst..., k) .= view(src, colsrc..., idx[k]...)
+    end
+    return dst
+end
+
+"""
+    gather(src, idx)
+
+Non-mutating version of [`gather!`](@ref).
```
DhairyaLGandhi, line 51: I would move the docstrings over to here; the non allocating version is usually an implementation detail and has a fairly standard practice of allowing the allocated array as the first arg.

DhairyaLGandhi:
I mentioned in the snippet its for the matrix case, but can easily be extended. Yes - benchmarking the correct thing. Its returning two things because that's how I had needed it for a different operation - sorry should have removed it.

You'd notice you need to transpose `src` for this version.

CarloLucibello:
so I used the transpose, and on previous example get this
```julia
julia> gather(src,index)
3×4×1 Array{Int64, 3}:
[:, :, 1] =
 3  4  5  6
 6  4  3  5
 5  7  7  5

julia> gather2(src',index)
([3 6 5; 4 4 7], [4 4 7; 3 6 5])

julia> @btime gather($(src), $index);
  60.218 ns (1 allocation: 176 bytes)

julia> @btime gather2($(src'), $index);
  68.138 ns (2 allocations: 256 bytes)
``

File: src/gather.jl (hunk 3 above)
Comment (CarloLucibello) [reply src/gather.jl]: done

**Comment by DhairyaLGandhi** [comment]
For the general case - Julia indexing (it might be better for longer term use to benefit from Base improvements if the perf isn't off which I don't think it is)

```julia
function gatherweave2(arr, inds) where {T,N}
  g = arr[:, inds]
  v = reshape(g, length(g),1) # vec(g)
  reshape(v, size(inds)...)
end
```

Huh, interesting, I definitely can't reproduce the 60 ns number - what Julia version are you on? I think the approach is fine. Be good to remove the `IntOrIntTuple` type, though.

CarloLucibello:
> Huh, interesting, I definitely can't reproduce the 60 ns number - what Julia version are you on? More generally - I think this approach is fine

yeah it's julia 1.6, probably they optimized the views.

> Be good to remove the IntOrIntTuple type, though.

it's used in a few places, would be inconvenient to type the union each time 

> For the general case - Julia indexing (it might be better for longer term use to benefit from Base improvements if the perf isn't off which I don't think it is)

I don't know why you have those reshape there, but I can try to use indexing in `gather` in the case `idx` is an array of integers instead of falling back to `gather!`.

CarloLucibello:
On the usual small example indexing and current gather perform the same
```julia
julia> function gather3(src::AbstractArray{Tsrc, Nsrc}, 
                 idx::AbstractArray{<:Integer}) where {Tsrc, Nsrc}
    colons = ntuple(i -> Colon(), Nsrc-1)
    return src[colons..., idx]
end

julia> @btime gather3($(src), $index);
  54.328 ns (1 allocation: 176 bytes)

julia> @btime gather($(src), $index);
  54.847 ns (1 allocation: 176 bytes)
```
I did a small tweak to `gather` that shave off a few nanoseconds.

DhairyaLGandhi:
We could potentially allow Cartesian indices in `idx` as well.

CarloLucibello:
> We could potentially allow Cartesian indices in idx as well.

we could. It would mean to clutter the code a bit though since you can't splat `idx[k]...` for cartesian indexes. I can't think o a concrete use case for it, so maybe we should postpone. 

Looking at the tensorflow implementation,
https://www.tensorflow.org/api_docs/python/tf/gather
the main missing things seem to be the `axis` and `batch_dims` keywords. Those are not hard to implement, but we already have anything we need for FluxGeometric and more. Since a few pieces, AD and gpu, still have to fall in place, I'd say let's merge this PR as it is, then do AD and gpu, and once everything has landed, look at the big picture and generalize

CarloLucibello:
bump

CarloLucibello:
@yuehhua I'm merging this. Could you file a PR with the AD rules and another one with the gpu kernels?
The gpu kernels can go in the newly created NNlibCUDA.jl in this repo (see  #286)

DhairyaLGandhi:
From an api pov, that's correct. But if we generalise it to use Cartesian, we can accept both and have a common implementation internally which would be cleaner too.

yuehhua:
Oh! Sure!😊 We can do that!

PR #283, opened 2021-03-01
Faster constant padding

DhairyaLGandhi:
We should clean up the dimension handling - which needs the extra `gen_pad` which uses arrays in some cases. We can get rid of those, I think.

```Julia
julia> @btime NNlib.pad_constant($r, (2,2,2,2,0,0,0,0));
  558.871 ns (5 allocations: 1.06 KiB)

# master
julia> @btime NNlib.pad_constant($r, (2,2,2,2,0,0,0,0));
  1.644 μs (29 allocations: 2.45 KiB)
```

### DIFF
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -7,7 +7,7 @@ export pad_constant, pad_repeat, pad_reflect, pad_zeros
 Pad the array `x` with zeros.
 Equivalent to [`pad_constant`](@ref) with the constant equal to 0. 
 """
-pad_zeros(x::AbstractArray, pad; dims = :) where M =
+pad_zeros(x::AbstractArray, pad; dims = :) =
   pad_constant(x, pad, 0; dims = dims)
 
 """
@@ -180,7 +180,6 @@ function pad_repeat(x::AbstractArray{F,N}, pad::NTuple{2,Int};
 
   return cat(xl, x, xr, dims = dims)
 end
-
 
 """
     pad_reflect(x, pad::Tuple; [dims])
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -55,6 +55,7 @@ end
 
 gen_pad(pad::Int, dims, N) = gen_pad(ntuple(_ -> pad, length(dims)), dims, N)
 gen_pad(pad::Int, dims::Colon, N) = ntuple(_ -> pad, 2N)
+gen_pad(pad, dims::Colon, N) = gen_pad(pad, ntuple(identity, N), N)
 function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
   if P == 2N
     return pad
@@ -68,8 +69,9 @@ function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
   elseif P == 2D
     is = pad_idx(pad, dims, N)
     ps = zeros(Int, 2N)
-    for (i,x) in enumerate(is)
-      ps[collect(x)] .= pad[collect(x)]
+    for x in is
+      i = collect(x)
+      @views ps[i] .= pad[i]
     end
     ntuple(i -> ps[i], 2N)
   else
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -69,7 +69,7 @@ function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
     is = pad_idx(pad, dims, N)
     ps = zeros(Int, 2N)
     for (i,x) in enumerate(is)
-      ps[collect(x)] .= pad[[i, i+1]]
+      ps[collect(x)] .= pad[collect(x)]
     end
     ntuple(i -> ps[i], 2N)
   else
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -45,7 +45,7 @@ julia> pad_constant(r, (1, 2, 3, 4), 8)
 ```
 """
 pad_constant(x::AbstractArray{T,N}, pad::Int, val = 0; dims = :) where {T,N} =
-  pad_constant(x, ntuple(_ -> pad, 2N), val)
+  pad_constant(x, gen_pad(pad, dims, N), val)
 pad_constant(x::AbstractArray{T,N}, pad::Tuple, val = 0; dims = :) where {T,N} =
   pad_constant(x, gen_pad(pad, dims, N), val)
 
@@ -54,6 +54,7 @@ function pad_idx(pad, dims, N)
 end
 
 gen_pad(pad::Int, dims, N) = gen_pad(ntuple(_ -> pad, length(dims)), dims, N)
+gen_pad(pad::Int, dims::Colon, N) = ntuple(_ -> pad, 2N)
 function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
   if P == 2N
     return pad
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -71,6 +71,8 @@ function gen_pad(pad::NTuple{P,Int}, dims::NTuple{D,Int}, N) where {D,P}
       ps[collect(x)] .= pad[[i, i+1]]
     end
     ntuple(i -> ps[i], 2N)
+  else
+    throw(ArgumentError("Passed padding $pad and dims $dims could not be parsed."))
   end
 end
 
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -7,7 +7,7 @@ export pad_constant, pad_repeat, pad_reflect, pad_zeros
 Pad the array `x` with zeros.
 Equivalent to [`pad_constant`](@ref) with the constant equal to 0. 
 """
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
I have also fixed how dims are handled compared to master

CarloLucibello:
closing and reopening since didn't trigger CI

CarloLucibello:
weird, what happened to github's ci?

CarloLucibello:
btw, I saw you introduced on master 2 space indentation for this file, but that is wrong, NNlib uses 4 spaces indents

DhairyaLGandhi:
We just have a bad GA CI script

DhairyaLGandhi:
This run is a bit meaningless anyway, since I've switched around the dims.

DhairyaLGandhi:
So API wise I was thinking we should have something that can work with the Conv machinery cleanly.

The idea might be to generate a closure that can handle the padding work

`Conv(..., pad = pad_constant((3,4,5)))` for example.

DhairyaLGandhi:
Why are we testing with threads separately?

DhairyaLGandhi:
Barring docs change for the dims rearrangement, this is done. We should change the implementation of the rest as well to follow this behaviour.

CarloLucibello:
could you add the docstring first? it's not clear to me how one is supposed to use this to do things like
```julia
# current master
x = rand(2,2,2,2)
pad_constant(x, (1,2), dims=2) # asymmetric padding in 2nd dimension

pad_constant(x, 2, dims=1:3) # symmetric padding in first 3 dims
```

DhairyaLGandhi:
Just convert the `dims` to tuples. Pass `dims = (1,)` and `dims = (1,2,3)`

CarloLucibello:
> Just convert the dims to tuples. Pass dims = (1,) and dims = (1,2,3)

`dims` keyword should handle integer and ranges, this the common behaviour and what people expect. Shouldn't be hard to fix this.

Besides that, following your suggestion, I see two different errors:
```julia
julia> n=2; x = rand(n,n,n,n);

julia> pad_constant(x, (1,2), dims=(2,)) # left and right inverted 
2×5×2×2 Array{Float64, 4}:
[:, :, 1, 1] =
 0.0  0.0  0.391147  0.459926  0.0
 0.0  0.0  0.758401  0.353851  0.0

[:, :, 2, 1] =
 0.0  0.0  0.291629  0.708728  0.0
 0.0  0.0  0.535743  0.566279  0.0

[:, :, 1, 2] =
 0.0  0.0  0.761029  0.727371   0.0
 0.0  0.0  0.158018  0.0933968  0.0

[:, :, 2, 2] =
 0.0  0.0  0.511264  0.548279  0.0
 0.0  0.0  0.287906  0.95368   0.0

julia> pad_constant(x, (1,2), dims=(1,2,3))
ERROR: ArgumentError: Could not parse padding (1, 2) and dims (1, 2, 3)
Stacktrace:
 [1] #251
   @ ~/.julia/dev/NNlib/src/padding.jl:83 [inlined]
 [2] ntuple
   @ ./ntuple.jl:19 [inlined]
 [3] gen_pad(pad::Tuple{Int64, Int64}, dims::Tuple{Int64, Int64, Int64}, N::Int64)
   @ NNlib ~/.julia/dev/NNlib/src/padding.jl:74
 [4] pad_constant(x::Array{Float64, 4}, pad::Tuple{Int64, Int64}, val::Int64; dims::Tuple{Int64, Int64, Int64})
   @ NNlib ~/.julia/dev/NNlib/src/padding.jl:49
 [5] top-level scope
   @ REPL[15]:1
```

DhairyaLGandhi:
> left and right inverted 

Yes that is a choice, basically to say the first of the tuple follows the Julia convention of growing the array in the direction of the "memory layout".

> pad_constant(x, (1,2), dims=(1,2,3))

I picked this syntax to mean "homogenous padding with the corresponding dimension", so you need to add a third element to the padding. Else, the case of `pad_constant(x, (1,2), dims=(1,2))` is ambiguous. Basically, should this expand to `(1,2,1,2,...)` or `(1,1,2,2,...)`. You should always pass a tuple of `2*length(dims)` for the assymetric case and reserve contracted cases for symmetric ones.

src/padding.jl
```diff
@@ -7,7 +7,7 @@ export pad_constant, pad_repeat, pad_reflect, pad_zeros
 Pad the array `x` with zeros.
 Equivalent to [`pad_constant`](@ref) with the constant equal to 0. 
 """
-pad_zeros(x::AbstractArray, pad::NTuple{M,Int}; dims = 1:M÷2) where M =
+pad_zeros(x::AbstractArray, pad; dims = :) where M =
```
CarloLucibello, line 10: There is an extra `where M`.
Also, when doing `pad_zeros(x, 2)`, I agree that in a generic context `dims = :` would be the correct default, but for NN `dims = 1:ndims(x)-3` is the one that makes most sense and it is going to correspond to what the flux's layer `PadZeros(2)`  will do. I don't dislike `:`, just saying we should be careful and consider the NN usage and interaction with Flux

CarloLucibello:
> > left and right inverted
> 
> Yes that is a choice, basically to say the first of the tuple follows the Julia convention of growing the array in the direction of the "memory layout".

can we revert this change? user side feels unnatural.

> 
> > pad_constant(x, (1,2), dims=(1,2,3))
> 
> I picked this syntax to mean "homogenous padding with the corresponding dimension", so you need to add a third element to the padding. Else, the case of `pad_constant(x, (1,2), dims=(1,2))` is ambiguous. Basically, should this expand to `(1,2,1,2,...)` or `(1,1,2,2,...)`. You should always pass a tuple of `2*length(dims)` for the assymetric case and reserve contracted cases for symmetric ones.

ok that is fine, maybe we could add a method taking two integers for that behavior: `pad_constant(x, 1, 2, dims=(1,2))`.

When tests are ready (possibly performing only the minimal changing over the current ones, so that it becomes clear if we are loosing features and it's easier to review) I'll do another review pass

src/padding.jl
DhairyaLGandhi: I wouldn't want an extra Pad* layer. There's no value to it.

DhairyaLGandhi:
We aren't losing anything, besides, the code was merged only to serve as a starting point. Perhaps we shouldn't have, oh well. I think enforcing unambiguous padding is a win and the tests haven't changed beyond fixing the API, and I have added tests for the stuff we have added here.

darsnack:
>>left and right inverted
>
>Yes that is a choice, basically to say the first of the tuple follows the Julia convention of growing the array in the direction of the "memory layout".

Can you explain more what was meant by this? Looking at a simple test below, I don't think the current behavior matches the memory layout. You can see that the _second_ padding value (2) was applied _first_ in the memory layout. I would also have expected the result to be the inverse of the current behavior, based both on user intuition and on memory layout.

```julia
julia> n=2; x = rand(n,n,n,n);

julia> xpad = pad_constant(x, (1,2), dims=(2,));

julia> vec(x)
16-element Array{Float64,1}:
 0.02489842663461883
 0.7423910882092584
 0.5848037448338126
 0.9244979370952127
 0.034271186055481406
 0.797783910306795
 0.43206249616027903
 ⋮
 0.23221892468913685
 0.11757916898318221
 0.9893431957475107
 0.5060607349925461
 0.6393465312024416
 0.4779310825890677
 0.3006844111512579

julia> vec(xpad)
40-element Array{Float64,1}:
 0.0
 0.0
 0.0
 0.0
 0.02489842663461883
 0.7423910882092584
 0.5848037448338126
 ⋮
 0.0
 0.5060607349925461
 0.6393465312024416
 0.4779310825890677
 0.3006844111512579
 0.0
 0.0
```

darsnack:
> `pad_constant(x, (1,2), dims=(1,2,3))`

What was the old behavior for this? I feel like either a single value or a tuple of equal length to `dims` makes most sense to me.

CarloLucibello:
@DhairyaLGandhi why did you delete your comment? Anyways, those considerations about memory layouts are not appropriate (also besides what @darsnack pointed out): this is an abstraction, you think about what goes on the left and what on the right, or before the start and after the end, inverting the two  arguments it's just weird

DhairyaLGandhi:
I didn't delete any comment...

The layout is that you first grow the dimensions in their usual direction first.

src/padding.jl
```diff
@@ -44,53 +44,77 @@ julia> pad_constant(r, (1, 2, 3, 4), 8)
...
-function rrule(::typeof(pad_constant), x::AbstractArray,
-               pad::NTuple{M,Int}, val=0; 
-               dims=1:M÷2) where M
-  szx = size(x)
-  y = pad_constant(x, pad, val; dims=dims)
-  
+
+# Expects length(pad) == 2M
+function pad_constant(x::AbstractArray{T,M}, pad::NTuple{N,Tuple{Int,Int}}, val = 0) where {T,M,N}
+  p = pad_zeros_tuple(tuplejoin(pad...), M)
```
DhairyaLGandhi, line 95:
```suggestion
  p = pad_zeros_tuple(tuplejoin(reverse.(pad)...), M)
```

PR #284, opened 2021-03-02
update ci.yml

CarloLucibello:
small cleanup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         arch:
           - x64
         julia-threads:
-          - "1"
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Why do you not want to run ci on branches, seems odd

CarloLucibello:
the problem is that if you have ci on both `branches` and `pull requests`, every time you create a pull request from a branch in this repo you trigger the CI twice. So `branches` should be restricted to master only. We could also get rid of `branches`  entirely since supposedly any change on master goes through a PR so it's already tested there, but a little redundant ci doesn't hurt either

PR #291, opened 2021-03-12
Fix batched_mul gradient, for cases with one matrix

### LINKED ISSUE #290: batched_mul(A, B) not differentiable when B is an Array{T, 2}
MWE:

```julia
using NNlib, Zygote

W = rand(Float32, 1, 1, 2)
x = rand(Float32, 1, 1) 

gz, back = Zygote.pullback(x -> NNlib.batched_mul(W, x), x)
back(gz)[1]
```

```julia
ERROR: DimensionMismatch("new dimensions (1, 1) must be consistent with array size 2")
Stacktrace:
 [1] (::Base.var"#throw_dmrsa#213")(::Tuple{Int64,Int64}, ::Int64) at ./reshapedarray.jl:41
 [2] reshape at ./reshapedarray.jl:45 [inlined]
 [3] #434 at /home/guillaume/.julia/packages/Zygote/KpME9/src/lib/array.jl:94 [inlined]
 [4] #2417#back at /home/guillaume/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59 [inlined]
 [5] #151 at /home/guillaume/.julia/packages/Zygote/KpME9/src/lib/lib.jl:191 [inlined]
 [6] #1694#back at /home/guillaume/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59 [inlined]
 [7] _semi_batched_mul at /home/guillaume/.julia/packages/NNlib/uX1eA/src/batched/batchedmul.jl:140 [inlined]
 [8] batched_mul at /home/guillaume/.julia/packages/NNlib/uX1eA/src/batched/batchedmul.jl:136 [inlined]
 [9] #5 at ./REPL[6]:1 [inlined]
 [10] (::Zygote.var"#41#42"{typeof(∂(#5))})(::Array{Float32,3}) at /home/guillaume/.julia/packages/Zygote/KpME9/src/compiler/interface.jl:40
 [11] top-level scope at REPL[7]:1
```

It works on Zygote v0.5.17 instead of v0.6.3 so maybe it's on them though, but you had to add:

`NNlib.batched_adjoint(A::AbstractMatrix) = NNlib.batched_adjoint(reshape(A, size(A)..., 1))`

mcabbott:
Closes #290

### DIFF
--- a/src/batched/batchedadjtrans.jl
+++ b/src/batched/batchedadjtrans.jl
@@ -50,15 +50,6 @@ batched_transpose(A::BatchedAdjoint{<:Complex}) = BatchedAdjoint(BatchedTranspos
 BatchedAdjoint(A) = BatchedAdjoint{Base.promote_op(adjoint,eltype(A)),typeof(A)}(A)
 BatchedTranspose(A) = BatchedTranspose{Base.promote_op(transpose,eltype(A)),typeof(A)}(A)
 
-# Action on matrices -- needed in gradient of 3-array ⊠ matrix
-batched_transpose(A::AbstractMatrix) = batched_transpose(reshape(A, size(A)..., 1))
-batched_transpose(A::Transpose{<:Number,<:AbstractMatrix}) = reshape(parent(A), size(parent(A))..., 1)
-batched_transpose(A::Adjoint{<:Real,<:AbstractMatrix}) = reshape(parent(A), size(parent(A))..., 1)
-
-batched_adjoint(A::AbstractMatrix) = batched_adjoint(reshape(A, size(A)..., 1))
-batched_adjoint(A::Transpose{<:Real,<:AbstractMatrix}) = reshape(parent(A), size(parent(A))..., 1)
-batched_adjoint(A::Adjoint{<:Number,<:AbstractMatrix}) = reshape(parent(A), size(parent(A))..., 1)
-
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
Erroring in the right place would avoid the asserts. It also looks like we can lift some operations out of the conditionals, I think that would be good

mcabbott:
> What's the current behavior of batched_mul with matrices?

`batched_mul` should never accept two matrices. But either A or B can be a matrix, if only the other one has a batch dimension. So the `@assert`s are genuine ones, in that they should be impossible to trigger, you will get a MethodError first, I believe. 

But `size(A,3)==1` is treated much the same as `ndims(A)==2`, so it is possible this fix can be simplified a bit.

DhairyaLGandhi:
Right, catching it in the method error is preferable, and thereby omitting the asserts. I would imagine reducing the batched_mul definition to a broadcasted mul would go a long way to ensuring base can do the heavy lifting for us. There's prolly no performance penalty either

mcabbott:
Well that's definitely out of scope, here. You can time it if curious but CuBLAS's batched mul is pretty quick.

And, again, the `@asserts` were there as a logic check. If you can trigger them, then the author's mental model of the rest of the code is wrong, not just the arguments. This is, I believe, the correct usage.

Issue #294, opened 2021-03-14
`pad_constant` fails for vectors


torfjelde:
## Problem

At the moment the following snippet fails

```julia
u = randn(10);
NNlib.pad_zeros(u, (2, 2))
```

with the error

```julia
ArgumentError: tuple length should be ≥ 0, got -1

Stacktrace:
 [1] macro expansion at ./ntuple.jl:47 [inlined]
 [2] ntuple at ./ntuple.jl:45 [inlined]
 [3] pad_constant(::Array{Float64,1}, ::Tuple{Tuple{Int64,Int64}}, ::Int64) at ./In[103]:6
 [4] pad_constant(::Array{Float64,1}, ::Tuple{Int64,Int64}, ::Int64; dims::Function) at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:103
 [5] pad_zeros(::Array{Float64,1}, ::Tuple{Int64,Int64}; dims::Function) at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:10
 [6] pad_zeros(::Array{Float64,1}, ::Tuple{Int64,Int64}) at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:10
```

It seems to me like there's not a particular reason why this ought to not be supported?

## Solution

AFAIK adding the following `if` statement ought to do the trick and shouldn't have any overhead:

```julia
function pad_constant(x::AbstractArray{T,M}, pad::NTuple{N,Tuple{Int,Int}}, val = 0) where {T,M,N}
    p = pad_zeros_tuple(tuplejoin(reverse.(pad)...), M)
    sz, c = size_and_center(x, p)
    res = fill!(similar(x, sz...), val)
    if M > 1
        res[c..., ntuple(_ -> Colon(), Val(M - 2))...] = x
    else
        res[c...] = x
    end
    res
end
```

DhairyaLGandhi:
I think we could generically drop the `ntuple`. Could you check?

torfjelde:
Ah yeah; `c` is the "centered" indices correspondng to the location of `x`, right? So should be sufficient to just do `res[c...]`. 
And yeah, some quick examples run without issues.

DhairyaLGandhi:
Correct, would you be up to PR the changes (tests) for this and the other issue?

PR #295, opened 2021-03-14
Fixes #293 and #294

### LINKED ISSUE #293: rrule for `pad_constant` fails when using (l1, r1, ..., ln, rn) padding
## Problem
When executing the following snippet:

```julia
using Zygote, NNlib

upad, ∂upad = Zygote.pullback(u -> NNlib.pad_zeros(u, (0, 1, 0, 1)), u)
upad # <= everything looks correct
∂upad(ones(size(upad)))
```

```julia
MethodError: no method matching size_and_center(::Array{Float64,2}, ::Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64}})
Closest candidates are:
  size_and_center(::Any, ::Tuple{Vararg{Int64,N}}) where N at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:157

Stacktrace:
 [1] (::NNlib.var"#pad_constant_pullback#263"{2,Colon,Array{Float64,2},NTuple{4,Int64}})(::Array{Float64,2}) at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:170
 [2] ZBack at /home/tor/.julia/packages/Zygote/KpME9/src/compiler/chainrules.jl:77 [inlined]
 [3] (::Zygote.var"#kw_zpullback#40"{NNlib.var"#pad_constant_pullback#263"{2,Colon,Array{Float64,2},NTuple{4,Int64}}})(::Array{Float64,2}) at /home/tor/.julia/packages/Zygote/KpME9/src/compiler/chainrules.jl:103
 [4] #pad_zeros#242 at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:10 [inlined]
 [5] pad_zeros at /home/tor/.julia/packages/NNlib/uX1eA/src/padding.jl:10 [inlined]
 [6] #65 at ./In[62]:1 [inlined]
 [7] (::Zygote.var"#41#42"{typeof(∂(#65))})(::Array{Float64,2}) at /home/tor/.julia/packages/Zygote/KpME9/src/compiler/interface.jl:40
 [8] top-level scope at In[64]:1
 [9] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

Seems like the issue is the line https://github.com/FluxML/NNlib.jl/blob/master/src/padding.jl#L170 which calls `size_and_center` which doesn't support `NTuple{N, NTuple{2, Int}}` despite `gen_pad` seeming only returning those formats https://github.com/FluxML/NNlib.jl/blob/master/src/padding.jl#L128-L145?

## Solution
Seems like adding the following implementation of `size_and_center` should fix it:

```julia
function size_and_center(x, pad::NTuple{N,NTuple{2, Int}}) where N
  sz = ntuple(i -> pad[i][1] + pad[i][2], N) .+ size(x)
  center = broadcast((x,y) -> x .+ y, axes(x), ntuple(i -> pad[i][2], N))
  sz, center
end
```

I can make a PR if this seems reasonable.

torfjelde:
This PR fixes #293 and #294, and in the process moves to using only `((l1, r1), ..., (ln, rn))` representation of padding internally instead of both that and `(l1, r1, ..., ln, rn)`. This means that we can simplify the codebase a bit.

### DIFF
--- a/test/padding.jl
+++ b/test/padding.jl
@@ -42,6 +42,7 @@
 
   gradtest(x -> pad_constant(x, 2), rand(2,2,2))
   gradtest(x -> pad_constant(x, (2, 1, 1, 2)), rand(2,2))
+  gradtest(x -> pad_constant(x, (2, 1,)), rand(2))
 end
 
 @testset "padding repeat" begin
--- a/test/padding.jl
+++ b/test/padding.jl
@@ -12,6 +12,9 @@
   p = NNlib.gen_pad(1, (1,2,3), 4)
   @test p == ((1, 1), (1, 1), (1, 1), (0, 0))
 
+  p = NNlib.gen_pad(3, :, 2)
+  @test p == ((3, 3), (3, 3))
+
   y = pad_constant(x, (3, 2, 4))
   @test size(y) == (8, 6, 10)
   @test y[4:5, 3:4, 5:6] ≈ x
@@ -35,7 +38,10 @@
 
   @test size(pad_constant(x, 1, dims = 1)) == (4,2,2)
 
+  @test all(pad_zeros(randn(2), (1, 2))[[1, 4, 5]] .== 0)
+
   gradtest(x -> pad_constant(x, 2), rand(2,2,2))
+  gradtest(x -> pad_constant(x, (2, 1, 1, 2)), rand(2,2))
 end
 
 @testset "padding repeat" begin
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -111,18 +111,8 @@ end
 @inline tuplejoin(x, y) = (x..., y...)
 @inline tuplejoin(x, y, z...) = tuplejoin(tuplejoin(x, y), z...)
 
-function pad_zeros_tuple(pad::NTuple{L,Int}, N) where L
-  ntuple(2N) do i
-    if i <= L
-      pad[i]
-    else
-      0
-    end
-  end
-end
-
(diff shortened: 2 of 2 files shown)


CarloLucibello:
Thanks for this! Can you add tests for fixed issues?

torfjelde:
Ah, yes ofc! I did an hour ago but forgot to push. Did so now :+1:

DhairyaLGandhi:
We might want to hold on to the function so we can get the correct length tuple when we don't have all the dims and padding information.

DhairyaLGandhi:
Could we also test for padding vectors?

test/padding.jl
```diff
@@ -35,7 +38,10 @@
 
   @test size(pad_constant(x, 1, dims = 1)) == (4,2,2)
 
+  @test all(pad_zeros(randn(2), (1, 2))[[1, 4, 5]] .== 0)
+
   gradtest(x -> pad_constant(x, 2), rand(2,2,2))
+  gradtest(x -> pad_constant(x, (2, 1, 1, 2)), rand(2,2))
```
DhairyaLGandhi, line 44:
```suggestion
  gradtest(x -> pad_constant(x, (2, 1, 1, 2)), rand(2,2))
  gradtest(x -> pad_constant(x, (2, 1,)), rand(2))
```

DhairyaLGandhi (commented):
I've added a gradtesr for vectors as well

DhairyaLGandhi:
Thanks!

PR #299, opened 2021-03-16
Fix #282

### LINKED ISSUE #282: possible huge performance regression in `batched_mul`
i'm still looking into the cause and quantifying the magnitude of the effect, but i think it is caused by https://github.com/FluxML/NNlib.jl/pull/271.  has anyone else observed this?

mcabbott:


test/batchedmul.jl
```diff
@@ -119,17 +119,21 @@ end
     end
 end
 
+perm_12(A) = PermutedDimsArray(A, (2,1,3))
```
DhairyaLGandhi, line 122: Can you move these inside the testset

mcabbott: These don't depend on the T being looped over there. Seems clearer not to put them at the top of the file with all the other functions.

Issue #300, opened 2021-03-16
register NNlibCUDA as a subpackage of NNlib


CarloLucibello:

CarloLucibello:
@JuliaRegistrator register()  subdir=lib/NNlibCUDA

DilumAluthge:
@JuliaRegistrator register() subdir=lib/NNlibCUDA

CarloLucibello:
@DilumAluthge there is also failing `import NNlibCUDA` with CUDA 2.6, while CUDA 3.0.0 not being released yet. I just wanted to start the registration process to make testing against CUDA#master easier, but if that's a problem we can just wait for CUDA 3.0 release

DilumAluthge:
Yeah no worries at all! We can merge it manually in 3 days.

DhairyaLGandhi:
It might make sense to register https://github.com/FluxML/NNlibCUDA.jl instead. We wouldn't have to deal with registering it oddly and it keeps the projects separate in their own little silos. Currently NNlibCUDA doesn't get used by NNlib internally and developing it means getting unintended code alongside.

CarloLucibello:
> We wouldn't have to deal with registering it oddly

don't see any oddity here, this subpackage thing is actually quite cool and already well supported by julia's toolings

>  it keeps the projects separate in their own little silos

That's exactly what we don't want since the two packages are tightly coupled. When we change NNlib we want to make sure everything is fine with NNlibCUDA by running its tests, and sometimes we have to make changes that involve both packages (adding some function requiring also a cuda kernel). Having NNlibCUDA as a subpackage is much safer since it allows combined CI very easily (I already put that in place), makes contributing easier, and relieves a lot of the maintenance burden.
 
> Currently NNlibCUDA doesn't get used by NNlib internally and developing it means getting unintended code alongside.

I don't see what's the problem in also seeing the folder lib/NNlibCUDA when developing, if anyone gets confused by that he probably shouldn't be developing NNlib at all. Actually, as I said above, the possibility to make changes to both packages in a single PR is great.

DhairyaLGandhi:
[...]

PR #301, opened 2021-03-16
Add a `LICENSE.md file` for the `NNlibCUDA.jl` package

DilumAluthge:
Each subdirectory package needs to have its own copy of the license file. This is because when we distribute tarballs (e.g. via the Pkg server), the tarball only includes the subdirectory package. So we need to make sure that the subdirectory has a license; otherwise, the tarball won't have a license.

cc: @CarloLucibello

### DIFF
--- a/lib/NNlibCUDA/LICENSE.md
+++ b/lib/NNlibCUDA/LICENSE.md
@@ -1,0 +1,22 @@
+The NNlib.jl package is licensed under the MIT "Expat" License:
+
+> Copyright (c) 2017-19: Julia Computing, Inc., Mike J Innes, and Contributors
+> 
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Thanks!

DilumAluthge:
@DhairyaLGandhi @CarloLucibello You will now need to re-trigger Registrator. This will automatically update the General PR: https://github.com/JuliaRegistries/General/pull/32119

DilumAluthge:
I just realized that I'm a FluxML member.

I've re-triggered Registrator here: https://github.com/FluxML/NNlib.jl/issues/300

PR #303, opened 2021-03-19
Add fold.jl

spazewalker:
Adding Fold/Unfold functions in reference to pytorch feature parity [here](https://github.com/FluxML/Flux.jl/issues/1431).


src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+    x_dim = size(X)
+    if w_dim[end] != w_dim[end-1]
+        throw(DimensionMismatch("input channels must be equal to output channels in the kernel"))
+    end
+
+    # Make DenseConvDims object
+    cdims = DenseConvDims(x_dim, w_dim; stride=stride, padding=padding, dilation=dilation)
+
+    # Calculate the total number of sliding blocks
+    col_dim = (im2col_dims(cdims))[1:2]
```
DhairyaLGandhi, line 23: What does the 1:2 stand for here? Are you trying to extract the features from eg the channel separation/ batching dimension?

src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+    col = undef
+
+    # Iterate through all batchs
+    for i = 1:x_dim[end]
+        temp = fill(0., col_dim)
+        im2col!(temp, X[:,:,:,:,i], cdims)
```
DhairyaLGandhi, line 29: This would allocate the datapoint everytime, maybe use views?

CarloLucibello:
from a quick glance looks mostly ok, thanks. Convenience methods for the 1d and 2d cases are missing.
Then some tests and `rrule`s

src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+    if w_dim[end] != w_dim[end-1]
+        throw(DimensionMismatch("input channels must be equal to output channels in the kernel"))
+    end
+
+    # Make DenseConvDims object
+    cdims = DenseConvDims(x_dim, w_dim; stride=stride, padding=padding, dilation=dilation)
+
+    # Calculate the total number of sliding blocks
+    col_dim = (im2col_dims(cdims))[1:2]
+    col = undef
```
DhairyaLGandhi, line 24: I'm a little uncomfortable using undef since it have erratic behaviour in corner cases.

src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+
+    # Iterate through all batchs
+    for i = 1:x_dim[end]
```
DhairyaLGandhi, line 27: We might want to consider the possibility of having some other dimsension to fold over as well. Not critical though

src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+        temp = fill(0., col_dim)
+        im2col!(temp, X[:,:,:,:,i], cdims)
+        if i == 1
+            col = reshape(temp, col_dim[1], col_dim[2], 1)
+        else
+            col = cat(dims=3, col, temp)
```
DhairyaLGandhi, line 33: Typo?

src/fold.jl
```diff
@@ -0,0 +1,68 @@
...
+        end
+    end
+    return col;
```
DhairyaLGandhi, line 36: Prefer to drop the return, and there is a trailing semicolon

DhairyaLGandhi (commented):
Thanks so much to look into this! I've left a couple thoughts but overall I'm wondering if you've checked this on the GPU as well

src/fold.jl
spazewalker: `im2col_dims()` returns a Tuple of 3 elements `(L, M, N)` where N is the number of thread. However, `im2col!()` assumes a 2d array `col` in the shape of `(L, M)`. Hence the `[1:2]`, to take only the first two.

src/fold.jl
spazewalker: Noted. I'll instead use `fill()` to initialise and then use views to fill it with actual values in the loop.

src/fold.jl
spazewalker: `x_dim[end]` is the number of batches. This loop iterates over all batches and unfolds them, and then stack each unfolded array on top of the previous one along a new dimension. Folding over other dimensions can be implemented on top of this implementation by adding a dummy singleton dimension.

src/fold.jl
spazewalker: Noted.

src/fold.jl
spazewalker: I'll drop this if-else clause and use views here too.

src/fold.jl
spazewalker: I'll remove the semicolon and are you talking about dropping the return statement or the return character?

spazewalker (commented):
> from a quick glance looks mostly ok, thanks. Convenience methods for the 1d and 2d cases are missing.
> Then some tests and `rrule`s

Should I implement them here or will they be taken care of while implementing a layer in Flux.jl.

spazewalker:
> Thanks so much to look into this! I've left a couple thoughts but overall I'm wondering if you've checked this on the GPU as well

I haven't tested it on GPU yet, I'll check it.

DhairyaLGandhi:
We might not have a layer in or maybe we need to just hold the config to construct the cdims object.

ToucheSir:
> > from a quick glance looks mostly ok, thanks. Convenience methods for the 1d and 2d cases are missing.
> > Then some tests and `rrule`s
> 
> Should I implement them here or will they be taken care of while implementing a layer in Flux.jl.

Thanks for the PR! Tests and rrules need to be here. I'm not sure what exactly is meant by convenience methods, but if it makes sense to call directly (as opposed to only as a helper for a layer struct) then it should be in NNlib as well.

spazewalker:
> We might not have a layer in or maybe we need to just hold the config to construct the cdims object.

Okay, so I'll try to implement those convenience wrappers.

> > > from a quick glance looks mostly ok, thanks. Convenience methods for the 1d and 2d cases are missing.
> > > Then some tests and `rrule`s
> > 
> > 
> > Should I implement them here or will they be taken care of while implementing a layer in Flux.jl.
> 
> Thanks for the PR! Tests and rrules need to be here. I'm not sure what exactly is meant by convenience methods, but if it makes sense to call directly (as opposed to only as a helper for a layer struct) then it should be in NNlib as well.

As per my understanding, Those are the wrappers around this implementation, so that it can handle 1d and 2d inputs too. I'll work on tests but I'm not sure what 'rrule`s are.

ToucheSir:
`rrules` are what allow for the AD system and gradients to work, more details at https://juliadiff.org/ChainRulesCore.jl/stable/#frule-and-rrule. For functions like this that use mutation and inner loops, you'll probably need another function that manually performs the backwards pass. They're a little abstracted, but it may be worth looking at how the conv rrules work: https://github.com/FluxML/NNlib.jl/blob/7b8ae45c7fdcb331d63229a0a848e588b7d164a1/src/conv.jl#L223.

src/fold.jl
```diff
@@ -0,0 +1,70 @@
...
+function fold(col::AbstractArray{T,3} where T, out_dim::NTuple{M}, w_dim::NTuple{M}; stride=1, padding=0, dilation=1) where M
+    # Validate input
+    if length(out_dim) > 3
+        throw(DimensionMismatch("output dimentions cannot be greater than 3, got $(ndims(out_dim))"))
+    end
+
+    # Create DenseConvDims object
+    col_dim = size(col)
+    channels = col_dim[2]÷prod(w_dim)
+    x_dim = (out_dim... , fill(3-length(out_dim))... , channels,col_dim[3])
```
DhairyaLGandhi, line 58: The calls to `fill` seem suspect. Did you need `ntuple` there?

src/fold.jl
```diff
@@ -0,0 +1,70 @@
...
+
+    img = fill(0., x_dim)
+
+    # Iterate through all batchs
+    @views for i = 1:x_dim[end]
+        col2im!(img[:,:,:,:,i], col[:,:,i], cdims)
+    end
+
+    return reshape(img, (out_dim... , channels,col_dim[3]))
+end
```
DhairyaLGandhi, line 70: Definitely don't want to see this. Possibly the editor is using file endings different from Linux. I'd check that.

src/fold.jl
spazewalker: needed a `ntuple` with dimensions of `out_dim` appended with dummy dimension to make it 3d. I think it should be `fill(1,3-length(out_dim))`. It's okay to use `fill` here right?

DhairyaLGandhi: Use `insert_singleton_dimension` instead. Or a `reshape`. `fill` is incorrect here.

src/fold.jl
```diff
@@ -0,0 +1,70 @@
+export unfold, fold
```
DhairyaLGandhi, line 1: Let's not export them, the names would dirty the name space very quickly.

mcabbott:
Replaced by #444

PR #304, opened 2021-03-22
Add benchmarks, part 1

dfdx:
Given the increasing importance of NNlib in the ML ecosystem, I believe it's time to add automatic benchmarks. This PR is based on the amazing PkgBenchmark.jl and [config from ProximalOperators.jl](https://github.com/kul-forbes/ProximalOperators.jl/tree/master/benchmark). This is only partial improvement since benchmark code compares a change with a baseline in master, but current master doesn't have any benchmarks at all, so **benchmark job will fail by design**. All the following PRs (with or without changes to the benchmarks) should work fine. 

To be precise, here's what this PR does: 

* adds a set of simple benchmarks which can be triggered from the command line using `julia --project=benchmark benchmark/runbenchmarks.jl`
* adds GitHub Actions job to run these benchmarks automatically on every PR

What this PR does not:

* doesn't provide finished and verified work - this kind of things can only be tested by merging into master and seeing CI's output; yet I'll try to finalize it reasonably quickly
* doesn't add a comprehensive benchmark set - I think we need to add them step by step

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,20 @@
+name: Benchmark
+on:
+  - pull_request
+jobs:
+  benchmark:
+    name: Benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1.5'
+      - run: git fetch origin '+refs/heads/master:refs/remotes/origin/master'
+      - run: git branch master origin/master
+      - run: |
+          julia --project=benchmark -e '
+            using Pkg
+            Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.instantiate()'
+      - run: julia --project=benchmark benchmark/runbenchmarks.jl
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ deps/usr
 deps.jl
 *.log
 ./Manifest.toml
+benchmark/*.json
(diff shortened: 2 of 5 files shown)


DhairyaLGandhi:
Thanks! Benchmarking is very important. I think we would want to consolidate these in FluxBench.jl (in this org), so that we can reliably track and reproduce our benchmarks. So maybe moving this over there would be better too.

dfdx:
Thanks, I didn't know about FluxBench! What's the intended usage for it? I thought about running the benchmarks on every PR automatically to notify the author about possible performance regressions before it gets merged. Do you have in mind something similar for FluxBench?

CarloLucibello:
can this be triggered only per request, instead of every commit of every PR? I'm thinking to something similar to `@nanosoldier` for Base.

DhairyaLGandhi:
That's the job of the package, to have things in one place and be called on by the keeper bot

CarloLucibello:
@DhairyaLGandhi can you point @dfdx in the right direction for contributing? I didn't know about FluxBench and keeperbot as well, there are no comments about it and no documentation anywhere. Is keeper bot working already?

johnnychen94:
@tkf provides a useful tool [BenchmarkCI](https://github.com/tkf/BenchmarkCI.jl) for this very specific purpose. It also supports an optional benchmark on a specific label.

DhairyaLGandhi:
Flux Bot*

BenchmarksCI is definitely a good shout. This is also important to not have to deal with shared workers or different system setups adding noise, bit rather working on dedicated benchmarking machines.

dfdx:
BenchmarkCI looks great! A couple of things I didn't understand from the discussion (perhaps, missing some context): 

* what is keeperbot / Flux Bot? I can't see any reference to them
* should we add benchmarks and CI for them in this repo or put everything to FluxBench?
* what is the optimal trigger for the benchmarks? 

I guess some of the previous comments answer exactly these questions, but I can't connect the dots.

CarloLucibello:
This can help https://github.com/marketplace/actions/pull-request-comment-trigger

PR #296, opened 2021-03-15
Add scatter for CUDA support

yuehhua:

PR #306, opened 2021-03-25
[nnlibcuda] introduce a manifest and remove cufunc

CarloLucibello:
Since we have to checkout the master branch of a few packages it is convenient to have a Manifest until CUDA.jl 3.0.0 is released.

Also seems `@cufunc` is no longer needed and has been removed from CUDA.jl.

cfr #296


CarloLucibello:
Needs also https://github.com/FluxML/Zygote.jl/pull/929

PR #308, opened 2021-04-07
scatter and gather support element type of idx to be CartesianIndex

yuehhua:


src/gather.jl
```diff
@@ -22,20 +25,21 @@ or multiple `dst` columns.
 
 See [`gather`](@ref) for an allocating version.
 """
+gather!(dst::AbstractArray, src::AbstractArray, idx::AbstractArray{<:IntOrIntTuple}) = gather!(dst, src, CartesianIndex.(idx))
```
CarloLucibello, line 28: Indexing with an array of integers is by far the most common use case (in fact it is the only one implemented in ScatterNNlib). Are we sure we want to take the performance impact of the unnecessary allocation `CartesianIndex.(idx)` just to support CartesianIndexes and avoid code duplication?

src/gather.jl
```diff
@@ -22,20 +25,21 @@ or multiple `dst` columns.
...
-    d = Ndst - Nidx 
+    d = Ndst - Nidx
     d == Nsrc - M || throw(ArgumentError("Incompatible input shapes."))
     size(dst)[1:d] == size(src)[1:d] || throw(ArgumentError("Incompatible input shapes."))
     size(dst)[d+1:end] == size(idx) || throw(ArgumentError("Incompatible input shapes."))
 
     colons = ntuple(i -> Colon(), d)
     for k in CartesianIndices(idx)
-        view(dst, colons..., k) .= view(src, colons..., idx[k]...)
+        view(dst, colons..., k) .= view(src, colons..., idx[k])
```
CarloLucibello, line 42: Maybe we can define
```julia
_view(src, colons, idx::AbstractArray{CartesianIndex}, k) = view(src, colons..., idx[k])
_view(src, colons, idx, k) = view(src, colons..., idx[k]...)
```
and use it here instead of converting everything to CartesianIndexes (see comment above)

DhairyaLGandhi:
One complaint that would be good to address is that NNlib should not be exporting gather and scatter. They are pretty generic names which may mean different things in different contexts. scatter for plotting - for ex

src/gather.jl
DhairyaLGandhi: Best to avoid unnecessary indirection

CarloLucibello: it's not unnecessary, I'm suggesting it in order to avoid a copy of the indexes, see comment above

DhairyaLGandhi: The compiler can avoid the complete copy if it can determine the types

CarloLucibello: I don't think the compiler can avoid producing a copy when doing `CartesianIndex.(idx)`

CarloLucibello:
unexporting gather and scatter is a breaking change

yuehhua:
However, scatter and gather are both essential low-level operations in computer science. These two operations here are to mimic the behavior of [gather-scatter](https://en.wikipedia.org/wiki/Gather-scatter_(vector_addressing)).
It is essential that convolution and inner product is to CNN what scatter and gather is to GNN.

src/gather.jl
yuehhua: I will rethink about it.

src/gather.jl
yuehhua: I checked the implementation of `CartesianIndex`. It contains a NTuple as a field, so it is surely that converting to a `CartesianIndex` will have new allocation or encapsulation.

DhairyaLGandhi:
Exporting them is a bug, which should come under semver. It isn't changing the behaviour of the functions itself.

src/gather.jl
DhairyaLGandhi: Do we need to loop over the CartesianIndices? Can't we use them directly?

DhairyaLGandhi: `CartesianIndices` instead of `AbstractArray{CartesianIndex}` ?

ChrisRackauckas:
The fact that it interferes with Plots.jl means that many online tutorials of Flux are broken by exporting it (at least, if they use `scatter` plots). That's pretty breaking. That at least deserves a major release, a blog post, and a bunch of downstream issues / comments around blogs to help people get updated. The exporting of such a common term with known breakage against libraries that are commonly used with Flux is such a huge breaking change that I am kind of shocked it went through without more consideration.

CarloLucibello:
ok, I'm filing a PR removing the exports and tagging a patch release

DhairyaLGandhi:
Ref the offending PR https://github.com/FluxML/NNlib.jl/pull/255

src/gather.jl
yuehhua: `CartesianIndices` requires the dimensions of both inner `CartesianIndex` and outer `AbstractArray` to be the same. You may check it here: `struct CartesianIndices{N,R<:NTuple{N,OrdinalRange{Int, Int}}} <: AbstractArray{CartesianIndex{N},N}`. However, it is not required to be the same in scatter and gather operations. You may find a test case of `Array{NTuple{1},2}` in my test case.

yuehhua:
@DhairyaLGandhi @CarloLucibello Is this ready to go?

src/scatter.jl
```diff
@@ -9,17 +9,39 @@
...
+function _check_dims(Nx, Ny, M, Nidx)
+    @assert Nx - M == Ny - Nidx "Incompatible input shapes of (dst, src, idx) = ($Nx, $Ny, $Nidx)."
+    dims = Nx - M
+    dims < 0 && throw(ArgumentError("dims must be non-negative but got dims=$dims."))
+    return dims
+end
+
+_view(X, colons, idx::AbstractArray{CartesianIndex}, k::CartesianIndex) = view(X, colons..., idx[k])
+_view(X, colons, idx, k::CartesianIndex) = view(X, colons..., idx[k]...)
+_view(X, colons, k::CartesianIndex) = view(X, colons..., k)
```
CarloLucibello, line 44:
```suggestion
_view(X, colons, k) = view(X, colons..., k...)
_view(X, colons, k::Union{Integer, CartesianIndex}) = view(X, colons..., k)
```

src/scatter.jl
```diff
@@ -42,30 +64,18 @@ index of `dst` and the value of `idx` must indicate the last few dimensions of `
...
-
-function scatter!(op, dst::AbstractArray{Tdst}, src::AbstractArray{Tsrc}, idx::AbstractArray{<:IntOrIntTuple},
-                  dims::Val{N}) where {Tdst,Tsrc,N}
+function scatter!(op, dst::AbstractArray, src::AbstractArray, idx::AbstractArray)
+    dims = _check_dims(dst, src, idx)
     colons = Base.ntuple(_->Colon(), dims)
     for k in CartesianIndices(idx)
-        dst_v = view(dst, colons..., idx[k]...)
-        src_v = view(src, colons..., k)
+        dst_v = _view(dst, colons, idx, k)
```
CarloLucibello, line 71:
```suggestion
        dst_v = _view(dst, colons, idx[k])
```

src/gather.jl
```diff
@@ -20,20 +20,11 @@ or multiple `dst` columns.
...
-    size(dst)[1:d] == size(src)[1:d] || throw(ArgumentError("Incompatible input shapes."))
-    size(dst)[d+1:end] == size(idx) || throw(ArgumentError("Incompatible input shapes."))
-
-    colons = ntuple(i -> Colon(), d)
+function gather!(dst::AbstractArray, src::AbstractArray, idx::AbstractArray)
+    dims = _check_dims(src, dst, idx)
+    colons = ntuple(i -> Colon(), dims)
     for k in CartesianIndices(idx)
-        view(dst, colons..., k) .= view(src, colons..., idx[k]...)
+        _view(dst, colons, k) .= _view(src, colons, idx, k)
```
CarloLucibello, line 27:
```suggestion
        _view(dst, colons, k) .= _view(src, colons, idx[k])
```

CarloLucibello:
besides the slight simplification I suggested this seems ready to go

CarloLucibello:
Almost there, could add at least one test

DhairyaLGandhi:
Seems like we can get rid of `_view`?

yuehhua:
I noticed that I forgot to support `gather` and I add support for that.

src/scatter.jl
```diff
@@ -9,17 +9,39 @@
...
+    M = typelength(Tidx)
+    dims = _check_dims(Nx, Ny, M, Nidx)
+    size(X)[1:dims] == size(Y)[1:dims] || throw(ArgumentError("Incompatible input shapes."))
+    size(Y)[dims+1:end] == size(idx) || throw(ArgumentError("Incompatible input shapes."))
     return dims
 end
 
-typelength(::Type{<:Number}) = 1
-typelength(::Type{<:NTuple{M}}) where M = M
+function _check_dims(X::AbstractArray{Tx,Nx},
```
DhairyaLGandhi, line 27: Seems like we can unify the two methods with just converting the Cartesian indices once?

CarloLucibello: these are very low-level methods, we should not do any allocation

PR #309, opened 2021-04-08
unexport gather/scatter

CarloLucibello:
see discussion in #308


DhairyaLGandhi:
Please remove the lines if they aren't being used, don't comment them.

PR #313, opened 2021-04-14
Avoiding scalar indexing in cuda

yuehhua:
Existing implementation use scalar indexing for cuarray. To avoid using scalar indexing, refactoring is done.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -12,7 +12,4 @@ Return the maximum value for each dimension. An array of dimensions `dims` is ac
 The maximum of each dimension in the element is computed.
 """
 maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
-
-function maximum_dims(dims::AbstractArray{<:Tuple})
-    Tuple(maximum(xs) for xs in zip(dims...))
-end
+maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = Tuple(maximum(x->x[i], dims) for i = 1:N)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -12,4 +12,4 @@ Return the maximum value for each dimension. An array of dimensions `dims` is ac
 The maximum of each dimension in the element is computed.
 """
 maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
-maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = Tuple(maximum(x->x[i], dims) for i = 1:N)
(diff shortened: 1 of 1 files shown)


ToucheSir:
The CI failure appears to be unrelated to this change? AFAICT none of the conv machinery uses `maximum_dims`.

yuehhua:
`maximum_dims` is used in `scatter` operations in scatter.jl. Then, I am sure that the CI failure is unrelated to this change.

src/utils.jl
```diff
@@ -12,7 +12,4 @@ Return the maximum value for each dimension. An array of dimensions `dims` is ac
 The maximum of each dimension in the element is computed.
 """
 maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
-
-function maximum_dims(dims::AbstractArray{<:Tuple})
-    Tuple(maximum(xs) for xs in zip(dims...))
-end
+maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = Tuple(maximum(x->x[i], dims) for i = 1:N)
```
DhairyaLGandhi, line 15: I think the correct answer might be to use `ntuple` instead of going over an array.

DhairyaLGandhi:
Mostly looks good, left a thought about a design improvement.

src/utils.jl
DhairyaLGandhi: Maybe not in the current setup though

DhairyaLGandhi:
Reopening to see what's up with CI

DhairyaLGandhi:
Yeah, seems CI is unrelated. @yuehhua happy to merge if you can address replacing this with `maximum`

yuehhua:
The functionality of this function is to calculate the maximum values across indices. For example, I have a array of tuples that contains a series of indices. 

```julia
julia> A = [(5, 2), (3, 4), (1, 6)];

julia> maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = Tuple(maximum(x->x[i], dims) for i = 1:N);

julia> maximum_dims(A)
(5, 6)

julia> maximum(A)
(5, 2)
```

The maximum values of first component across indices is 5 and the maximum values of second component is 6. So, I want `maximum_dims` to return me the tuple `(5, 6)`, while `maximum` give me the element which has maximum value in its first component.

src/utils.jl
yuehhua: I have no idea how to use `ntuple` in this case. Could you give me more hints?

CarloLucibello, line 15:
```suggestion
maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = ntuple(i -> maximum(x->x[i], dims), N)
```

DhairyaLGandhi:
Thanks all

PR #314, opened 2021-04-14
Move NNlibCUDA

DhairyaLGandhi:
Moves NNlibCUDA to its own repo (ref #300), and reinstates the ability to test outside of the CI environment. Good to unblock @maleadt with a CUDA compatible Flux release.


CarloLucibello (approved):

test/runtests.jl
```diff
@@ -56,3 +56,19 @@ end
...
+        using NNlibCUDA
+        @testset "CUDA" begin
+            Pkg.test("NNlibCUDA")
+        end
+    else
+        @info "Skipping CUDA tests, set NNLIB_TEST_CUDA=true to run them"
+    end
+else
+    @info "Insufficient version or CUDA not found; Skipping CUDA tests"
+end
```
CarloLucibello, line 74:
```suggestion
if VERSION >= v"1.6" 
   if get(ENV, "NNLIB_TEST_CUDA", "false") == "true"
        import Pkg
        Pkg.develop(url = "https://github.com/FluxML/NNlibCUDA.jl")
        using NNlibCUDA
        @testset "CUDA" begin
            Pkg.test("NNlibCUDA")
        end
    else
        @info "Skipping CUDA tests, set NNLIB_TEST_CUDA=true to run them"
    end
else
    @info "Can test on CUDA only on julia  >=  1.6"
end
```

CarloLucibello: If we avoid importing CUDA.jl we can run CI also on julia 1.3 (currently failing).  
The CUDA.functional() test should be executed by NNlibCUDA's tests in any case

DhairyaLGandhi: I had it in here specifically to run cuda tests for kernels that don't need rewrites and can be tested independent of the other package. Potentially we can check for `CUDA.has_cuda` instead.

CarloLucibello: ok. So we have to fix the compat issue on julia 1.3 or just ditch julia 1.3 (seems fine to move NNlib to julia >= 1.5). 
The compat issue is 
```
ERROR: Unsatisfiable requirements detected for package CUDA [052768ef]:
 CUDA [052768ef] log:
 ├─possible versions are: [0.1.0, 1.0.0-1.0.2, 1.1.0, 1.2.0-1.2.1, 1.3.0-1.3.3, 2.0.0-2.0.2, 2.1.0, 2.2.0-2.2.1, 2.3.0, 2.4.0-2.4.3, 2.5.0, 2.6.0-2.6.3, 3.0.0-3.0.2] or uninstalled
 ├─restricted to versions * by an explicit requirement, leaving only versions [0.1.0, 1.0.0-1.0.2, 1.1.0, 1.2.0-1.2.1, 1.3.0-1.3.3, 2.0.0-2.0.2, 2.1.0, 2.2.0-2.2.1, 2.3.0, 2.4.0-2.4.3, 2.5.0, 2.6.0-2.6.3, 3.0.0-3.0.2]
 ├─restricted by compatibility requirements with NNlib [872c559c] to versions: [1.0.2, 1.1.0, 1.2.0-1.2.1, 1.3.0-1.3.3, 2.0.0-2.0.2, 2.1.0, 2.2.0-2.2.1, 2.3.0, 2.4.0-2.4.3, 2.5.0, 2.6.0-2.6.3, 3.0.0-3.0.2] or uninstalled, leaving only versions: [1.0.2, 1.1.0, 1.2.0-1.2.1, 1.3.0-1.3.3, 2.0.0-2.0.2, 2.1.0, 2.2.0-2.2.1, 2.3.0, 2.4.0-2.4.3, 2.5.0, 2.6.0-2.6.3, 3.0.0-3.0.2]
 │ └─NNlib [872c559c] log:
 │   ├─possible versions are: 0.7.18 or uninstalled
 │   └─NNlib [872c559c] is fixed to version 0.7.18
 ├─restricted by julia compatibility requirements to versions: [0.1.0, 1.0.0-1.0.2, 1.3.0-1.3.3] or uninstalled, leaving only versions: [1.0.2, 1.3.0-1.3.3]
 └─restricted by compatibility requirements with Adapt [79e6a3ab] to versions: [2.4.0-2.4.3, 2.5.0, 2.6.0-2.6.3, 3.0.0-3.0.2] or uninstalled — no versions left
   └─Adapt [79e6a3ab] log:
     ├─possible versions are: [0.3.0-0.3.1, 0.4.0-0.4.2, 1.0.0-1.0.1, 1.1.0, 2.0.0-2.0.2, 2.1.0, 2.2.0, 2.3.0, 2.4.0, 3.0.0, 3.1.0-3.1.1, 3.2.0, 3.3.0] or uninstalled
     ├─restricted to versions 3.2.0-3 by NNlib [872c559c], leaving only versions [3.2.0, 3.3.0]
     │ └─NNlib [872c559c] log: see above
     └─restricted to versions 3.3.0 by an explicit requirement, leaving only versions 3.3.0
```
Probably we can solve it if we relax Adapt bounds

DhairyaLGandhi: We can't use NNlibCUDA on < 1.6 anyway

DhairyaLGandhi:
Our tests have gotten much more fragile recently. It might have to with the new code for AD being brittle.

test/runtests.jl
```diff
@@ -56,3 +56,19 @@ end
     include("utils.jl")
 end
 
+using CUDA
+
+if VERSION >= v"1.6" && CUDA.functional()
+    if get(ENV, "NNLIB_TEST_CUDA", "false") == "true"
+        import Pkg
+        Pkg.develop(url = "https://github.com/FluxML/NNlibCUDA.jl")
```
DhairyaLGandhi, line 64: Move to a test dep once this is tagged

DhairyaLGandhi:
Merging now

ToucheSir:
It's a little late now, but should we classify this kind of a change as breaking in future? https://github.com/FluxML/NNlib.jl/issues/322 is a demonstration of how a minor version upgrade could drop previously working functionality (NNlib + CUDA used to be enough, now NNlibCUDA is required as well).

PR #316, opened 2021-04-22
Add NNlibCUDA to test deps

DhairyaLGandhi:
Best to test with released NNlibCUDA now that its registered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
We will have to drop support for 1.5 to see this through for Pkg to resolve this correctly and to not accidentally give users packages that aren't required. NNlib itself should be fine to work on 1.5, but I think moving forward we might have to lower limit NNlib in Flux. We can release Flux since https://github.com/FluxML/Flux.jl/pull/1571 handles the changes gracefully since Flux will bump its Julia requirement.

PR #318, opened 2021-05-16
Add gradient for gather

yuehhua:


src/gather.jl
```diff
@@ -71,3 +71,13 @@ function gather(src::AbstractArray{Tsrc, Nsrc},
     colons = ntuple(i -> Colon(), Nsrc-1)
     return src[colons..., idx]
 end
+
+# Gradient
+
+∇gather_src!(Δ, src, idx) = scatter!(+, fill!(similar(src), 0), Δ, idx)
+
+function rrule(::typeof(gather!), dst::AbstractArray, src::AbstractArray, idx::AbstractArray)
```
mcabbott, line 79: Why a rule for `gather!`, not for `gather`? The latter would be the more standard no-mutation-in-AD story. But perhaps this makes sense since `dst` is completely overwritten?

I think your `∇gather_src!` doesn't really want a `!`, as it doesn't mutate any arguments. And perhaps it could just call `scatter`, does it need `src` as an argument at all?

yuehhua: `gather` is totally rely on `gather!`, since `dst` is overwritten.

Yeah, `∇gather_src!` didn't mutate any arguments. It is just the gradient for `gather!`. Is there any new naming suggestions?

mcabbott: Re `gather!` not `gather`, I guess it it works that's OK. 

Re the gradient function, I think omit the `!` unless it mutates its own argument. (The gradients of `gather` and `gather!` would be identical anyway, right?). You can write a mutating one and a non-mutating one which calls that, which which if wrapped in an `InplaceableThunk` may someday save some memory... but not obligatory. 

It would be nice not to close over `src` though. Could it just be `scatter(+, Δ, ind)`? Or if you need `size(src)`, could you pass that instead?

src/gather.jl
```diff
@@ -71,3 +71,13 @@ function gather(src::AbstractArray{Tsrc, Nsrc},
     colons = ntuple(i -> Colon(), Nsrc-1)
     return src[colons..., idx]
 end
+
+# Gradient
+
+∇gather_src!(Δ, src, idx) = scatter!(+, fill!(similar(src), 0), Δ, idx)
```
CarloLucibello, line 77:
```suggestion
∇gather_src(Δ, src, idx) = scatter!(+, fill!(similar(src), 0), Δ, idx)
```

yuehhua: Done.

src/gather.jl
yuehhua: Done. Thank you for suggestions!

CarloLucibello: > It would be nice not to close over src though. Could it just be scatter(+, Δ, ind)? Or if you need size(src), could you pass that instead?

Right, the more can be GCed the better. It should be somthing like
```julia
src_size = size(src)
gather!_pullback(Δ) = (NO_FIELDS, DoesNotExist(), scatter!(+, fill!(similar(Δ, eltype(Δ), src_size), 0), Δ, idx), DoesNotExist())
```

CarloLucibello:
Maybe add a test for the tuple case?

CarloLucibello:
thanks!

PR #319, opened 2021-05-24
scatter cleanup and fix min/max

### LINKED ISSUE #317: add rrules for max/min scatter
Couldn't see this through in  #297 since there was a bug

CarloLucibello:
Fix #317. The main problem was in the use of FiniteDifferences' central difference method on singularities of non-smooth functions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,7 +5,6 @@ version = "0.7.19"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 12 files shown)


yuehhua:
I am quite curious about the semantic difference between `DoesNotExist` and `NoTangent`.

CarloLucibello:
> I am quite curious about the semantic difference between `DoesNotExist` and `NoTangent`.

no difference, just a recent renaming in ChainRulesCore.jl

yuehhua:
Maybe we could come out with a new name for `_check_dims`? Maybe `_check_scatter_dims` or something else?

yuehhua:
For `scatter`, I have been used to write something like this.

```julia
scatter(op,
        src::AbstractArray{Tsrc,Nsrc},
        idx::AbstractArray{Tidx,Nidx};
        init = scatter_empty(op, Tsrc))
```
I think it can avoid the branching on `isnothing` or not.

yuehhua:
The dimension checking is slightly different for `scatter` and `gather`. `scatter` calls `_check_dims`/`scatter_dims` with `(dst, src, idx)`, while, reversely, `gather` calls `_check_dims`/`scatter_dims` with `(src, dst, idx)`. For abstraction, I give `_check_dims`/`scatter_dims` a signature of `(X, Y, idx)`, so they are both available for `scatter` and `gather`.

yuehhua:
I am good with this PR.

CarloLucibello:
buildkite fail is unrelated, merging

PR #320, opened 2021-05-24
introduce linear upsampling

maxfreu:
Hi! This introduces linear upsampling along the first dimension of a 3D array. GPU PR will follow. Again, this adds duplicate code - either I can work longer on this PR and boil it down or in a subsequent PR, as you like it.

Edit: Breaking tests *seem* unrelated :)


CarloLucibello:
nice, thanks!

PR #323, opened 2021-06-03
Update ChainRulesCore

willtebbutt:


willtebbutt:
@CarloLucibello the test failures don't look related to this PR to me, and they're different in different CI runs. Any thoughts?

CarloLucibello:
Yes they are unrelated, we should do something about them

PR #326, opened 2021-06-17
Make reverse_indices gpu compatible

yuehhua:
Resolve issue for FluxML/NNlibCUDA.jl#13.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -16,11 +16,11 @@ maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = ntuple(i -> maximu
 maximum_dims(dims::AbstractArray{CartesianIndex{N}}) where {N} = ntuple(i -> maximum(x->x[i], dims), N)
 
 function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
-    for (ind, val) = pairs(Array(idx))
+    for (ind, val) in pairs(Array(idx))
         push!(rev[val], ind)
     end
     # if CUDA supports `unique`, a more efficient version:
-    # cidx = CartesianIndices(idx)
+    # cidx in CartesianIndices(idx)
     # for i = unique(idx)
     #     rev[i] = cidx[idx .== i]
     # end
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -15,15 +15,26 @@ maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
...
+    # cidx = CartesianIndices(idx)
+    # for i = unique(idx)
+    #     rev[i] = cidx[idx .== i]
+    # end
     rev
 end
 
+function reverse_indices(idx::AbstractArray)
+    rev = Array{Vector{CartesianIndex}}(undef, maximum_dims(idx)...)
+    for i = eachindex(rev)
```
CarloLucibello, line 32:
```suggestion
    for i in eachindex(rev)
```

src/utils.jl
```diff
@@ -15,15 +15,26 @@ maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
 maximum_dims(dims::AbstractArray{NTuple{N, T}}) where {N,T} = ntuple(i -> maximum(x->x[i], dims), N)
 maximum_dims(dims::AbstractArray{CartesianIndex{N}}) where {N} = ntuple(i -> maximum(x->x[i], dims), N)
 
-function reverse_indices(idx::Array{T}) where T
-    rev = Dict{T,Vector{CartesianIndex}}()
-    for (ind, val) = pairs(idx)
-        rev[val] = get(rev, val, CartesianIndex[])
+function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
+    for (ind, val) = pairs(Array(idx))
```
CarloLucibello, line 19:
```suggestion
    for (ind, val) in pairs(Array(idx))
```

src/utils.jl
```diff
@@ -15,15 +15,26 @@ maximum_dims(dims::AbstractArray{<:Integer}) = (maximum(dims), )
...
         push!(rev[val], ind)
     end
+    # if CUDA supports `unique`, a more efficient version:
+    # cidx = CartesianIndices(idx)
+    # for i = unique(idx)
```
CarloLucibello, line 24:
```suggestion
    # for i in unique(idx)
```

CarloLucibello (changes requested):

src/utils.jl
yuehhua: Done.

src/utils.jl

src/utils.jl

CarloLucibello:
It's a pity we have to perform the whole computation on cpu, but I guess we don't have a better solution right now

PR #328, opened 2021-06-21
Fix the output type of reverse_indices

yuehhua:
Solve the issue related to FluxML/NNlibCUDA.jl#13.
`CartesianIndex` itself is not sufficient, but `CartesianIndex{N}` is the full type.
A full type is required to convert an object into cuarray and pass it into a cuda kernel.
At the same time, I am considering making an `reverse_indices` API for cuda.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -35,9 +35,11 @@ function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
 end
 
 function reverse_indices(idx::AbstractArray)
-    rev = Array{Vector{CartesianIndex}}(undef, maximum_dims(idx)...)
+    max_dims = maximum_dims(idx)
+    T = CartesianIndex{length(size(idx))}
+    rev = Array{Vector{T}}(undef, max_dims...)
     for i in eachindex(rev)
-        rev[i] = CartesianIndex[]
+        rev[i] = T[]
     end
     return reverse_indices!(rev, idx)
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -34,9 +34,18 @@ function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
     rev
 end
 
-function reverse_indices(idx::AbstractArray)
+"""
+    reverse_indices(idx)
+
+Return the reverse indices of `idx`. The indices of `idx` will be values, and values of `idx` will be index.
+
+# Arguments
+
+- `idx`: The indices to be reversed. Accepts array or cuarray of integer, tuple or `CartesianIndex`.
+"""
+function reverse_indices(idx::AbstractArray{<:Any,N}) where N
     max_dims = maximum_dims(idx)
-    T = CartesianIndex{length(size(idx))}
+    T = CartesianIndex{N}
     rev = Array{Vector{T}}(undef, max_dims...)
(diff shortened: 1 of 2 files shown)


yuehhua:
I found that it is possible to generate reverse index for cuarray. However, it will remain the form of array of cuarray and the type is `Array{CuArray{CartesianIndex{2}, 1}}`.

```julia
julia> res = [
               CartesianIndex.([(1,1), (2,3)]),
               CartesianIndex.([(1,2), (2,2)]),
               CartesianIndex.([(3,1), (1,3), (2,4), (3,4)]),
               CartesianIndex.([(2,1), (1,4)]),
               CartesianIndex.([(3,2), (3,3)])
           ];

julia> cures = [cu(x) for x in res]
5-element Vector{CuArray{CartesianIndex{2}, 1}}:
 CartesianIndex{2}[CartesianIndex(1, 1), CartesianIndex(2, 3)]
 CartesianIndex{2}[CartesianIndex(1, 2), CartesianIndex(2, 2)]
 CartesianIndex{2}[CartesianIndex(3, 1), CartesianIndex(1, 3), CartesianIndex(2, 4), CartesianIndex(3, 4)]
 CartesianIndex{2}[CartesianIndex(2, 1), CartesianIndex(1, 4)]
 CartesianIndex{2}[CartesianIndex(3, 2), CartesianIndex(3, 3)]
```

Is there a way for generating nested array where the inner array is a cuarray easily, like `similar`?

yuehhua:
trigger CI

src/utils.jl
```diff
@@ -34,10 +34,14 @@ function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
     rev
 end
 
-function reverse_indices(idx::AbstractArray)
-    rev = Array{Vector{CartesianIndex}}(undef, maximum_dims(idx)...)
+reverse_indices(idx::AbstractArray) = _reverse_indices(idx)
```
CarloLucibello, line 37: doesn't seem we need `_reverse_indices`, we can just move the implementation to `rerverse_indices` itself

Also, it would be good to add a docstring

src/utils.jl
```diff
@@ -34,10 +34,14 @@ function reverse_indices!(rev::AbstractArray, idx::AbstractArray)
...
+
+function _reverse_indices(idx::AbstractArray)
+    max_dims = maximum_dims(idx)
+    T = CartesianIndex{length(size(idx))}
```
CarloLucibello, line 41:
```suggestion
function _reverse_indices(idx::AbstractArray{<:Any,N}) where N
    max_dims = maximum_dims(idx)
    T = CartesianIndex{N}
```

src/utils.jl
yuehhua: I put another API for `reverse_indices(idx::AnyCuArray)`, and in that API it calls `_reverse_indices(idx)`.

src/utils.jl

src/utils.jl
CarloLucibello: I think it's fine you copy the implementation in NNlibCUDA since it is quite short, this way we rely less on internals and it also becomes quite clear that we are going through cpu operations and at some point we will need a proper kernel

PR #334, opened 2021-07-15
add groups to DenseConvDims (w/o implementation of grouped conv)

CarloLucibello:
Factor out this small change from #289 to accelerate things.
After this PR, we can proceed with cudnn integration (https://github.com/FluxML/NNlibCUDA.jl/pull/9)


DhairyaLGandhi:
Unsure why this is done, the groups PR got the reviews it needed and is ready to merge too.

CarloLucibello:
> Unsure why this is done, the groups PR got the reviews it needed and is ready to merge too.

because it has been in the work for 4 months, and since this is an outstanding missing feature we have I wanted to speed up things, e.g. work on the CUDA side of this, but having the support for froups in DenseConvDims is a necessary condition. If #289 is nearly fineshed I'm totally happy, just didn't look complete nor actively worked

DhairyaLGandhi:
It got the review I had wanted just yesterday (late night my time too), and have been actively maintaining it. It's done otherwise as you can see.

PR #338, opened 2021-07-27
Bump ChainRulesCore

oxinabox:
I haven't looked closely to see if anything needs to be changed, lets see what CI says


CarloLucibello:
CI is not picking ChainRulesCore 1.0 since we have Zygote as a test dependence which is blocking it

oxinabox:
Zygote#master supports 1.0.
@DhairyaLGandhi  and I were thinking of waiting til after JuliaCon to tag it, just to be safe

CarloLucibello:
merge?

PR #339, opened 2021-07-29
dstsize for scatter

CarloLucibello:
Add a destination size hint for scatter + fix scatter rrule not accepting the keyword args + slim down a bit the tested types


DhairyaLGandhi:
Where was this needed?

CarloLucibello:
will be needed by GeometricFlux, otherwise there is the risk of scatter operations dropping isolated nodes

PR #340, opened 2021-07-31
Add compat entry for ChainRulesCore v1, bump version

dfdx:


dfdx:
Ah, just noticed 2 other PRs doing the same thing : D

PR #345, opened 2021-08-07
Add faster `tanh` implementation(s)

mcabbott:
This proposes to add faster, lower-precision, versions of `tanh` and friends. Motivation:
```julia
julia> @btime x * x   setup=(x=randn(Float32,100,100));
  20.250 μs (2 allocations: 39.11 KiB)  # mac M1, julia 1.7 + rosetta
  52.718 μs (2 allocations: 39.11 KiB)  # xeon E5-2603

julia> @btime y .= tanh.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  107.167 μs (2 allocations: 39.11 KiB)
  229.725 μs (2 allocations: 39.11 KiB)

julia> @btime y .= tanh_faster.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  27.125 μs (2 allocations: 39.11 KiB)
  67.623 μs (2 allocations: 39.11 KiB)
```
There has been talk of using LoopVectorization to speed this up instead. At least for the computers I tried, this lower-precision function is faster. Of course that may speed up other things, but `tanh` seem the outlier in terms of speed -- I quote times with matrix multiplication above, to illustrate that it dominates. 
```julia
julia> using LoopVectorization

julia> @btime vmap!(tanh, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  62.542 μs (2 allocations: 39.11 KiB)
  99.213 μs (2 allocations: 39.11 KiB)

julia> @btime vmap!(tanh_faster, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));  # with IfElse
  27.292 μs (2 allocations: 39.11 KiB)
  60.970 μs (2 allocations: 39.11 KiB)
```
WIP, `tanh_fast` was the initial idea, but at least for Float32, a rational approximation like `tanh_faster` seems like a better idea, roughly one less digit & another factor of 3 in speed. Both need some care not to give crazy answers at large `x`, the precise implementation could use some adjusting. Times & error measures in [this gist](https://gist.github.com/mcabbott/0acd91671894bced65209e054bf80796).

I also have no idea hot this interacts with GPU just yet.

### DIFF
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -665,6 +665,8 @@ may be wrong by 5 eps, a reduction by less than one decimal digit.
 
 For `x::Float32` this is usually about 10 times faster,
 with a smaller speedup for `x::Float64`.
+
+See also [`sigmoid_fast`](@ref).
 """
 @inline function tanh_fast(x::Float32)
     x2 = abs2(x)
@@ -689,6 +691,9 @@ tanh_fast(x::Float16) = Base.tanh(x) # Other approximations are very badly behav
     sigmoid_fast(x)
 
 This is a faster, and very slightly less accurate, version of `sigmoid`.
+For `x::Float32, perhaps 3 times faster, and maximum errors 2 eps instead of 1.
+
+See also [`tanh_fast`](@ref).
 """
 @inline function sigmoid_fast(x::Real)
     t = @fastmath exp(-abs(x))
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -694,10 +694,10 @@ end
     exp2x = @fastmath exp(x + x)
     y = (exp2x - 1) / (exp2x + 1) 
     # That has large errors near zero; using `expm1` would more accurate, but about as slow as `tanh`.
-    # Instead, we switch to an Taylor series. This is very accurate within its range:
+    # Instead, we switch to a polynomial, which is very accurate within its range, < 2 eps:
     x2 = x * x
-    ypoly = x * evalpoly(x2, (1.0, -0.33333333333333337, 0.1333333333332623, -0.0539682539194502, 0.021869476267930975, -0.00886184042142138, 0.0035188503873932893))
-    ifelse(x2 > 900.0, sign(y), ifelse(x2 < 0.02, oftype(y, ypoly), y))
+    ypoly = x * evalpoly(x2, (1.0, -0.33333333333324583, 0.13333333325511604, -0.05396823125794372, 0.02186660872609521, -0.008697141630499953))
+    ifelse(x2 > 900.0, sign(y), ifelse(x2 < 0.017, oftype(y, ypoly), y))
 end
 
 tanh_fast(x::Float16) = Base.tanh(x) # Other approximations are very badly behaved for Float16; none are fast.
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -669,6 +669,7 @@ may be wrong by 5 eps, a reduction by less than one decimal digit.
 
 For `x::Float32` this is usually about 10 times faster,
 with a smaller speedup for `x::Float64`.
+For any other number types, it just calls `tanh`.
 
 See also [`sigmoid_fast`](@ref).
 
@@ -690,7 +691,7 @@ julia> hard_tanh(0.5f0)
     ifelse(x2 < 66f0, x * (n / d), sign(x))
 end
 
-@inline function tanh_fast(x::Real)
+@inline function tanh_fast(x::Float64)
     exp2x = @fastmath exp(x + x)
     y = (exp2x - 1) / (exp2x + 1) 
     # That has large errors near zero; using `expm1` would more accurate, but about as slow as `tanh`.
@@ -700,7 +701,9 @@ end
     ifelse(x2 > 900.0, sign(y), ifelse(x2 < 0.017, oftype(y, ypoly), y))
 end
 
-tanh_fast(x::Float16) = Base.tanh(x) # Other approximations are very badly behaved for Float16; none are fast.
+# These approximations are very badly behaved for Float16; none are fast.
+# They are also a bit slower with ForwardDiff.Dual numbers, let's use Base:
+tanh_fast(x::Real) = Base.tanh(x)
 
 """
     sigmoid_fast(x)
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -4,7 +4,7 @@
 # https://github.com/JuliaGPU/CuArrays.jl/issues/614
 
 ACTIVATIONS = [
-    :σ, :hardσ, :tanh_fast, :hardtanh, :relu,
+    :σ, :hardσ, :hardtanh, :relu,
     :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu,
     :celu, :softplus, :softsign, :logσ, :logcosh,
     :mish, :tanhshrink, :softshrink, :trelu, :lisht,
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -772,21 +772,3 @@ for (f, df1, df2) in BINARY_ACTS
         return Ω, $pullback
     end
 end
-
-# For some activation functions, the gradient can be written only in terms of Ω.
-# That means that `conv_bias_act` etc. can replace `Ω = σ.(x)` with `x .= σ.(x)`
-# to save allocations. (Not so easy to opt-in to this mechanism, sadly.)
-INPLACE_ACTS = [
-    (:σ,            :(conj(Ω * (1 - Ω)))),
-    (:relu,         :(Ω > 0)),
-    (:leakyrelu,    :(ifelse(Ω > 0, one(Ω), oftf(Ω, 0.01)))),  # only the 1-argument method
-    (:hardtanh,     :(-1 < Ω < 1)),
-    (:selu,         :(deriv_selu(Ω))),
-    (:elu,          :(deriv_elu(Ω))),
-
-    (:identity,     :true),
-    (:tanh,         :(conj(1 - Ω^2))),
-    (:tanh_fast,    :(conj(1 - Ω^2))),
-    (:tanh_faster,  :(conj(1 - Ω^2))),
-    (:sigmoid_fast, :(conj(Ω * (1 - Ω)))),
-    ]
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -654,58 +654,66 @@ for f in ACTIVATIONS
 end
 
 ## Faster, less accurate, versions of some.
-## TODO: mechanism to use automatically, and worry about GPUs.
+
+"""
+    tanh_fast(x::Float32)
+
+This is a faster but less accurate version of `tanh`.
+It has error of 5eps, instead of 1.5 eps for Julia's;
+under one less decimal digit. Usually about 10x faster.
+
+`tanh_new(x::Float16)` uses the same rational approximation,
+about twice the error of Julia's, about twice the speed
+(but still 5x slower than `Float32`). Might be better to remove this.
+"""
+@inline function tanh_fast(x::Float32)
+    x2 = abs2(x)
+    n = evalpoly(x2, (1.0f0, 0.1346604f0, 0.0035974074f0, 2.2332108f-5, 1.587199f-8))
+    d = evalpoly(x2, (1.0f0, 0.4679937f0, 0.026262015f0, 0.0003453992f0, 8.7767893f-7))
+    ifelse(x2 < 66f0, x * (n / d), sign(x))
+end
+
+# tanh_fast(x::Float16) # probably better. tanh_fast(x::Real) is very badly behaved for Float16.
+
+@inline function tanh_fast(x::Float16)
+    x2 = abs2(x)
+    n = evalpoly(x2, (Float16(1.0), Float16(0.1346), Float16(0.003597), Float16(2.235e-5)))
+    d = evalpoly(x2, (Float16(1.0), Float16(0.468), Float16(0.02626), Float16(0.0003455), Float16(9.0e-7)))
+    ifelse(x2 < Float16(18), x * (n / d), sign(x))
+end
 
 """
     tanh_fast(x) = exp(2x) / (1 + exp(2x))
 
 This is faster, but slightly less accurate, version of `tanh`.
+For `x::Float64`, this has errors up to about `3e-15`, sometimes `> 10 * eps(x)`,
+and should be about 2-5 times faster.
 
-For `x::Float64`, this has errors up to about `3e-15`, sometimes `> 10 * eps(x)`.
-For `x::Float32`, errors up to about `1f-6`. About 2-5 times faster.
+But `tanh_new(x::Float32)` is more important, and a different approximation.
 """
-@inline function tanh_fast(x)
+@inline function tanh_fast(x::Real)
    exp2x = @fastmath exp(x + x)
    y = (exp2x - 1) / (exp2x + 1)
-   stop = _fast_stop(tanh_fast, x)
-   ifelse(x > stop, one(y), ifelse(x < -stop, -one(y), y))
+   ifelse(x > 30, one(y), ifelse(x < -30, -one(y), y))
 end
 # Badly behaved examples, without the stop:
 # tanh_fast(400.0), tanh_fast(60f0), tanh_fast(Float16(6)) 
-_fast_stop(::typeof(tanh_fast), x) = oftype(x, 30)
-_fast_stop(::typeof(tanh_fast), ::Float16) = Float16(5)
+
 
 """
-    tanh_faster(x)
+    sigmoid_fast(x)
 
-This is even faster, via a rational approximation, 4/5 terms, from here:
-https://math.stackexchange.com/questions/107292/rapid-approximation-of-tanhx
-
-Errors of up to about `3e-5` or `3f-5`, on Float64 & Float32 numbers.
-About 8-10 times faster. 
+Like `tanh_fast`, except for `sigmoid`. 
+This needs updating to match latest tanh version.
 """
-@inline function tanh_faster(x::Real)
-    T = float(typeof(x))
-    x2 = 4x^2
-    # y = 94.9339451088 * x * ((( x2 + 1.06315869e+03 ) * x2 + 1.88748783e+05 ) * x2 + 5.86237309e+06 ) / ((( ( x2 + 4.03183926e+03 ) * x2 + 1.64253046e+06 ) * x2 + 1.28592857e+08 ) * x2 + 1.11307745e+09 )
-    num = @fastmath T(94.9339451088) * 2x * ((( x2 + T(1.06315869e+03) ) * x2 + T(1.88748783e+05) ) * x2 + T(5.86237309e+06) )
-    den = @fastmath ((( ( x2 + T(4.03183926e+03) ) * x2 + T(1.64253046e+06) ) * x2 + T(1.28592857e+08) ) * x2 + T(1.11307745e+09) )
-    y = num / den
-    stop = _fast_stop(tanh_faster, x)
-    ifelse(x > stop, one(y), ifelse(x < -stop, -one(y), y))
-end
-_fast_stop(::typeof(tanh_faster), x) = oftype(x, 10)
-
 @inline function sigmoid_fast(x::Real)
    s = @fastmath exp(x)
    y = s / (s + 1)
    stop = _fast_stop(sigmoid_fast, x)
-   ifelse(x > stop, one(y), ifelse(x < -stop, zero(y), y))
+   ifelse(x > 60, one(y), ifelse(x < -stop, zero(y), y))
 end
-_fast_stop(::typeof(sigmoid_fast), x) = oftype(x, 60)
-_fast_stop(::typeof(sigmoid_fast), ::Float16) = Float16(10)
 
-@inline function sigmoid_faster(x::Real)
+@inline function sigmoid_fast(x::Float32)
     T = float(typeof(x))
     x2 = x^2
     num = @fastmath T(47.4669725544) * x * ((( x2 + T(1.06315869e+03) ) * x2 + T(1.88748783e+05) ) * x2 + T(5.86237309e+06) )
@@ -713,6 +721,7 @@ _fast_stop(::typeof(sigmoid_fast), ::Float16) = Float16(10)
     T(0.5) + num / den  # this is really hacked together
 end
 
+sigmoid_fast(x::Float16) = sigmoid(x)  # sigmoid_fast is extremely badly behaved at large x
 
 ## Define rrules for some activation functions, along with the
 ## broadcasted rrule activation functions.
@@ -732,7 +741,6 @@ UNARY_ACTS = [ # f, df
     (:softplus,     :(σ(x))),
 
     (:tanh_fast,    :(conj(1 - Ω^2))),
-    (:tanh_faster,  :(conj(1 - Ω^2))),
     (:sigmoid_fast, :(conj(Ω * (1 - Ω)))),
     ]
 
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -3,12 +3,13 @@
 # Some of activation functions have its wrapper function for GPU in NNlibCUDA.jl.
 # https://github.com/JuliaGPU/CuArrays.jl/issues/614
 
-const ACTIVATIONS = 
-    [:σ, :hardσ, :hardtanh, :relu, 
-    :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu, 
-    :celu, :softplus, :softsign, :logσ, :logcosh, 
-    :mish, :tanhshrink, :softshrink, :trelu, 
-    :lisht]
+ACTIVATIONS = [
+    :σ, :hardσ, :tanh_fast, :hardtanh, :relu,
+    :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu,
+    :celu, :softplus, :softsign, :logσ, :logcosh,
+    :mish, :tanhshrink, :softshrink, :trelu, :lisht,
+    :tanh_fast, :tanh_faster, :sigmoid_fast,
+    ]
 
 for f in ACTIVATIONS
     @eval export $(f)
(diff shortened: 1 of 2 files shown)


src/activations.jl
```diff
@@ -75,7 +76,7 @@ Leaky [Rectified Linear Unit](https://en.wikipedia.org/wiki/Rectifier_(neural_ne
 activation function.
 You can also specify the coefficient explicitly, e.g. `leakyrelu(x, 0.01)`.
 """
-leakyrelu(x, a=oftf(x, 0.01)) = max(a * x, x)
+leakyrelu(x, a=oftf(x, 0.01)) = ifelse(x>0, float(x), oftf(x, a*x))  # max(a*x, x) is 3x slower
```
mcabbott, line 79: https://github.com/FluxML/NNlib.jl/pull/345/commits/e6befa18f6ce5d1d71d776159eb7b43a3120253e is really orthogonal, does this:
```julia
# Unrelated, improve relu slightly:
julia> x = randn(Float32, 10^3); y = similar(x);
julia> relu_max(x) = max(zero(x), x);
julia> relu_if(x) = ifelse(x>0, x, zero(x));
julia> x = randn(Float32, 10^3); y = similar(x);
julia> @btime $y .= relu_max.($x);
  110.752 ns (0 allocations: 0 bytes)
julia> @btime $y .= relu_if.($x);
  92.354 ns (0 allocations: 0 bytes)
  
# and improve leakyrelu quite a bit
julia> oftf(x, y) = oftype(float(x), y);
julia> leakyrelu_max(x, a=oftf(x, 0.01)) = max(a * x, x);
julia> leakyrelu_if(x, a=oftf(x, 0.01)) = ifelse(x>0, float(x), oftf(x, a*x))
leakyrelu_if (generic function with 2 methods)
julia> @btime $y .= leakyrelu_max.($x);
  558.396 ns (0 allocations: 0 bytes)
julia> @btime $y .= leakyrelu_if.($x);
  164.928 ns (0 allocations: 0 bytes)
  ```

Comment (ToucheSir) [reply src/activations.jl:79]: How does perf look without requiring `float(x)`? e.g. for mixed, lower or higher precision.

Comment (mcabbott) [reply src/activations.jl:79]: I put `float(x)` there to match the other branch, as `ifelse` doesn't promote, but `max` does. Without the float, `@code_warntype leakyrelu_if(1)` is bad.

Both seem type-stable with anything I could think of, e.g. `@code_warntype leakyrelu_if(Float16(1), 0.3)`; I'm not entirely sure why `leakyrelu_max` is so much slower.

**Comment by DhairyaLGandhi** [comment]
We have been discussing a more structured approach to this lead by @chriselrod and NNlibCPU.

File: src/activations.jl (hunk 1 above)
Comment (ToucheSir) [reply src/activations.jl:79]: Right, my point was this means `typeof(leakyrelu(Float16(1))) == typeof(leakyrelu(Float64(1)))`. Namely one gets promoted and the other reduced.

Comment (mcabbott) [reply src/activations.jl:79]: No, it preserves Float16:
```
julia> typeof(leakyrelu_if(Float16(1))) == typeof(leakyrelu_if(Float64(1)))
false

julia> typeof(leakyrelu_max(Float16(1))) == typeof(leakyrelu_max(Float64(1)))
false

julia> leakyrelu_if(Float16(1))
Float16(1.0)

julia> leakyrelu_if(Float16(-1))
Float16(-0.01)

julia> leakyrelu_if(Float16(1), 3//4)
Float16(1.0)

julia> leakyrelu_if(Float16(-1), 3//4)
Float16(-0.75)
```

Comment (ToucheSir) [reply src/activations.jl:79]: I wasn't aware `float` didn't promote `Float_` types, interesting. As for the lowering, it carries over into llvm as well:
```julia
julia> @code_llvm leakyrelu_if(Float16(1))
;  @ REPL[12]:1 within `leakyrelu_if`
define half @julia_leakyrelu_if_403(half %0) #0 {
top:
;  @ REPL[12]:1 within `leakyrelu_if` @ REPL[12]:1
; ┌ @ operators.jl:378 within `>`
; │┌ @ float.jl:511 within `<` @ float.jl:444
    %1 = fpext half %0 to float
    %2 = fcmp ule float %1, 0.000000e+00
; └└
; ┌ @ float.jl:403 within `*`
   %3 = fmul float %1, 0x3F847C0000000000
   %4 = fptrunc float %3 to half
; └
  %5 = select i1 %2, half %4, half %0
;  @ REPL[12]:1 within `leakyrelu_if`
  ret half %5
}

julia> @code_llvm leakyrelu_if(Float32(1))
;  @ REPL[12]:1 within `leakyrelu_if`
define float @julia_leakyrelu_if_412(float %0) #0 {
top:
;  @ REPL[12]:1 within `leakyrelu_if` @ REPL[12]:1
; ┌ @ operators.jl:378 within `>`
; │┌ @ float.jl:490 within `<` @ float.jl:443
    %1 = fcmp ule float %0, 0.000000e+00
; └└
; ┌ @ float.jl:404 within `*`
   %2 = fmul float %0, 0x3F847AE140000000
; └
  %3 = select i1 %1, float %2, float %0
;  @ REPL[12]:1 within `leakyrelu_if`
  ret float %3
}
```
As well as mentioned in https://github.com/JuliaLang/julia/blob/bb5b98e72a151c41471d8cc14cacb495d647fb7f/src/llvm-demote-float16.cpp#L38.

Using `@fastmath` saves on one truncation:
```julia
julia> leakyrelu_if_fast(x, a=oftf(x, 0.01)) = @fastmath ifelse(x>0, float(x), oftf(x, a*x));

julia> @code_llvm leakyrelu_if_fast(Float16(1))
;  @ REPL[15]:1 within `leakyrelu_if_fast`
define half @julia_leakyrelu_if_fast_378(half %0) #0 {
top:
;  @ REPL[15]:1 within `leakyrelu_if_fast` @ REPL[15]:1
; ┌ @ operators.jl:378 within `>`
; │┌ @ float.jl:511 within `<` @ float.jl:444
    %1 = fpext half %0 to float
    %2 = fcmp ule float %1, 0.000000e+00
; └└
; ┌ @ fastmath.jl:167 within `mul_fast`
   %3 = fmul fast half %0, 0xH211F
; └
  %4 = select i1 %2, half %3, half %0
;  @ REPL[15]:1 within `leakyrelu_if_fast`
  ret half %4
}

julia> @code_llvm leakyrelu_if_fast(Float32(1))
;  @ REPL[15]:1 within `leakyrelu_if_fast`
define float @julia_leakyrelu_if_fast_380(float %0) #0 {
top:
;  @ REPL[15]:1 within `leakyrelu_if_fast` @ REPL[15]:1
; ┌ @ operators.jl:378 within `>`
; │┌ @ float.jl:490 within `<` @ float.jl:443
    %1 = fcmp ule float %0, 0.000000e+00
; └└
; ┌ @ fastmath.jl:167 within `mul_fast`
   %2 = fmul fast float %0, 0x3F847AE140000000
; └
  %3 = select i1 %1, float %2, float %0
;  @ REPL[15]:1 within `leakyrelu_if_fast`
  ret float %3
}
```

mcabbott: Interesting. Both are super-slow on Float16 though. 
```
julia> x = randn(Float16, 10^3); y = similar(x);

julia> @btime $y .= leakyrelu_max.($x);
  29.583 μs (0 allocations: 0 bytes)

julia> @btime $y .= leakyrelu_if.($x);
  7.708 μs (0 allocations: 0 bytes)

julia> @btime $y .= leakyrelu_if_fast.($x);
  7.729 μs (0 allocations: 0 bytes)
```

ToucheSir: Makes sense to me, the native asm (at least on x86) generated is far more complex for the Float16 path. Probably something to do with hardware support.

chriselrod:
I could add `tanh_faster` [here](https://github.com/FluxML/NNlibCPU.jl/pull/1).
I wont be able to make the ML meeting this week, as I have a vet appointment at that time. But I'll try and merge that PR with basic dense, conv, and batch layers working.

chriselrod:
Also, that Xeon in your benchmark seems to be very slow. Using:
```julia
using LoopVectorization, BenchmarkTools, LinearAlgebra
Sys.ARCH === :x86_64 && using MKL
BLAS.set_num_threads(1)
using IfElse: ifelse
@inline function tanh_faster(x::Real)
  T = float(typeof(x))
  x2 = 4x^2
  # y = 94.9339451088 * x * ((( x2 + 1.06315869e+03 ) * x2 + 1.88748783e+05 ) * x2 + 5.86237309e+06 ) / ((( ( x2 + 4.03183926e+03 ) * x2 + 1.64253046e+06 ) * x2 + 1.28592857e+08 ) * x2 + 1.11307745e+09 )
  num = @fastmath T(94.9339451088) * 2x * ((( x2 + T(1.06315869e+03) ) * x2 + T(1.88748783e+05) ) * x2 + T(5.86237309e+06) )
  den = @fastmath ((( ( x2 + T(4.03183926e+03) ) * x2 + T(1.64253046e+06) ) * x2 + T(1.28592857e+08) ) * x2 + T(1.11307745e+09) )
  y = num / den
  stop = _fast_stop(tanh_faster, x)
  ifelse(x > stop, one(y), ifelse(x < -stop, -one(y), y))
end
_fast_stop(::typeof(tanh_faster), x) = oftype(x, 10)

@btime x * x   setup=(x=randn(Float32,100,100));
@btime y .= tanh.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
@btime y .= tanh_faster.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
@btime vmap!(tanh, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
@btime vmap!(tanh_faster, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));  # with IfElse
```
I am getting:
```julia
julia> @btime x * x   setup=(x=randn(Float32,100,100));
  16.220 μs (2 allocations: 39.11 KiB) # i7-1165G7 (laptop)
  11.967 μs (2 allocations: 39.11 KiB) # i9-10980XE (HEDT)
  25.125 μs (2 allocations: 39.11 KiB) # M1 native

julia> @btime y .= tanh.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  75.780 μs (2 allocations: 39.11 KiB) # i7-1165G7 (laptop)
  94.014 μs (2 allocations: 39.11 KiB) # i9-10980XE (HEDT)
  83.917 μs (2 allocations: 39.11 KiB) # M1 native

julia> @btime y .= tanh_faster.(x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  19.565 μs (2 allocations: 39.11 KiB) # i7-1165G7 (laptop)
  15.447 μs (2 allocations: 39.11 KiB) # i9-10980XE (HEDT)
  28.417 μs (2 allocations: 39.11 KiB) # M1 native

julia> @btime vmap!(tanh, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));
  27.560 μs (2 allocations: 39.11 KiB) # i7-1165G7 (laptop)
  21.269 μs (2 allocations: 39.11 KiB) # i9-10980XE (HEDT)
  45.625 μs (2 allocations: 39.11 KiB) # M1 native

julia> @btime vmap!(tanh_faster, y, x * x)   setup=(x=randn(Float32,100,100); y=similar(x));  # with IfElse
  19.097 μs (2 allocations: 39.11 KiB) # i7-1165G7 (laptop)
  13.739 μs (2 allocations: 39.11 KiB) # i9-10980XE (HEDT)
  28.708 μs (2 allocations: 39.11 KiB) # M1 native
```

mcabbott:
This was supposed to be a super-quick fix, but floating point stuff is messy!

Agree the nicest possible outcome would be a super-low-precision library. I think that for most ML purposes a smooth function which is <1pixel wrong on a graph is going to be plenty. My guess is that smoothness is more important than worst-case error, i.e. a steps are worse than ramps. NNlib might not be a terrible place for such things to live, though, if it's just `tanh` and some friends -- it's supposed to be a collection of NN-ish functions you can use for other purposes. They could move out once someone finds a need & time write other functions.

NNlibCPU.jl looks more ambitious. From very informal investigation there's lots of room for faster CPU `conv` etc.

If that's an optional add-on, then the argument for `tanh_faster` not being in there is that it might be useful without that. Motivated by things like [this thread](https://discourse.julialang.org/t/flux-vs-pytorch-cpu-performance/42667) -- it seems embarrassing that in the most obvious casual benchmark, 80% of the time is in `tanh`.

oscardssmith:
One thing to try is using exp2 instead of exp. It may lose a tiny bit of precision, but be notably faster. exp needs some extra work to do the reduction, while exp2 gets it for free.

ToucheSir:
Is PyTorch doing something extra besides using `tanh(f)`? If not, why would there still be a discrepancy between libm and https://github.com/JuliaLang/julia/pull/38382?

mcabbott:
I never looked inside that PR somehow. The times for `tanh` will depend a lot on which branches you hit. And the time for `tanh_kernel` suggests that my rational `tanh_faster` has room for improvement:
```julia
julia> @btime y .= tanh.(x)  setup=(x = randn(Float32, 10^3); y = similar(x););  # as above
  2.963 μs (0 allocations: 0 bytes)

julia> @btime y .= tanh.(x)  setup=(x = 0.1f0 .* randn(Float32, 10^3); y = similar(x););  # small => kernel
  1.129 μs (0 allocations: 0 bytes)

julia> @btime y .= tanh.(x)  setup=(x = 100 .* randn(Float32, 10^3); y = similar(x););  # large => flat
  627.491 ns (0 allocations: 0 bytes)

julia> @btime y .= Base.Math.tanh_kernel.(x)  setup=(x = rand(Float32, 10^3); y = similar(x););
  144.404 ns (0 allocations: 0 bytes)

julia> Base.Math.TANH_SMALL_X(Float32)  # polynomial below this
1.3862944f0

julia> Base.Math.TANH_LARGE_X(Float32)  # flat above this, but why so large?
18.0f0

julia> tanh(9f0)  # already constant at half that.
1.0f0
```

oscardssmith:
Just wondering, how did you generate the coefficients? Theoretically, Remez.jl can make minimax rational approximations, but I've never had much luck (I believe due to problems with numerical conditioning).

mcabbott:
I copied them from here: https://math.stackexchange.com/questions/107292/rapid-approximation-of-tanhx

mcabbott:
I never tried Remez but here's a quick attempt to find a function to cover the gap from the kernel to the flat (which presently calls `exp`). It indeed seems finicky as to what orders it can solve for, maybe there are tricks to make it behave better. Are there tricks for how to round the coefficients it finds to Float32?
```julia
using Remez, Plots
N,D,E,X = ratfn_minimax(tanh, (big"1.38", big"9.0"), 7, 4); Float32(E) # 1.2949664f-8
f(x) = evalpoly(x, N) / evalpoly(x, D)
xs = big.(1.3862944:0.01:9); plot(xs, Float64.(f.(xs) .- tanh.(xs))) # error in approx
E / eps(1f0) # 0.1
Tuple(Float32.(N))
Tuple(Float32.(D))
@inline function f74(x::Float32)
    n = evalpoly(x, (0.01731622f0, 0.9280657f0, 0.21158579f0, 0.08456047f0, 0.031536844f0, 0.00071064394f0, -3.1970143f-5, 6.226407f-7))
    d = evalpoly(x, (1.0f0, 0.08685698f0, 0.52843076f0, 0.016710332f0, 0.040428344f0))
    n / d
end
xs = 1.3862944:0.01:9; plot!(xs, (f74.(Float32.(xs)) .- tanh.(xs))) # error with Float32, up to 3e-7
3e-7 / eps(tanh(7.2f0)) # 5

using Base.Math: TANH_LARGE_X, TANH_SMALL_X, tanh_kernel
@inline function tanh_ifelse(x::T) where T<:Union{Float32}
    abs2x = abs(2x)
    ifelse(abs2x >= TANH_LARGE_X(T), copysign(one(T), x),
        ifelse(abs2x <= TANH_SMALL_X(T), x*tanh_kernel(x*x), 
            copysign(f74(abs(x)), x) ))  # clearly should re-use abs2x
end
xs = -10:0.01f0:10; plot(xs, tanh_ifelse.(xs)); plot!(xs, tanh.(xs))  # sanity check

@btime y .= tanh_ifelse.(x)  setup=(x = randn(Float32, 10^3); y = similar(x););
# 596.369 ns (0 allocations: 0 bytes)
# 3.109 μs (0 allocations: 0 bytes)  with branches instead of ifelse

# Accuracy of tanh_faster rational: 4e-7 until 8, and 1 - tanh(8f0) < 3e-7
xs = 0:0.001:8; plot(xs, @. Float64(abs(tanh_faster(Float32(xs)) - tanh(big(xs)))))
```

tanh_errors

mcabbott:
Ok, I delete some details, they are in the gist. But after messing with Remez.jl a bit, I think this function is within 5 eps of true, abs error 2.929f-7. It's a little faster than my initial `tanh_faster` (which is a similar rational function) which was much less accurate, 300eps, abs error 3.629f-5.
```julia
function tanh_new(x::Float32)
    x2 = abs2(x)
    n = evalpoly(x2, (1.0f0, 0.1346604f0, 0.0035974074f0, 2.2332108f-5, 1.587199f-8))
    d = evalpoly(x2, (1.0f0, 0.4679937f0, 0.026262015f0, 0.0003453992f0, 8.7767893f-7))
    ifelse(x2 < 66f0, x * (n / d), sign(x))
end
```
When I try things on the GPU, I believe Julia's `tanh` is not being used, but the function used has only slightly larger errors. If you use this `tanh_new` you seem to get almost identical accuracy to using it on the CPU. I can't measure a time for any of these, they all take exactly as long as broadcasting `identity`.

oscardssmith:
There are some papers on how to make the remez algorithm more numerically stable, and on optimal minimax polynomials with floating point coefficients, but neither have been implimented yet (both are on my list of things to do if I have time). The other thing that should be relatively easy to add is ways of specifying fixed values for some coefficients. This is useful for generating even/odd polynomials, as well as ones where the first few terms can be represented without rounding

src/activations.jl
mcabbott: Moved to #347 now.

src/activations.jl
```diff
@@ -237,65 +237,74 @@ for f in ACTIVATIONS
...
+and should be about 2-5 times faster.
 
-For `x::Float64`, this has errors up to about `3e-15`, sometimes `> 10 * eps(x)`.
-For `x::Float32`, errors up to about `1f-6`. About 2-5 times faster.
+But `tanh_new(x::Float32)` is more important, and a different approximation.
 """
-@inline function tanh_fast(x)
+@inline function tanh_fast(x::Real)
    exp2x = @fastmath exp(x + x)
    y = (exp2x - 1) / (exp2x + 1)
```
oscardssmith, line 279: You should use expm1 here. It will be about 30% slower, but you will get correct answers for small x. This implimentation has catestrophic cancellation.

mcabbott: Probably! Agree this looks dodgy, will re-check Float64 times & accuracy at some point.

I also wonder if there's a clever formula for `tanh(x)/x`, since it seems a little odd to get accuracy near 0 by expm1 when the desired result is close to linear?

oscardssmith: doesn't look like it.

oscardssmith:
Any news on this?

src/activations.jl
mcabbott: Switching to `expm1` seems to cost me the entire speed gain:
```
julia> xs = collect(0.01:0.001:10); ys = similar(xs);

julia> @btime @. $ys = tanh($xs);
  44.291 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh_fast($xs);  # with expm1
  47.958 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh_fast($xs);  # with exp(2x)-1
  11.500 μs (0 allocations: 0 bytes)
```
Accuracy, on this same range of points, improves 65eps -> 1.7eps relative worst case, but no change in absolute worst case 1.6e-16. So, as you say, the error is all for very small x.

Maybe it can evaluate some polynomial for `abs(x) < 0.13` or so, if 4eps is the goal.

src/activations.jl
```diff
@@ -256,32 +255,17 @@ about twice the error of Julia's, about twice the speed
...
 @inline function tanh_fast(x::Real)
-   exp2x = @fastmath exp(x + x)
-   y = (exp2x - 1) / (exp2x + 1)
-   ifelse(x > 30, one(y), ifelse(x < -30, -one(y), y))
+    exp2x = @fastmath exp(x + x)
+    y = (exp2x - 1) / (exp2x + 1) 
+    # That has bad errors near zero; using expm1 would be slower & more accurate.
+    # Instead, we switch to an taylor series; seems to add about 50% to time.
+    x2 = x * x
+    ypoly = x * evalpoly(x2, (0.9999999999999999, -0.33333333333309806, 0.13333333318143492, -0.053968217983868146 , 0.021865628148606587, -0.008671836868790176))
```
oscardssmith, line 264: How do these coeffs do? `(1.0, 0.33333333333333337, 0.1333333333332623, -0.0539682539194502, 0.021869476267930975, -0.00886184042142138, 0.0035188503873932893)` In general, I've found that forcing the leading coefficients to exact values can help with error since you remove the error in the coeficient. These were generated with the following code.
```
coefs = [1/big(3) , 2/big(15) ,- 17/big(315) , 62/big(2835) ,- 1382/big(155925) , 21844/big(6081075) ,- 929569/big(638512875) , 6404582/big(10854718875) ,- 443861162/big(1856156927625)]

julia> g(x)=evalpoly(x, coefs)

julia> Tuple(Float64.(ratfn_minimax(g, (1e-15,.017^2),6,0)[1]))
```

src/activations.jl
```diff
@@ -256,32 +255,17 @@ about twice the error of Julia's, about twice the speed
...
-For `x::Float64`, this has errors up to about `3e-15`, sometimes `> 10 * eps(x)`,
-and should be about 2-5 times faster.
-
-But `tanh_new(x::Float32)` is more important, and a different approximation.
-"""
 @inline function tanh_fast(x::Real)
-   exp2x = @fastmath exp(x + x)
-   y = (exp2x - 1) / (exp2x + 1)
-   ifelse(x > 30, one(y), ifelse(x < -30, -one(y), y))
+    exp2x = @fastmath exp(x + x)
```
chriselrod, line 259: What about
```julia
 @inline exp_inline(x) = Base.Math.exp_impl(x, Val(:ℯ))
```
and defining an associated diff rule, for the versions of Julia that define an inline `Base.Math.exp_impl`?

This should let the compiler SIMD the code.

src/activations.jl
mcabbott: That is better (with -0.3333 surely), indeed. In fact this looks to be within 1eps within the range for which I used the polynomial:
```
julia> maximum(x -> abs(tanh(big(x)) - tanh_fast(x)) / eps(Float64(tanh(big(x)))), 0.000_01:0.000_01:0.13) |> Float64  0.9878056794145185  # with yours
1.8541809732394265  # with mine
```
Whereas the point which is 5eps is somewhere `0.2<x<0.3`. So perhaps the ideal changeover point is quite a bit larger... I think 0.13 was a very quick guess at where the error got to 4eps or so. I'm not being maximally systematic about this, clearly.

src/activations.jl
mcabbott: I saw this in the Zulip thread and had a go... but for me this is somewhere between the `@fastmath` variant and the plain `exp` in speed:
```
julia> @btime @. $ys = tanh_fast($xs);  # with @fastmath exp(x + x), and polynomial etc, M1 + Rosetta
  30.125 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh_fast($xs);  # with exp_inline
  47.208 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh_fast($xs);  # with Base.exp
  85.125 μs (0 allocations: 0 bytes)
```

oscardssmith: What about `Base.Math.exp_impl_fast(x, Val(:ℯ))`

oscardssmith: (I think that needs 1.7)

mcabbott: Using `Base.Math.exp_impl_fast` seems to make no difference at all, strangely? I don't see an `@inline`.

src/activations.jl
```diff
@@ -235,22 +236,72 @@ for f in ACTIVATIONS
...
+"""
+    sigmoid_fast(x)
+
+This is a faster, and very slightly less accurate, version of `sigmoid`.
+"""
+@inline function sigmoid_fast(x::Real)
+    t = @fastmath exp(-abs(x))
+    y = ifelse(x ≥ 0, inv(1 + t), t / (1 + t))
+    ifelse(x > 40, one(y), ifelse(x < -80, zero(y), y))
+end
```
mcabbott, line 280: For Float32, this `sigmoid_fast` isn't as big an improvement as `tanh_fast`. That rational function symmetric about zero gives you terrible relative accuracy for negative `x`, once it's going to zero. 
```
julia> xs = collect(0.01:0.001:10); ys = similar(xs);

julia> xs_32 = Float32.(collect(xs)); ys_32 = similar(xs_32);

# Float32

julia> @btime @. $ys_32 = sigmoid($xs_32);
  26.250 μs (0 allocations: 0 bytes)

julia> @btime @. $ys_32 = sigmoid_fast($xs_32);
  7.135 μs (0 allocations: 0 bytes)

julia> @btime @. $ys_32 = sigmoid_fast_old($xs_32);  # version with polynomial from tanh_fast
  3.870 μs (0 allocations: 0 bytes)

# Float64

julia> @btime @. $ys = sigmoid($xs);
  34.250 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = sigmoid_fast($xs);
  12.000 μs (0 allocations: 0 bytes)
```
Compare tanh:
```
julia> @btime @. $ys_32 = tanh($xs_32);
  27.125 μs (0 allocations: 0 bytes)

julia> @btime @. $ys_32 = tanh_fast($xs_32);
  3.594 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh($xs);
  44.958 μs (0 allocations: 0 bytes)

julia> @btime @. $ys = tanh_fast($xs);
  17.041 μs (0 allocations: 0 bytes)
```

src/activations.jl
oscardssmith: The other option is to lower the degree polynomial. The one I gave is a minimax polynomial, so it won't be anywhere near optimal outside the range. Do you have an error threshold in mind? If so, I'll do some searching to find the minimal degree one with the required accuracy of the range.

mcabbott: I vaguely had 5eps in mind, just because that was doable for the Float32 rational function --- there I could fit more more terms, but after rounding etc. couldn't do much better. I'm not sure how much NNlib cares about the Float64 case, it might just be for our amusement. But sure, if you are set up to search then a better polynomial + threshold would be welcome. 

BTW I tried replacing `exp(x+x)` with `exp2(α x)` as suggested, but didn't see much change in speed. The accuracy does seem a little worse on one side, so perhaps it wants to be `exp(abs(2x))` & restore the sign later.

oscardssmith: So, it looks like to maintain 5 ULPs, we need a polynomial for `|x| < 0.063`. Given that, we can lower the degree to 4, which gives coefficients of `(1.0,  -0.3333333333331805,  0.13333333295910244, -0.05396796907132604,  0.02178397571543739)` which have a maximal error of 1.2 ULPs in the range. Alternatively, bumping up to degree 5 lets the polynomial stay accurate to 2.5 ULPs over the whole range with coefficients `(1.0, -0.33333333333324583, 0.13333333325511604, -0.05396823125794372, 0.02186660872609521, -0.008697141630499953)`. I would probably go for this version, since the 1 extra fma won't change performance too much, and will give 2x less accuracy, which imo is worth it.

CarloLucibello:
Where are we here?

mcabbott:
Status is that I drifted off. When last I was fiddling, I was trying to use Oscar's numbers here https://github.com/FluxML/NNlib.jl/pull/345#discussion_r698043287 but IIRC got terrible accuracy & wasn't sure why. It was nearly done, though.

oscardssmith:
my numbers were minimax for a specific range, so if you were using a different range, they would be all wrong.

mcabbott:
OK, b53155f changes to use these coefficients. I may well have messed up the ranges last time. Re-reading today, I'm not entirely sure what range these are for, `x^2 < 0.017` comes from trying it out, and seems to keep within 5 eps.

Something I don't think I checked until today, if I time `@btime ForwardDiff.derivative(tanh, x[])  setup=(x=Ref(rand(Float32)));` they are a bit slower. I guess ideally they would have their own rules somewhere. (For ChainRules there are rules here, including broadcasting.)

oscardssmith:
Those coefs with that range (you got the range right) should get you 2.5 ULP if I'm doing the math right.

mcabbott:
OK, great. I don't see more than 2, at `0.129701`, inside the range. But outside the polynomial's range, 5 at `0.140901`, `-0.206101`. This seems fine really, it does what we wanted.

And the main goal anyway is the Float32 version.

oscardssmith:
Ah, I'd done the testing assuming the regular (not fastmath) versions.

mcabbott:
Oh right, with that I get worst 2 everywhere. But it's no longer obviously faster than `tanh`:
```
julia> xs = collect(0.01:0.001:10); ys = similar(xs);

julia> @btime @. $ys = tanh($xs);
  98.333 μs (0 allocations: 0 bytes)             # M1 + rosetta + 1.7
  min 44.916 μs, mean 45.177 μs (0 allocations)  # M1 native + 1.8

julia> @btime @. $ys = tanh_fast($xs);  # without @fastmath
  90.416 μs (0 allocations: 0 bytes)
  min 48.083 μs, mean 48.747 μs (0 allocations)

julia> @btime @. $ys = tanh_fast($xs);  # with @fastmath
  30.375 μs (0 allocations: 0 bytes)
  min 33.167 μs, mean 33.580 μs (0 allocations)
  ```

**Comment by oscardssmith** [comment]
no, you definitely want the fastmath version, it just changes the tradeoff a little. I'm looking now to see if I can quickly get a better parameter tune given the tradeoff.

**Comment by oscardssmith** [comment]
After a little experimenting, it appears that the coefficients were actually pretty good. Lowering the error further isn't really possible (since the fastmath version needs to go up to `x=2.55` for the error to go down noticeably (which would take 2 extra coefficients to match the accuracy), and you can't remove a coefficient without significantly raising the error.

**Comment by mcabbott** [comment]
OK, thanks for checking. Maybe we should call this ready? 

My last change is to restrict this polynomial story to Float64, so that if anyone tries it with dual numbers or bigfloats or something, they will get ordinary tanh. 

We can speed up other function in this file by large factors, too, e.g. `elu_fast, gelu_fast, swish_fast`. But maybe they don't need to have slow versions at all. And maybe nobody uses them anyway? That can wait for another PR.

File: src/activations.jl [hunk 6]
```diff
@@ -3,12 +3,13 @@
...
-    :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu, 
-    :celu, :softplus, :softsign, :logσ, :logcosh, 
-    :mish, :tanhshrink, :softshrink, :trelu, 
-    :lisht]
+ACTIVATIONS = [
+    :σ, :hardσ, :tanh_fast, :hardtanh, :relu,
+    :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu,
+    :celu, :softplus, :softsign, :logσ, :logcosh,
+    :mish, :tanhshrink, :softshrink, :trelu, :lisht,
+    :tanh_fast, :sigmoid_fast,
```
CarloLucibello, line 11:
```suggestion
    :σ, :hardσ, :hardtanh, :relu,
    :leakyrelu, :relu6, :rrelu, :elu, :gelu, :swish, :selu,
    :celu, :softplus, :softsign, :logσ, :logcosh,
    :mish, :tanhshrink, :softshrink, :trelu, :lisht,
    :tanh_fast, :sigmoid_fast,
```

CarloLucibello (approved):

CarloLucibello:
Should we

oscardssmith:
My guess is people do use gelu since that appears to be what Transformer models are moving towards.

Issue #333, opened 2021-07-13
`batched_mul!` is not exported


bjarthur:
... as i've been [told](https://github.com/FluxML/NNlibCUDA.jl/issues/11#issuecomment-862891946) it was.

```
julia> using NNlib

julia> batched_    # tab complete
batched_adjoint   batched_mul        batched_transpose  batched_vec

julia> NNlib.batched_   # tab complete
batched_adjoint      batched_mul           batched_mul_generic!  batched_vec
batched_gemm!        batched_mul!          batched_transpose
```

mcabbott:
Apparently not, sorry! Should be easy to fix. 

While there I see there's also no `batched_vec!` at all.

PR #348, opened 2021-08-16
export batched_mul!

bjarthur:
fixes https://github.com/FluxML/NNlib.jl/issues/333


bjarthur:
@mcabbott

DhairyaLGandhi:
Thank you, I've started ci on it. Do we also care about a `batched_vec`? We can and probably should reuse the batched ops directly.

ToucheSir:
`batched_vec` is already exported, I think the question in #333 was whether there should be an in-place implementation of that as well.

DhairyaLGandhi:
Ah, of course. Just checked that it's exported.

PR #350, opened 2021-09-07
simplify scatter mean

CarloLucibello:


src/scatter.jl
```diff
@@ -148,10 +148,22 @@ end
...
+    print("START")
+    counts = fill!(similar(Δ, Int, size(Δ)[end-M+1:end]), 0)
+    scatter!(+, counts, fill!(similar(idx), 1), idx)
+    den = gather(counts, idx)
+    
+    # make num and den broadcast compatible
+    for i in 1:ndims(num)-ndims(den)
+        den = unsqueeze(den)
+    end
+    return safe_div.(num, den)
```
DhairyaLGandhi, line 163: `save_div` looks odd, is there precedent to doing this sort of division?

CarloLucibello: it's already used in the scatter-mean forward pass. Here it's needed to correctly backpropagate zeros instead of infs

DhairyaLGandhi: Could you link to where pytorch/ tf does this?

CarloLucibello: I'm not familiar with those codebases

CarloLucibello:
gpu tests not passing is expected, NNlibCUDA is relying on an internal function deleted here that is not needed in or in NNlibCUDA anymore. Will file a PR fixing NNlibCUDA

CarloLucibello:
btw, stuff like this that have to coordinate NNlib and NNlibCUDA PRs have been very annoying lately, I think we should revisit #300

DhairyaLGandhi:
Both of them would need updating anyway, and tests can be run with branches (which is the cleaner approach). Either way, best to have an answer about the division before evaluating the PR.

src/scatter.jl
maxfreu: Just as a side note: the most relevant pytorch code is in these directories:

https://github.com/pytorch/pytorch/tree/master/aten/src/ATen/native
https://github.com/pytorch/pytorch/tree/master/aten/src/ATen/native/cpu
https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cpu/ScatterGatherKernel.cpp

PR #353, opened 2021-09-30
add topk

### LINKED ISSUE #352: add topk / maxk
We should add what is called elsewhere `topk` or `maxk`, e.g. https://pytorch.org/docs/stable/generated/torch.topk.html

In Base we have `partialsort`  and `partialsortperm` doing something similar but limited to 1d arrays

CarloLucibello:
Initial steps to fix #352 

TODO
- [ ] settle for an interface
- [ ] add rrule or write it in an AD friendly way

### DIFF
--- a/src/pooling.jl
+++ b/src/pooling.jl
@@ -177,3 +177,33 @@ for pool in [:maxpool, :meanpool]
         return Ω, $pullback
     end
 end
+
+
+function topk(x::AbstractArray{T,N}, k; rev=false, dims=nothing) where {T,N}
+    if dims === nothing
+        y = vec(x)
+        perm = partialsortperm(y, 1:k; rev)
+        return y[perm], linear_to_cartesian(x, perm)
+    else
+        @assert dims isa Int
+        sz1 = size(x)[1:dims-1]
+        sz2 = size(x)[dims+1:end]
+        slice1 = CartesianIndices(sz1)
+        slice2 = CartesianIndices(sz2)
+        perm = similar(x, Int, (sz1..., k, sz2...))
+        y = similar(x, (sz1..., k, sz2...))
+        for I1 in slice1
+            for I2 in slice2
+                xI = x[I1,:,I2]
+                permI = partialsortperm(x[I1,:,I2], 1:k; rev)
+                perm[I1,:,I2] .= permI
+                y[I1,:,I2] .= xI[permI]
+            end
+        end
+        return y, perm
+    end
+end
+
+function linear_to_cartesian(x, i)
+    CartesianIndices(x)[i]
+end


mcabbott:
I think the obvious way to make this AD-able would be just rely on the gradient for getindex, which will store the indices for the backwards pass. Maybe it's as simple as this:
```
topk(x::AbstractArray, k::Integer; kw...) = x[topkind(x, k; kw...)]

topkind(x::AbstractVector, k::Integer; dims::Integer=1, rev=true, kw...) = (@assert dims==1; partialsortperm(x, 1:k; rev=rev, kw...))
topkind(x::AbstractArray, k::Integer; dims::Integer=1, rev=true, kw...) = mapslices(y -> topkind(y, k; rev=rev, kw...), x; dims=dims)
@nograd topkperm
```
Unlike the PR this doesn't return the permutation, but do you need it for anything else? It also defaults to first dimension, and won't accept `dims=:`.

But to make this GPU-friendly seems tricker. There should be no aliasing issues with the gradient. But neither  `partialsortperm` nor `mapslices` will work there. Maybe `topkind(::CuArray, ::Int)` can be written in terms of `accumulate` with a little thought.

ToucheSir:
For GPU compat, TF's XLA implementation should be using all out of place ops: https://github.com/tensorflow/tensorflow/blob/8d72537c6abf5a44103b57b9c2e22c14f5f49698/tensorflow/core/tpu/kernels/topk_ops.cc. Of course we can't rely on any of the optimizations in XLA, so a more ideal implementation would probably look the PyTorch one here: https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/TensorTopK.cu.

mcabbott:
What I hoped might work, re-using what CUDA.jl already has, doesn't seem to -- can this easily be fixed? 
```julia
julia> sort(tuple.(cu(rand(10)), 1:10), by=first)
ERROR: InvalidIRError: compiling kernel qsort_kernel(CuDeviceVector{Tuple{Float32, Int64}, 1}, Int64, Int64, Bool, Val{true}, Int64, Nothing, typeof(isless), typeof(first), Val{1}) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to zero)
Stacktrace:
 [1] bitonic_median
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:217
 [2] qsort_kernel
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:405
 [3] qsort_kernel
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:378
Reason: unsupported dynamic function invocation (call to zero)
Stacktrace:
 [1] bubble_sort
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:274
 [2] qsort_kernel
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:401
 [3] qsort_kernel
   @ ~/.julia/packages/CUDA/9T5Sq/src/sorting.jl:378
Stacktrace:
  [1] check_ir(job::GPUCompiler.CompilerJob{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams, GPUCompiler.FunctionSpec{typeof(CUDA.Quicksort.qsort_kernel), Tuple{CuDeviceVector{Tuple{Float32, Int64}, 1}, Int64, Int64, Bool, Val{true}, Int64, Nothing, typeof(isless), typeof(first), Val{1}}}}, args::LLVM.Module)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/fG3xK/src/validation.jl:111
  [2] macro expansion
    @ ~/.julia/packages/GPUCompiler/fG3xK/src/driver.jl:319 [inlined]
```

CarloLucibello:
@mcabbott thanks for all these suggestions, feel free to push on this branch.

I like the `topk` - `topkperm`  decoupling, so that it mimics base's sort functions. 

Representation of the (partial) permutation as an array of CartesianIndex it's redundant, since it would be enough to 
return an array of integers

```
topkout[i1, ..,iN, k] = x[i1, ...,iN, perm[i1, ..., iN, k]]
```

We could return a custom type for `perm` storing the integer permutation and `dims`.  `getindex` could be overloaded so that we can do `x[perm]`. But maybe this adds complication that turns out to be not so useful in the end, so I would be totally fine with returning cartesian indexes and revisiting later if needed.

PR #356, opened 2021-10-18
Add pictures for activation functions

mcabbott:
Might be nice to have?

Not entirely sure how well these will render in different places. Also whether this package tests its own `jldoctest`s.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NNlib"
 uuid = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
-version = "0.7.29"
+version = "0.7.30"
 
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
We could render them in the docs but large help sections that aren't critical can be annoying to use in the REPL.

CarloLucibello:
I think it is very helpful to have these plots under standard help. I also like the fact that they will show up in the documentation site. 
Also, the docstrings' length shouldn't be a concern here, they are very short and you consult them very rarely

mcabbott:
Build failures are (at least) two different `isapprox` tests of convolutions, plus Zygote on Julia 1.8

PR #360, opened 2021-10-31
Add Gated Linear Unit (GLU)

mashu:
GLU  from "Language Modeling with Gated Convolutional Networks" Dauphin et al. 2017

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -1,0 +1,19 @@
+using NNlib: sigmoid
+"""
+    glu(x, dim)
+
+The gated linear unit.
+
+Calculates `glu(x, dim) = a ⊗ σ(b)`, where
+  `x` is split in half along given dimension `dim` to form `a` and `b`
+  then the element-wise product between above matrices is computed
+See ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083).
+"""
+function glu(x, dim)
+    maxdim = maximum(axes(x,dim))
+    @assert maxdim % 2 == 0 "Dimension must be even"
+    half = Int(maxdim / 2)
+    a,b = selectdim(x, dim, 1:half), selectdim(x,dim, half+1:maxdim)
+    a.*sigmoid.(b)
+end
+
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -2,12 +2,9 @@ using NNlib: sigmoid
 """
     glu(x, dim = 1)
 
-The gated linear unit.
+The gated linear unit from the ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083) paper.
 
-Calculates `glu(x, dim) = a ⊗ σ(b)`, where
-  `x` is split in half along given dimension `dim` to form `a` and `b`
-  then the element-wise product between above matrices is computed
-See ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083).
+Calculates `a .* sigmoid(b)`, where `x` is split in half along given dimension `dim` to form `a` and `b`.
 """
 function glu(x, dim = 1)
     maxdim = size(x, dim)
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -1,6 +1,6 @@
 using NNlib: sigmoid
 """
-    glu(x, dim)
+    glu(x, dim = 1)
 
 The gated linear unit.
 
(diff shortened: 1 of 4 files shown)


CarloLucibello:
This is not an activation (an elementwise function) and it is not presented as such in the paper AFAIK.
This is part of a layer's forward pass definition. We could consider adding a GLU layer to Flux.jl, but this PR should be closed

mashu:
Good point it's a layer, and it was incorrectly named "activation" somewhere else. When it comes to implementation isn't it also what PyTorch does? Please look here https://pytorch.org/docs/stable/_modules/torch/nn/functional.html#glu
I can wrap it up as a layer and make new PR, if that's ok with you. I need it for the language model implementation from that paper.

ToucheSir:
I'd say this functional piece can still belong in NNlib, just not as an activation. Now the question becomes which file to put it in (maybe a new one?)

CarloLucibello:
Maybe a `src/functions.jl`? Besides that, this PR has many missing features: 
- quality and style of the docstring should be improved, see the other docstrings as a reference
- the new file should be included by `src/NNlib.jl`
- tests are missing

ToucheSir:
I suggested `gates.jl` or `gated.jl` in anticipation of moving some RNN primitives here eventually as well, but `functions.jl` works too.

mashu:
I fixed docstring to start with function itself, moved this to src/functions.jl as suggested and added one regression test and one for assertion.

src/functions.jl
```diff
@@ -0,0 +1,19 @@
+using NNlib: sigmoid
+"""
+    glu(x, dim)
+
+The gated linear unit.
+
+Calculates `glu(x, dim) = a ⊗ σ(b)`, where
+  `x` is split in half along given dimension `dim` to form `a` and `b`
+  then the element-wise product between above matrices is computed
+See ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083).
```
CarloLucibello, line 10:
```suggestion
The gated linear unit of the ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083) paper.

Calculates `glu(x, dim) = a ⊗ σ(b)`, where `x` is split in half along given dimension `dim` to form `a` and `b`
then the element-wise product between above matrices is computed.
```

src/functions.jl
```diff
@@ -0,0 +1,19 @@
...
+"""
+function glu(x, dim)
+    maxdim = maximum(axes(x,dim))
+    @assert maxdim % 2 == 0 "Dimension must be even"
+    half = Int(maxdim / 2)
+    a,b = selectdim(x, dim, 1:half), selectdim(x,dim, half+1:maxdim)
+    a.*sigmoid.(b)
```
CarloLucibello, line 17:
```suggestion
    maxdim = maximum(axes(x, dim))
    @assert maxdim % 2 == 0 "Dimension must be even"
    half = Int(maxdim / 2)
    a, b = selectdim(x, dim, 1:half), selectdim(x, dim, half+1:maxdim)
    a .* sigmoid.(b)
```

src/functions.jl
```diff
@@ -0,0 +1,19 @@
+using NNlib: sigmoid
+"""
+    glu(x, dim)
```
CarloLucibello, line 3:
```suggestion
    glu(x, dim = 1)
```

src/functions.jl
```diff
@@ -0,0 +1,19 @@
...
+
+The gated linear unit.
+
+Calculates `glu(x, dim) = a ⊗ σ(b)`, where
+  `x` is split in half along given dimension `dim` to form `a` and `b`
+  then the element-wise product between above matrices is computed
+See ["Language Modeling with Gated Convolutional Networks"](https://arxiv.org/abs/1612.08083).
+"""
+function glu(x, dim)
```
CarloLucibello, line 12:
```suggestion
function glu(x, dim = 1)
```

test/functions.jl
```diff
@@ -0,0 +1,8 @@
+using NNlib: glu
+
+@testset "glu" begin
+    x = [1 2 3; 4 5 6; 7 8 9; 10 11 12]	 
+    @test ceil.(glu(x,1)) == [1 2 3; 4 5 6]
+    @test_throws AssertionError glu(x,2)
+end
+
```
ToucheSir, line 8: Gradients should be tested as well, see https://github.com/FluxML/NNlib.jl/blob/542e5cdb537b727ca0f3b97e3bef7bf9cf897a3c/test/activations.jl#L224.

CarloLucibello:
thanks for this!

mashu:
No problem, by the way, is there anything missing that it does not appear in [docs](https://fluxml.ai/Flux.jl/dev/models/nnlib/)?

CarloLucibello:
Those have to be updated with a separate PR to Flux.jl. Unfortunately, the current documentation setup is not ideal

Issue #359, opened 2021-10-30
(Flaky?) CI failures on GHA latest + Buildkite


ToucheSir:
The Appveyor badge in the readme should be swapped out for a Buildkite one as well.

DhairyaLGandhi:
Is it flaky or have many of the tests been consistently failing? GPU tests and some convolution tests come to mind.

ToucheSir:
At least on the couple of commits I checked, different AD tests were failing between the different CI services. Which seems like a fun problem :sweat_smile:

ToucheSir:
After some more digging, these failures really do seem random. I've yet to see all 3 conv spatial autodiff suites fail in one run, but it's not uncommon to see tests in 2/3 fail. Perhaps the tolerances are too strict?

ToucheSir:
Does anyone know how many threads Buildkite runs by default? I had a look back through test failures on merge commits on GHA, and they are solely limited to the config with 2 threads (e.g. [1](https://github.com/FluxML/NNlib.jl/runs/3533818348?check_suite_focus=true), [2](https://github.com/FluxML/NNlib.jl/runs/4070591455?check_suite_focus=true), [3](https://github.com/FluxML/NNlib.jl/runs/3922909573?check_suite_focus=true)). My hunch is that Buildkite is using more threads by default and thus running into issues more frequently.

ToucheSir:
Ok, I've been able to replicate this semi-frequently locally with 4/8 threads (on 4 cores). Running the depthwise conv fuzz tests is pretty damning:
```
[ Info: Starting Depthwise Convolutional fuzzing tests; this can take a few minutes...
.......................fuzzing: Test Failed at /home/brianc/projects/julia-dev/NNlib.jl/test/conv.jl:635
  Expression: dw_direct ≈ dw_im2col
   Evaluated: [0.18744640672662355] ≈ [0.00810922314783625]
Stacktrace:
 [1] macro expansion
   @ ~/projects/julia-dev/NNlib.jl/test/conv.jl:635 [inlined]
 [2] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [3] top-level scope
   @ ~/projects/julia-dev/NNlib.jl/test/conv.jl:565
fuzzing: Test Failed at /home/brianc/projects/julia-dev/NNlib.jl/test/conv.jl:635
[...]

PR #361, opened 2021-11-02
Add grid sampling

pxl-th:
This PR adds [grid sampling](https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html) for the 4D case.

This is needed for things like self-supervised depth estimation, where a neural network would learn a depth `D` for the current image and projection transformation `P` for the adjacent images and does "warping" of the images using `D` and `P`.
And this function does image sampling given projected coordinates (a.k.a. `grid`).

The implementation differs from the PyTorch's version in that it assumes `align_corners = True` and uses only bilinear interpolation. It was tested to give the same results as PyTorch's version.

[PR](https://github.com/FluxML/NNlibCUDA.jl/pull/31) with the CUDA kernels.

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -1,7 +1,8 @@
-@inline in_bounds(h, w, H, W) = 1 ≤ h ≤ H && 1 ≤ w ≤ W
-@inline clip_coordinate(coordinate, dim_size) = min(dim_size, max(1, coordinate))
+export grid_sample, ∇grid_sample
 
-# Borders are considered out-of-bounds.
+@inline in_bounds(h, w, H, W) = 1 ≤ h ≤ H && 1 ≤ w ≤ W
+# Borders are considered out-of-bounds for gradient.
+@inline clip_coordinate(coordinate, dim_size) = min(dim_size, max(1, coordinate))
 @inline function ∇clip_coordinate(coordinate::C, dim_size) where C
     if coordinate ≤ 1
         return C(1), C(0)
@@ -11,7 +12,7 @@
     coordinate, C(1)
 end
 
-@inline unnormalize(coordinate, dim_size) = ((coordinate + 1.0) / 2.0) * (dim_size - 1.0) + 1.0
+@inline unnormalize(coordinate, dim_size) = ((coordinate + 1.0) * 0.5) * (dim_size - 1.0) + 1.0
 @inline ∇unnormalize(coordinate, dim_size) = unnormalize(coordinate, dim_size), (dim_size - 1.0) * 0.5
 
 @inline compute_source_index(coordinate, dim_size, ::Val{:zeros}) = unnormalize(coordinate, dim_size)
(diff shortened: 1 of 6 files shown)


pxl-th:
I've commented out failing test for the depthwise conv for now. Should I bring it back?

CarloLucibello:
They were failing intermittently, mostly in the multi-threading environment. Since we have a few multithreaded methods, better have the multithreaded CI passing, so I'm ok with keeping those problematic tests out.  We should open an issue for those depthwiseconv failures if there is no one already. 

I'll try to review in the next few days, thanks for this contribution, at a first glance looks perfect already!

ToucheSir:
@CarloLucibello RE CI failures, feel free to rename and re-purpose https://github.com/FluxML/NNlib.jl/issues/359.

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+
+@inline ∇compute_source_index(coordinate, dim_size, ::Val{:zeros}) = ∇unnormalize(coordinate, dim_size)
+@inline function ∇compute_source_index(coordinate, dim_size, ::Val{:border})
+    source_coordinate, grad_in = ∇unnormalize(coordinate, dim_size)
+    source_coordinate, grad_clip = ∇clip_coordinate(source_coordinate, dim_size)
+    source_coordinate, grad_in * grad_clip
+end
+
+"""
+    grid_sample(input::AbstractArray{T, 4}, grid::AbstractArray{T, 4}; padding_mode = Val(:zeros))
```
CarloLucibello, line 29:
```suggestion
    grid_sample(input::AbstractArray{T, 4}, grid::AbstractArray{T, 4}; padding_mode = :zeros)
```
Our API should not require users to wrap symbols in a `Val`

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+(i.e. align corners is `true`).
+
+# Arguments
+
+- `input`: Input array in `(W_in, H_in, C, N)` shape.
+- `grid`: Input grid in `(2, W_out, H_out, N)` shape.
+    Where for each `(W_out, H_out, N)` grid contains `(x, y)`
+    coordinates that specify sampling locations normalized by the `input` shape.
+
+    Therefore, it should have values mostly in `[-1, 1]` range.
```
CarloLucibello, line 45:
```suggestion
    Therefore, `x` and `y`  should have values mostly in `[-1, 1]` range.
```
Also, why "mostly?"

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+    For example, values `x = -1, y = -1` is the left-top pixel of `input`,
+    and values `x = 1, y = 1` is the right-bottom pixel of `input`.
```
CarloLucibello, line 47:
```suggestion
    For example,  `(x = -1, y = -1)` is the left-top pixel of `input`,
    and `(x = 1, y = 1)` is the right-bottom pixel of `input`.
```

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+
+    Out-of-bound values are handled accroding to the `padding_mode`.
```
CarloLucibello, line 49:
```suggestion
    Out-of-bound values are handled according to the `padding_mode`.
```

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+        end
+        if in_bounds(iy_se, ix_se, iH, iW)
+            r += input[ix_se, iy_se, c, n] * se
+        end
+        output[w, h, c, n] = r
+    end
+end
+
+"""
+    ∇grid_sample(Δ::AbstractArray{T, 4}, input::AbstractArray{T, 4}, grid::AbstractArray{T, 4}; padding_mode = Val(:zeros)) where T
```
CarloLucibello, line 155:
```suggestion
    ∇grid_sample(Δ::AbstractArray{T, 4}, input::AbstractArray{T, 4}, grid::AbstractArray{T, 4}; padding_mode = :zeros) where T
```

src/sampling.jl
```diff
@@ -0,0 +1,253 @@
...
+end
+
+@inline function _safe_add!(dx, value, ix, iy, c, n)
+    @inbounds dx[ix, iy, c, n] += value
+end
+
+function rrule(::typeof(grid_sample), x, grid; padding_mode)
+    y = grid_sample(x, grid; padding_mode=padding_mode)
+    function grid_sample_pullback(Δ)
+        ∇x, ∇grid = @thunk(∇grid_sample(unthunk(Δ), x, grid; padding_mode=padding_mode))
```
CarloLucibello, line 249:
```suggestion
        ∇x, ∇grid = ∇grid_sample(unthunk(Δ), x, grid; padding_mode=padding_mode)
```
Ideally, we want to separate the two gradient computations into 2 different thunks. Since the computation is shared here, wrapping into a thunk should be entirely useless

src/sampling.jl
pxl-th: >Also, why "mostly?"

It was added as a hint that the values can be outside, but probably makes no sense.

maxfreu:
Hi! Just to leave this here, as I ported the tri/bi/linear upsampling code from pytorch: 1) Setting `align_corners = true` is not the best choice, as the gradients change with changing input image size. Pytorch's default therefore is `false`. But that doesn't change any results much. 2) The code you wrote reminds me a lot of what I wrote for upsampling. The kernels probably can't be shared, although they look quite similar. But the [utility functions](https://github.com/FluxML/NNlib.jl/blob/063992bc295568551395207f99e135dd1e802138/src/upsample.jl#L84-L98) can be shared I think - but that's maybe also for the future.

gRox167:
Just to confirm, is there any plan to add 5D support for grid_sampling? 
This is really helpful for doing 3D image registration, which is widely use in neuro and medical imaging field. Networks in medical image registration utilize this function extensively to wrap the moving image to a fixed image, for example VoxelMorph.

CarloLucibello:
No active plans that I know of, but contribution are welcome. Maybe @maxfreu could provide some guidance.

maxfreu:
Hi @gRox167 ! As imaging science phd student this is the perfect task for you :D In principle it should be quite simple:

1. Fork NNlib
2. Come up with simple test cases for the forward and backward pass that you want to pass and write those first.
3. Add specializations for `AbstractArray{T,4}` to the current implementation where it is needed.
4. Copy the code for the 4D case, change the specialization to 5D.
5. Add the depth dimension where sizes are checked etc.
6. Change the sampling code in the kernel. You can use the pytorch code as a reference, which you can find [here](https://github.com/pytorch/pytorch/blob/f7000350905be5073892e0b23df681c0281be0f0/aten/src/ATen/native/cuda/GridSampler.cu#L156) for the GPU and [here](https://github.com/pytorch/pytorch/blob/f7000350905be5073892e0b23df681c0281be0f0/aten/src/ATen/native/GridSampler.cpp#L41) for the CPU.
7. Ping me or @pxl-th in case you hit bumps in the road :)
8. Enjoy how little code remains of the C++ hell.

It should be pretty straight forward and more of a writing than a thinking task. Optionally, you can simply use ChatGPT, show it the current julia and the pytorch code and tell it to expand it. I guess that should also work pretty well nowadays. The key to success in any case are good tests for the forward and backward pass.

gRox167:
> Hi @gRox167 ! As imaging science phd student this is the perfect task for you :D In principle it should be quite simple:
> 
>     1. Fork NNlib
> 
>     2. Come up with simple test cases for the forward and backward pass that you want to pass and write those first.
> 
>     3. Add specializations for `AbstractArray{T,4}` to the current implementation where it is needed.
> 
>     4. Copy the code for the 4D case, change the specialization to 5D.
> 
>     5. Add the depth dimension where sizes are checked etc.
> 
>     6. Change the sampling code in the kernel. You can use the pytorch code as a reference, which you can find [here](https://github.com/pytorch/pytorch/blob/f7000350905be5073892e0b23df681c0281be0f0/aten/src/ATen/native/cuda/GridSampler.cu#L156) for the GPU and [here](https://github.com/pytorch/pytorch/blob/f7000350905be5073892e0b23df681c0281be0f0/aten/src/ATen/native/GridSampler.cpp#L41) for the CPU.
> 
>     7. Ping me or @pxl-th in case you hit bumps in the road :)
> 
>     8. Enjoy how little code remains of the C++ hell.
> 
> 
> It should be pretty straight forward and more of a writing than a thinking task. Optionally, you can simply use ChatGPT, show it the current julia and the pytorch code and tell it to expand it. I guess that should also work pretty well nowadays. The key to success in any case are good tests for the forward and backward pass.

Thanks for your guidance, I think I can take a try.

gRox167:
Please refer to PR #627 .

Issue #354, opened 2021-10-01
add Sparsemax activation


tylerjthomas9:
Source paper: http://arxiv.org/abs/1602.02068

PyTorch implementation: https://github.com/Qwicen/node/blob/master/lib/nn_utils.py

I started working on implementing `sparsemax `in Julia for TabNet. I thought that it would best fit in NNlib.jl. It should have the exact same functionality as `softmax`.

darsnack:
If you have an implementation, then a PR would be welcome! We can iterate the design there.

tylerjthomas9:
Still trying to get the jacobian to work, but I have the initial forward pass

```julia
using NNlib
using LinearAlgebra
using Zygote

sparsemax(x; dims = 1) = sparsemax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)

sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)

function sparsemax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
     # only 2D tensors are supported
     @assert dims in (1, 2)
    
     max_ = maximum(x; dims=dims)
     x .-= max_
 
     # make ix like
     d = size(x, dims)
     if dims==1
         rhos = reshape(collect(1:d), d, 1) |> typeof(x)
     elseif dims == 2 
         rhos = reshape(collect(1:d), 1, d) |> typeof(x)
     end
         
 
     # compute threshold and support
     x_sorted = sort(x; dims=dims, rev=true)
     x_cumsum = cumsum(x_sorted; dims=dims) .- 1.0
     support =  rhos .* x_sorted .> x_cumsum
     support_size = vec(sum(support; dims=dims)) |> Vector{Int64}
     if dims == 1
         tau = diag(NNlib.gather(transpose(x_cumsum), support_size))
     elseif dims == 2
         tau = diag(NNlib.gather(x_cumsum, support_size))
     end
     tau ./= support_size
     
     if dims == 1
         out = clamp.(x .- transpose(tau), 0, Inf)
     elseif dims == 2
         out =  clamp.(x .- tau, 0, Inf)
     end
end

x = [0.3367 -0.1863; 0.1288 2.2082; 0.2345 -0.638; 0.2303 0.4617; -1.1229 0.2674] 
println("Sparsemax probabilities")
sparsemax(x; dims=1)
```


```
Sparsemax probabilities

2×5 Matrix{Float64}:
[...]

PR #364, opened 2021-11-19
add sparsemax

tylerjthomas9:
Initial sparsemax implementation #354

TODO
- [x] add support for vectors (similar to softmax) 
- [ ] add comprehensive testing
- [x] compare gradients with PyTorch implementations

### DIFF
--- a/src/sparsemax.jl
+++ b/src/sparsemax.jl
@@ -4,8 +4,9 @@
 [Sparsemax](https://arxiv.org/abs/1602.02068) turns input array `x`
 into sparse probability distributions that sum to 1 along the dimensions specified by `dims`.
 
-For a matrix input `x` it will by default (`dims = 1`) treat it as a batch of vectors,
-with each column independent. Keyword `dims = 2` will instead treat rows independently, and so on.
+Similar to softmax, each dimension is considered independent. For a matrix input `x` it will 
+by default (`dims = 1`) treat it as a batch of vectors, with each column independent. Keyword 
+`dims = 2` will instead treat rows independently.
 
 # Examples
 
@@ -28,7 +29,7 @@ sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)
 function sparsemax!(out::AbstractArray, x::AbstractArray; dims = 1)
     # only 2D tensors are supported
     @assert dims in (1, 2)
-    x = x .- maximum(x; dims=dims)
+    x = x .- maximum(x; dims=dims) 
 
     # make ix like
     d = size(x, dims)
--- a/src/sparsemax.jl
+++ b/src/sparsemax.jl
@@ -21,7 +21,7 @@ julia> sparsemax([1 2 3; 2 2 2]; dims=2)
 0.333333  0.333333  0.333333
 ```
 """
-sparsemax(x; dims = 1) = sparsemax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)
+sparsemax(x; dims = 1) = sparsemax!(similar(x, (eltype)(x)), x; dims = dims)
 
 sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)
 
--- a/src/sparsemax.jl
+++ b/src/sparsemax.jl
@@ -37,15 +37,21 @@ julia> sparsemax(reshape(1:15, 5, 3) ./ 20) # Scale matters
 ```
 """
 function sparsemax(x::AbstractArray; dims::Integer=1)
+    n_dims = length(size(x))
+    if n_dims < dims
+        error_msg = "Dims ($dims) is larger than the array dimensions ($n_dims)."
+        throw(DimensionMismatch(error_msg))
+    end
+
     if isa(x, AbstractVector)
-        z = _sort_vector(x)
+        z = _sort_vector(x; dims=dims)
     else
         z = sort(float(x); dims=dims, rev=true) 
     end
     mask = _sparsemax_mask(z, dims)
     tausum = sum(z .* mask; dims) 
     kay = sum(mask; dims)
-    z = max.(x  .- (tausum .- 1) ./ kay, 0) # relu
+    z = _relu.(x  .- (tausum .- 1) ./ kay)
 end
 
 function _sort_vector(x::AbstractVector; dims::Integer=1, rev=true)
@@ -53,18 +59,26 @@ function _sort_vector(x::AbstractVector; dims::Integer=1, rev=true)
         z = sort(float(x); rev=rev)
     elseif dims == 2
         z = x
-    else
-        error("Dims must be in (1,2) for a vector. Not " * string(dims))
     end
     return z
 end
 
 function _sparsemax_mask(z::AbstractArray, dim::Integer)
     acc = cumsum(z; dims=dim)
-    # This isn't type-stable. Writing into `acc` ensures the whole function still is:
-    cnt = reshape(axes(z, dim), ntuple(_->1, dim-1)..., :)
-    acc = 1 .+ cnt .* z .> acc
+    if dim == 1
+        acc .= 1 .+ axes(z,1) .* z .> acc
+    elseif dim == 2
+        acc .= 1 .+ axes(z,2)' .* z .> acc
+    else
+        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
+        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
+        acc .= 1 .+ cnt .* z .> acc
+    end
+    acc
 end
+
+_ifelse(p, x, y) = ifelse(p, promote(x, y)...)
+_relu(x) = _ifelse(x>0, x, false)  # different gradient at zero
 
 function ∇sparsemax(Δ::AbstractArray, y::AbstractArray; dims = 1)
     nonzeros = Δ.!=0.0
--- a/src/sparsemax.jl
+++ b/src/sparsemax.jl
@@ -1,0 +1,72 @@
+"""
+    sparsemax(x; dims = 1)
+
+[Sparsemax](https://arxiv.org/abs/1602.02068) turns input array `x`
+into sparse probability distributions that sum to 1 along the dimensions specified by `dims`.
+
+For a matrix input `x` it will by default (`dims = 1`) treat it as a batch of vectors,
+with each column independent. Keyword `dims = 2` will instead treat rows independently, and so on.
+
+# Examples
+
+```jldoctest
+julia> sparsemax([1 2 3; 2 2 2])  # dims=1
+2×3 Matrix{Float64}:
+0.0  0.5  1.0
+1.0  0.5  0.0
+
+julia> sparsemax([1 2 3; 2 2 2]; dims=2)
+2×3 Matrix{Float64}:
+0.0       0.0       1.0
+0.333333  0.333333  0.333333
+```
+"""
+sparsemax(x; dims = 1) = sparsemax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)
+
+sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)
+
+function sparsemax!(out::AbstractArray, x::AbstractArray; dims = 1)
+    # only 2D tensors are supported
+    @assert dims in (1, 2)
+    x = x .- maximum(x; dims=dims)
+
+    # make ix like
+    d = size(x, dims)
+    if dims==1
+        rhos = reshape(collect(1:d), d, 1) |> typeof(x)
+    elseif dims == 2 
+        rhos = reshape(collect(1:d), 1, d) |> typeof(x)
+    end
+
+
+    # compute threshold and support
+    x_sorted = slicemap(x -> reverse(sort(x)), x; dims=dims)
+    x_cumsum = cumsum(x_sorted; dims=dims) .- 1.0
+    support =  rhos .* x_sorted .> x_cumsum
+    support_size = vec(sum(support; dims=dims)) |> Vector{Int64}
+    if dims == 1
+        tau = diag(gather(transpose(x_cumsum), support_size)) ./ support_size
+    elseif dims == 2
+        tau = diag(gather(x_cumsum, support_size)) ./ support_size
+    end
+
+    if dims == 1
+        out = clamp.(x .- transpose(tau), 0, Inf)
+    elseif dims == 2
+        out =  clamp.(x .- tau, 0, Inf)
+    end
+end
+
+
+function ∇sparsemax!(Δ::AbstractArray, x::AbstractArray, y::AbstractArray; dims = 1)
+   nonzeros = x[x.!=0.0]
+   out .= Δ .* y
+   sum = sum(out .* nonzeros; dims=dims) / sum(nonzeros; dims=dims)
+   out .= nonzeros .* (out .- sum)
+end
+
+function rrule(::typeof(sparsemax), xs; dims=1)
+   y = sparsemax(xs; dims=dims)
+   sparsemax_pullback(Δ) = (NoTangent(), ∇sparsemax(unthunk(Δ), xs, y, dims = dims))
+   return y, sparsemax_pullback
+end
--- a/src/sparsemax.jl
+++ b/src/sparsemax.jl
@@ -22,52 +22,40 @@ julia> sparsemax([1 2 3; 2 2 2]; dims=2)
 0.333333  0.333333  0.333333
 ```
 """
-sparsemax(x; dims = 1) = sparsemax!(similar(x, (eltype)(x)), x; dims = dims)
-
-sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)
-
-function sparsemax!(out::AbstractArray, x::AbstractArray; dims = 1)
-    # only 2D tensors are supported
-    @assert dims in (1, 2)
-    x = x .- maximum(x; dims=dims) 
-
-    # make ix like
-    d = size(x, dims)
-    if dims==1
-        rhos = reshape(collect(1:d), d, 1) |> typeof(x)
-    elseif dims == 2 
-        rhos = reshape(collect(1:d), 1, d) |> typeof(x)
-    end
-
-
-    # compute threshold and support
-    x_sorted = slicemap(x -> reverse(sort(x)), x; dims=dims)
-    x_cumsum = cumsum(x_sorted; dims=dims) .- 1.0
-    support =  rhos .* x_sorted .> x_cumsum
-    support_size = vec(sum(support; dims=dims)) |> Vector{Int64}
-    if dims == 1
-        tau = diag(gather(transpose(x_cumsum), support_size)) ./ support_size
-    elseif dims == 2
-        tau = diag(gather(x_cumsum, support_size)) ./ support_size
-    end
-
-    if dims == 1
-        out = clamp.(x .- transpose(tau), 0, Inf)
-    elseif dims == 2
-        out =  clamp.(x .- tau, 0, Inf)
-    end
+function sparsemax(x::AbstractArray; dims::Integer=1)
+    z = slicemap(x -> reverse(sort(x)), x; dims=dims)  # float is usually free, except on integers etc.
+    mask = _sparsemax_mask(z, dims)
+    tausum = sum(z .* mask; dims)  # no longer need z
+    kay = sum(mask; dims)
+    z = _relu.(x  .- (tausum .- 1) ./ kay)
 end
 
+function _sparsemax_mask(z::AbstractArray, dim::Integer)
+    acc = cumsum(z; dims=dim)
+    if dim == 1
+        acc = 1 .+ axes(z,1) .* z .> acc
+    elseif dim == 2
+        acc = 1 .+ axes(z,2)' .* z .> acc
+    else
+        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
+        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
+        acc = 1 .+ cnt .* z .> acc
+    end
+    acc
+end
 
-function ∇sparsemax!(Δ::AbstractArray, x::AbstractArray, y::AbstractArray; dims = 1)
-   nonzeros = x[x.!=0.0]
-   out .= Δ .* y
-   sum = sum(out .* nonzeros; dims=dims) / sum(nonzeros; dims=dims)
-   out .= nonzeros .* (out .- sum)
+_relu(x) = _ifelse(x>0, x, false)  # different gradient at zero
+
+_ifelse(p, x, y) = ifelse(p, promote(x, y)...)
+
+function ∇sparsemax(dy::AbstractArray, y::AbstractArray; dims::Integer=1)
+    vee = sum(dy .* (y .> 0); dims)
+    kay = count(>(0), y; dims)  # could also keep from forward pass?
+    _ifelse.(y .> 0, dy .- vee ./ kay, 0)
 end
 
 function rrule(::typeof(sparsemax), xs; dims=1)
-   y = sparsemax(xs; dims=dims)
-   sparsemax_pullback(Δ) = (NoTangent(), ∇sparsemax(unthunk(Δ), xs, y, dims = dims))
-   return y, sparsemax_pullback
-end
+    y = sparsemax(xs; dims=dims)
+    sparsemax_pullback(Δ) = (NoTangent(), ∇sparsemax(xs, y, dims = dims))
+    return y, sparsemax_pullback
+ end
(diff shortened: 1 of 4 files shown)


src/sparsemax.jl
```diff
@@ -0,0 +1,72 @@
...
+0.0  0.5  1.0
+1.0  0.5  0.0
+
+julia> sparsemax([1 2 3; 2 2 2]; dims=2)
+2×3 Matrix{Float64}:
+0.0       0.0       1.0
+0.333333  0.333333  0.333333
+```
+"""
+sparsemax(x; dims = 1) = sparsemax!(similar(x, (float ∘ eltype)(x)), x; dims = dims)
```
DhairyaLGandhi, line 24: Do we need to convert to a float here?

src/sparsemax.jl
```diff
@@ -0,0 +1,72 @@
...
+
+sparsemax!(x; dims = 1) = sparsemax!(x, x; dims = dims)
```
DhairyaLGandhi, line 26: I understand that it's probably find to have this if every dimension could be considered independent. Good to add a comment about this so it's known why it's a fine optimisation

DhairyaLGandhi (commented):
Looks mostly for, thank you! I've left a couple of thoughts.

What would you think of generalising this across dimensions instead checking for dims explicitly?

src/sparsemax.jl
tylerjthomas9: Just tested removing the float conversion. We do not need it

src/sparsemax.jl
tylerjthomas9: > What would you think of generalising this across dimensions instead checking for dims explicitly?

looking into generalizing the implementation

src/sparsemax.jl
```diff
@@ -0,0 +1,73 @@
...
+    elseif dims == 2 
+        rhos = reshape(collect(1:d), 1, d) |> typeof(x)
+    end
+
+
+    # compute threshold and support
+    x_sorted = slicemap(x -> reverse(sort(x)), x; dims=dims)
+    x_cumsum = cumsum(x_sorted; dims=dims) .- 1.0
+    support =  rhos .* x_sorted .> x_cumsum
+    support_size = vec(sum(support; dims=dims)) |> Vector{Int64}
```
CarloLucibello, line 47:
```suggestion
    support_size = vec(sum(support; dims=dims))
```
Why this cast?

src/sparsemax.jl
```diff
@@ -0,0 +1,73 @@
...
+        tau = diag(gather(transpose(x_cumsum), support_size)) ./ support_size
+    elseif dims == 2
+        tau = diag(gather(x_cumsum, support_size)) ./ support_size
+    end
+
+    if dims == 1
+        out = clamp.(x .- transpose(tau), 0, Inf)
+    elseif dims == 2
+        out =  clamp.(x .- tau, 0, Inf)
+    end
```
CarloLucibello, line 58:
```suggestion
    if dims == 1
        out .= clamp.(x .- transpose(tau), 0, Inf)
    elseif dims == 2
        out .=  clamp.(x .- tau, 0, Inf)
    end
```
otherwise the `out` argument won't be mutated

CarloLucibello:
I think we can avoid defining a mutating version `sparsemax!` and just go with `sparsemax`, since the implementation of `sparsemax!` does a lot of allocations in any case

src/sparsemax.jl
```diff
@@ -0,0 +1,73 @@
...
+end
+
+
+function ∇sparsemax!(Δ::AbstractArray, x::AbstractArray, y::AbstractArray; dims = 1)
+   nonzeros = x[x.!=0.0]
+   out .= Δ .* y
```
CarloLucibello, line 64:
```suggestion
   out = Δ .* y
```
`out` does not exist before

test/sparsemax.jl
```diff
@@ -0,0 +1,5 @@
+using Statistics: mean
+
+@testset "sparsemax integer input" begin
+    @test sparsemax(Int[0 0; 0 0]) == [0.5 0.5; 0.5 0.5]
+end
```
CarloLucibello, line 5: needs some gradient testing

src/sparsemax.jl
```diff
@@ -0,0 +1,73 @@
...
+    # compute threshold and support
+    x_sorted = slicemap(x -> reverse(sort(x)), x; dims=dims)
+    x_cumsum = cumsum(x_sorted; dims=dims) .- 1.0
+    support =  rhos .* x_sorted .> x_cumsum
+    support_size = vec(sum(support; dims=dims)) |> Vector{Int64}
+    if dims == 1
+        tau = diag(gather(transpose(x_cumsum), support_size)) ./ support_size
+    elseif dims == 2
+        tau = diag(gather(x_cumsum, support_size)) ./ support_size
+    end
```
CarloLucibello, line 52: This tells us tht we should add a `dims` keyword to gather to handle this behaviour

src/sparsemax.jl
```diff
@@ -0,0 +1,73 @@
...
+    elseif dims == 2
+        out =  clamp.(x .- tau, 0, Inf)
+    end
+end
+
+
+function ∇sparsemax!(Δ::AbstractArray, x::AbstractArray, y::AbstractArray; dims = 1)
+   nonzeros = x[x.!=0.0]
+   out .= Δ .* y
+   sum = sum(out .* nonzeros; dims=dims) / sum(nonzeros; dims=dims)
```
CarloLucibello, line 65: let's not call `sun` a variable

mcabbott:
Here's an attempt at a faster version:
```julia
function sm4(x; dims::Integer=1)
    z = sort(x; dims, rev=true)
    mask = _sm4mask(z, dims)
    tausum = sum(z .= z .* mask; dims)  # no longer need z, so can overwrite it
    kay = sum(mask; dims)
    # tau = (tausum .- 1) ./ kay  # but we can fuse that with next step:
    relu.(x .- (tausum .- 1) ./ kay)
end

function _sm4mask(z::AbstractArray, dim::Integer)
    acc = cumsum(z; dims=dim)
    if dim == 1
        acc .= 1 .+ axes(z,1) .* z .> acc
    elseif dim == 2
        acc .= 1 .+ axes(z,2)' .* z .> acc
    else
        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
        # cnt = reshape(collect(axes(x, dim)), ntuple(d -> d==dim ? (:) : 1, ndims(z)))
        acc .= 1 .+ cnt .* z .> acc
    end
    acc
end

x = rand(1:5, 5,10) ./ 5
sparsemax(x)
sm4(x)

sparsemax(x) ≈ sm4(x)
all(≈(1), sum(sm4(x), dims=1))
```

```
julia> @btime sparsemax($x);
  min 4.470 μs, mean 4.970 μs (63 allocations, 7.11 KiB. GC mean 5.49%)

julia> @btime sm1($x);
  min 1.188 μs, mean 1.311 μs (25 allocations, 2.06 KiB. GC mean 4.23%)

julia> @btime sm4($x);
  min 549.909 ns, mean 629.318 ns (7 allocations, 1.83 KiB. GC mean 6.72%)
```

src/sparsemax.jl
tylerjthomas9: oops, didn't notice this poor notation. removing it

src/sparsemax.jl
```diff
@@ -0,0 +1,61 @@
...
+1.0  0.5  0.0
+
+julia> sparsemax([1 2 3; 2 2 2]; dims=2)
+2×3 Matrix{Float64}:
+0.0       0.0       1.0
+0.333333  0.333333  0.333333
+```
+"""
+function sparsemax(x::AbstractArray; dims::Integer=1)
+    z = slicemap(x -> reverse(sort(x)), x; dims=dims)  # float is usually free, except on integers etc.
```
mcabbott, line 26: Why make slices? `sort(x; dims)` does what you want.

The only catch is that, when x is a vector, `sort([3,1,2]; dims=1)` will fail. So you probably need a branch `if x isa AbstractVector`.

src/sparsemax.jl
```diff
@@ -0,0 +1,61 @@
...
+```jldoctest
+julia> sparsemax([1 2 3; 2 2 2])  # dims=1
+2×3 Matrix{Float64}:
+0.0  0.5  1.0
+1.0  0.5  0.0
+
+julia> sparsemax([1 2 3; 2 2 2]; dims=2)
+2×3 Matrix{Float64}:
+0.0       0.0       1.0
+0.333333  0.333333  0.333333
```
mcabbott, line 22: These examples could come from a much simpler function. An idea for how to show that scale matters but shift doesn't is: 
```suggestion
julia> sparsemax(reshape(1:15, 5,3) ./ 2)
5×3 Matrix{Float64}:
 0.0   0.0   0.0
 0.0   0.0   0.0
 0.0   0.0   0.0
 0.25  0.25  0.25
 0.75  0.75  0.75

julia> sparsemax(reshape(1:15, 5,3) ./ 20)
5×3 Matrix{Float64}:
 0.1   0.1   0.1
 0.15  0.15  0.15
 0.2   0.2   0.2
 0.25  0.25  0.25
 0.3   0.3   0.3
```

src/sparsemax.jl
```diff
@@ -0,0 +1,61 @@
...
+    acc
+end
+
+_relu(x) = _ifelse(x>0, x, false)  # different gradient at zero
+
+_ifelse(p, x, y) = ifelse(p, promote(x, y)...)
+
+function ∇sparsemax(dy::AbstractArray, y::AbstractArray; dims::Integer=1)
+    vee = sum(dy .* (y .> 0); dims)
+    kay = count(>(0), y; dims)  # could also keep from forward pass?
```
mcabbott, line 53: One thing worth noticing is that this count isn't always identical to the one used before. E.g., inserting `@show`:
```
julia> sm4(reshape(1:50, 5,10) ./ 10)
kay = [4.0 4.0 4.0 4.0 4.0 5.0 4.0 4.0 4.0 4.0]
5×10 Matrix{Float64}:
 0.0  0.0  2.22045e-16  2.22045e-16  4.44089e-16  4.44089e-16  4.44089e-16  4.44089e-16  0.0  0.0
 0.1  0.1  0.1          0.1          0.1          0.1          0.1          0.1          0.1  0.1
 0.2  0.2  0.2          0.2          0.2          0.2          0.2          0.2          0.2  0.2
 0.3  0.3  0.3          0.3          0.3          0.3          0.3          0.3          0.3  0.3
 0.4  0.4  0.4          0.4          0.4          0.4          0.4          0.4          0.4  0.4

julia> count(>(0), ans; dims=1, init=Int8(0))
1×10 Matrix{Int8}:
 4  4  5  5  5  5  5  5  4  4
```
Does this mean the gradients are slightly wrong? Maybe it's a different subgradient, maybe not?

Also, note that `k` is never zero. So `_ifelse.(y .> 0, dy .- vee ./ kay, 0)` could possibly be replaced by `(y .> 0) .* (dy .- vee ./ kay)` without getting NaN, although I'm not sure this is better.

src/sparsemax.jl
tylerjthomas9: I was using slicemap, because i was getting problems with the jacobian when using `sort(;rev=true).` However, I am no longer getting these issues, so I will change it back to `sort(x; dims)`

src/sparsemax.jl
tylerjthomas9: For now, I am going to revert ∇sparsemax to mimic this pytorch implementation: https://github.com/KrisKorrel/sparsemax-pytorch/blob/master/sparsemax.py

src/sparsemax.jl
tylerjthomas9: I had that, so that it was an Integer vector when it was passed as `idx `to `NNlib.gather`.  Otherwise, I would get an invalid index error. However, @mcabbott's implementation alleviated the need for this

`ERROR: ArgumentError: invalid index: 1.0 of type Float64`

src/sparsemax.jl
```diff
@@ -0,0 +1,90 @@
...
+        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
+        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
+        acc = 1 .+ cnt .* z .> acc
+    end
+    acc
+end
+
+_relu(x) = _ifelse(x>0, x, false)  # different gradient at zero
+
+_ifelse(p, x, y) = ifelse(p, promote(x, y)...)
```
DhairyaLGandhi, line 77: We really should not need this manual promotion

src/sparsemax.jl
```diff
@@ -0,0 +1,90 @@
...
+        acc = 1 .+ axes(z,2)' .* z .> acc
+    else
+        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
+        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
+        acc = 1 .+ cnt .* z .> acc
+    end
+    acc
+end
+
+_relu(x) = _ifelse(x>0, x, false)  # different gradient at zero
```
DhairyaLGandhi, line 75: This will override the current relu fallback, which isn't the intention. Why is this required?

src/sparsemax.jl
```diff
@@ -0,0 +1,90 @@
...
+function _sparsemax_mask(z::AbstractArray, dim::Integer)
+    acc = cumsum(z; dims=dim)
+    if dim == 1
+        acc = 1 .+ axes(z,1) .* z .> acc
+    elseif dim == 2
+        acc = 1 .+ axes(z,2)' .* z .> acc
+    else
+        # This isn't type-stable. Writing into `acc` ensures the whole function still is:
+        cnt = reshape(axes(x, dim), ntuple(_->1, dim-1)..., :)
+        acc = 1 .+ cnt .* z .> acc
```
DhairyaLGandhi, line 70: Is it the bit mask that adds the instability? Also maybe use `one(T)` where T is inferred from the method signature

src/sparsemax.jl
```diff
@@ -0,0 +1,90 @@
...
+    tausum = sum(z .* mask; dims)  # no longer need z
+    kay = sum(mask; dims)
+    z = _relu.(x  .- (tausum .- 1) ./ kay)
+end
+
+
+
+function _sparsemax_mask(z::AbstractArray, dim::Integer)
+    acc = cumsum(z; dims=dim)
+    if dim == 1
```
DhairyaLGandhi, line 63: Ideally I'd avoid having branches at all in this function.

src/sparsemax.jl
```diff
@@ -0,0 +1,90 @@
...
+        acc = 1 .+ axes(z,1) .* z .> acc
+    elseif dim == 2
+        acc = 1 .+ axes(z,2)' .* z .> acc
```
DhairyaLGandhi, line 66: This adjoint call in here seems odd, we should be able to use the condition in `else` here I thought?

tylerjthomas9: Good catch, we don't need the if/else here

src/sparsemax.jl
tylerjthomas9: I changed the code to have a function for sorting x slightly differently if x is an AbstractVector instead of using multiple dispatch

src/sparsemax.jl
tylerjthomas9: I think a normal relu can be used/just `max.(x, 0)`

src/sparsemax.jl
tylerjthomas9: We get some type instability with the cnt. 

```julia
julia> z = rand(2,3,4);
julia> dim=1; cnt = reshape(axes(z, dim), ntuple(_->1, dim-1)..., :)
Base.OneTo(2)
julia> dim=2; cnt = reshape(axes(z, dim), ntuple(_->1, dim-1)..., :)
1×3 reshape(::Base.OneTo{Int64}, 1, 3) with eltype Int64:
 1  2  3
julia> dim=3; cnt = reshape(axes(z, dim), ntuple(_->1, dim-1)..., :)
1×1×4 reshape(::Base.OneTo{Int64}, 1, 1, 4) with eltype Int64:
[:, :, 1] =
 1

[:, :, 2] =
 2

[:, :, 3] =
 3

[:, :, 4] =
 4
```

mcabbott:
> Current version has the correct implementation

Trying to check the gradient numerically:
```
julia> x = rand(1:5, 5,10) ./ 5;

julia> delta = 0 .* x .+ randn.();

julia> ∇sparsemax(delta, sparsemax(x))  # cde15de
5×10 Matrix{Float64}:
 -0.0307726  -0.0315741  -0.197077   -0.0307726   …   0.0445302  -0.0307726  -0.0307726
 -0.498682    0.390107   -0.0632101  -0.532938       -0.0307726  -0.0307726  -0.456715
 -0.0307726  -0.0307726   1.13294    -0.00631409     -0.109703    0.194811   -0.534492
  0.275757   -0.0307726   0.132314    0.438519       -0.166756    0.0616505  -0.0307726
 -0.191945   -0.0307726  -0.0307726  -0.0307726      -0.0984743   0.0760821  -0.0307726

julia> ∇sm5(delta, sm5(x))
5×10 Matrix{Float64}:
  0.0       -0.526767  -1.57823    0.0       …   0.0        1.35172    0.0         0.0
 -0.488485   0.526767  -0.685784  -1.17488       0.0        0.0        0.0         0.0777773
  0.0        0.0        1.6463     0.268066      0.579856  -0.470059  -0.0775806  -0.0777773
  1.30843    0.0        0.61771    0.906813     -0.579856  -0.45652   -0.22328     0.0
 -0.819949   0.0        0.0        0.0           0.0       -0.425142   0.300861    0.0

julia> using ForwardDiff

julia> ForwardDiff.gradient(x -> sum(delta .* sparsemax(x)), x)
5×10 Matrix{Float64}:
  0.0       -0.526767  -1.57823    0.0       …  -0.0475105   1.35172    0.0         0.0
 -0.488485   0.526767  -0.685784  -1.17488       0.242721    0.0        0.0         0.0777773
  0.0        0.0        1.6463     0.268066      0.967818   -0.470059  -0.0775806  -0.0777773
  1.30843    0.0        0.61771    0.906813     -0.191893   -0.45652   -0.22328     0.0
 -0.819949   0.0        0.0        0.0          -0.971135   -0.425142   0.300861    0.0

julia> using FiniteDifferences, ZChop

julia> zchop(grad(central_fdm(5, 1), x -> sum(delta .* sm4(x)), x)[1], 1e-10)
5×10 Matrix{Float64}:
  0.0       -0.16144   -1.57823    0.0       …  -0.145158    1.35172    0.0         0.0
 -0.488485   0.892093  -0.685784  -1.17488      -0.0484138   0.0        0.0         0.0777773
  0.0        0.0        1.6463     0.268066      0.773837   -0.470059  -0.0775806  -0.0777773
  1.30843   -0.730653   0.61771    0.906813     -0.385874   -0.45652   -0.22328     0.0
 -0.819949   0.0        0.0        0.0          -0.453033   -0.425142   0.300861    0.0
```
Is it clear that the pytorch code gets this right? It's not so obvious to me what it's doing. If you are set up to do so, can you run them all on the same (tricky) input & compare?

The mathematics is simple but, as mentioned above, there will quite often be issues of which subgradient to choose. My function `_relu` tries to ensure that ForwardDiff picks the same one that is easy to implement using `y .> 0`, without storing its arguments.

Finite differencing will tend to average out sub-gradients, but need not produce something which is a valid choice at all at such points. I'm not sure which class the 2nd column in this example is in. The other columns are something of a sanity-check. This is what `test_rrule` will do, which is called by `gradtest`, but because of such issues it may not be usable for this function. 

Perhaps the test should just be that Zygote agrees with ForwardDiff. For such a test to be safe, I think it's best to have a function like `_relu` (it doesn't matter what it's called) so that this won't break if conventions change elsewhere. (No `relu`, no `max`. Also, `max` is weirdly slow.)

The current commit also omits many `.=` from `sm5` above, which means that it is not type-stable, and allocates twice as much memory. Why? And now removes much of the detail of the mask function, which wasn't there by accident -- see annotations now added above, and benchmarks.
```
julia> @code_warntype sparsemax(x)
Body::Any
1 ─ %1 = Main.:(var"#sparsemax#33")(1, #self#, x)::Any
└──      return %1

julia> @code_warntype sm5(x)
Body::Matrix{Float32}
1 ─ %1 = Main.:(var"#sm4#25")(1, #self#, x)::Matrix{Float32}
└──      return %1

julia> @test @inferred(sparsemax([1,2,3])) == [0,0,1]  # some of these can be in tests
ERROR: return type Vector{Float64} does not match inferred return type Any

julia> @test sparsemax([1,2,3]; dims=3) == [1,2,3]
ERROR: UndefVarError: z not defined
```

src/sparsemax.jl
mcabbott: No, these were there for performance, in the most common case.

src/sparsemax.jl
mcabbott: The mask creation is in a separate function precisely to contain any instability. It writes back into `acc` because this is an easy way to make its return type inferable (and to save allocations). And it avoids calling this reshape in the most common cases, as that made quite a difference to speed.

src/sparsemax.jl
mcabbott: The main reason not to write `max` or ordinary `relu` is "different gradient at zero" for ForwardDiff. See explanation below.

tylerjthomas9:
> I think it's best to have a function like _relu (it doesn't matter what it's called) so that this won't break if conventions change elsewhere. (No relu, no max. Also, max is weirdly slow.)
> My function _relu tries to ensure that ForwardDiff picks the same one that is easy to implement using y .> 0, without storing its arguments.
Great to know, thank you!
> The current commit also omits many .= from sm5 above, which means that it is not type-stable, and allocates twice as much memory. Why?
Reverted this, I am not sure why I removed it.
> Is it clear that the pytorch code gets this right? It's not so obvious to me what it's doing. If you are set up to do so, can you run them all on the same (tricky) input & compare?
I will test it vs the python results.

tylerjthomas9:
I think the gradient is working. I also went through the paper again, and we are calculating it exactly how the authors did. I have attached an image of the paper's derived jacobian. We are using `Δ` instead of `p` for the sparsemax outputs, `nonzeros` instead of `s` for the nonzero subset of `p`. Additionally, we are using `out` instead of `v`, and`total` instead of `\hat{v}'.
image

src/sparsemax.jl
tylerjthomas9: Without the promotion, `sparsemax` gives us a Real type for the output array. Here is an example with the promote:

```julia
julia> Random.seed!(42)
julia> x = rand(1:5, 5, 2) ./ 5
julia> sparsemax(x)
5×2 Matrix{Float64}:
 0.44  0.333333
 0.04  0.0
 0.44  0.333333
 0.04  0.333333
 0.04  0.0
```

Here is what happens when the promote is removed:
```julia
julia> sparsemax(x)
5×2 Matrix{Real}:
 0.44      0.333333
 0.04  false
 0.44      0.333333
 0.04      0.333333
 0.04  false
```


Will this Real type cause performance issues?

mcabbott:
No, `Δ` is the backward gradient we receive, `y` is the sparsemax output. 

I'm not sure I ever decoded the paper's notation. But the question of whether the gradient is correct doesn't require that -- the paper may have clues but this is simple enough that deriving it might be quicker than decoding them. Or the paper may have mistakes. The python code might too. 

We can check numerically. The easy case is small numbers, like `rand(3,4)/10`, where there should be no zeros in the output. I thought that large numbers (hence zeros) would be harder but am not sure today.

I am surprised that tests pass here. Why doesn't it call `test_rrule`, and fail? 
```
julia> using ChainRulesTestUtils

julia> test_rrule(sparsemax, rand(3,4))
nonzeros = Bool[1 1 1 1; 1 1 1 1; 1 1 1 1]
nonzeros = Bool[1 1 1 1; 1 1 1 1; 1 1 1 1]
test_rrule: sparsemax on Matrix{Float64}: Test Failed at /Users/me/.julia/packages/ChainRulesTestUtils/73Y9Q/src/check_result.jl:24
  Expression: isapprox(actual, expected; kwargs...)
   Evaluated: isapprox([-0.612288804854263 -1.3615411120031258 0.7652273269279669 -3.179013619758306; 0.11318318165635466 2.061264365438702 1.309074111596146 -0.9704853614354765; 0.3360509543675956 0.1273032261697271 -0.5543590617133406 1.9655847936080213], [-1.590000000000055 -5.723333333333301 -2.1000000000001666 -6.1266666666666945; 1.2299999999999827 4.456666666666624 2.0999999999998784 -1.6066666666666587; 0.36000000000001586 1.2666666666663948 -3.3757265396302482e-15 7.733333333333414]; rtol = 1.0e-9, atol = 1.0e-9)
```

PR #367, opened 2021-12-04
Fix direct Depthwise convolution implementation

staticfloat:
There was a silly typo bug in the direct depthwise convolution algorithm
that broke multi-channel testsets.  Perhaps more disturbingly, our
default test set did not catch this, because multi-channel workloads are
not run unless the fuzzing tests are run.

### DIFF
--- a/test/conv.jl
+++ b/test/conv.jl
@@ -242,6 +242,9 @@ conv_answer_dict = Dict(
     ),
 )
 
+# A "drop channels and batch dimension" helper
+ddims(x) = dropdims(x, dims=(ndims(x)-1, ndims(x)))
+
 @testset "Dense Convolution" begin
     # Start with some easy-to-debug cases that we have worked through and _know_ work
     for rank in (1,2,3)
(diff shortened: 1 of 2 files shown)


test/conv.jl
```diff
@@ -557,6 +556,14 @@ end
             end
         end
     end
+
+    # Do some real depthwise convolution tests
+    x = Float64.(reshape(1:2, (1,2,1)))
+    w = Float64.(reshape(1:6, (3,1,2)))
+    cdims = DepthwiseConvDims(x, w; padding=1)
+    for conv in (NNlib.depthwiseconv, NNlib.depthwiseconv_im2col, NNlib.depthwiseconv_direct)
+        @test isapprox(conv(x, w, cdims)[:], [2, 10], rtol=1e-7)
```
CarloLucibello, line 565:
```suggestion
        @test conv(x, w, cdims)[:] ≈ [2, 10]  rtol=1e-7
```

CarloLucibello (approved):

DhairyaLGandhi:
Thanks for catching this @staticfloat , maybe we should have an optional GHA action for fuzzing tests. It lgtm, let me know if you're happy and we can merge this.

Issue #19, opened 2017-12-21
BatchNorm and Dropout


dfdx:
Does it make sense to put functional forms of [BatchNorm and Dropout](https://github.com/FluxML/Flux.jl/blob/86097e76fdaa149b2caf815d5404c77d16c4f754/src/layers/normalisation.jl) into NNlib so that other packages could simply import them from here?

MikeInnes:
Yeah, that's a great idea.

staticfloat:
BatchNorm is going to be tricky because of its "statistics update step".  The current thinking behind this with Zygote is to do something along the lines of https://gist.github.com/staticfloat/a509b1e1cb1fb556028779722c2531e6

ToucheSir:
Now that Flux's normalization interface has been re-worked and GPU batchnorm moved from CUDA.jl -> NNlibCUDA, perhaps we should revisit this. The only reason https://github.com/FluxML/Flux.jl/tree/master/src/cuda exists at all now is to accommodate a non-standard implementation of batchnorm, so getting rid of that would be great.

CarloLucibello:
I was looking into porting the functional form of normalization layers here, but I'm not sure how to handle the `Zygote.ignore` [block](https://github.com/FluxML/Flux.jl/blob/06970a5fbbb1cb485c5d2cba597a78fb453fc713/src/layers/normalise.jl#L180) without having NNlib depend on Zygote

DhairyaLGandhi:
The concern has been raised earlier and is fixed by https://github.com/FluxML/Flux.jl/pull/1509

ToucheSir:
I don't think there's any reason these [`dropout` functions](https://github.com/FluxML/Flux.jl/blob/v0.12.4/src/layers/normalise.jl#L7-L47) need to live in Flux. Shall we move them over? Happy to volunteer for a copy-paste PR if we're all in agreement.

racinmat:
Any update on this? I can move it to NNlib if we want to do that.

CarloLucibello:
we have batchnorm and dropout now

PR #370, opened 2022-01-08
Fix convolution & pooling type-stability

pxl-th:
This PR fixes type-stability for convolutions and pooling (and closes https://github.com/FluxML/Flux.jl/issues/1178). It replaces current `ConvDims` structure and its descendants with the one that does not encode parameters in its type, but instead stores them as struct fields. This allows for the type inference and significantly improves compile times.

---

<details>
  <summary>Code used to benchmark:</summary>

```julia
function main()
    device = gpu

    x = device(zeros(Float32, 28, 28, 1, 5))
    y = device(zeros(Float32, 10, 5))
    m = device(Chain(
        Conv((3, 3), 1 => 2, relu),
        Conv((3, 3), 2 => 2, relu),
        Conv((3, 3), 2 => 3, relu),
        Conv((3, 3), 3 => 3, relu),
        Conv((3, 3), 3 => 3, relu),
        Conv((3, 3), 3 => 3, relu),
        x -> reshape(x, :, size(x, 4)),
        Dense(768, 10), softmax))
    θ = params(m)

    # F
    @time m(x)
    # FB
    BenchmarkTools.@btime $m($x)
    # B
    @time gradient(θ) do
        Flux.crossentropy(m(x), y)
    end
end
```
</details>

Benchmarks:

**Before:**

- CPU:
```julia
(F)  7.112869 seconds (31.64 M allocations: 1.649 GiB, 11.51% gc time, 99.94% compilation time)
(FB) 315.728 μs (387 allocations: 1.86 MiB)
(B)  33.839269 seconds (70.09 M allocations: 3.635 GiB, 3.73% gc time, 99.95% compilation time)
```
- GPU:
```julia
(F)  32.785819 seconds (63.27 M allocations: 3.296 GiB, 5.00% gc time, 51.70% compilation time)
(FB) 145.766 μs (617 allocations: 32.97 KiB)
(B)  47.313573 seconds (103.05 M allocations: 5.359 GiB, 5.16% gc time, 91.07% compilation time)
```

**After:**

- CPU:
```julia
(F)  2.725777 seconds (9.29 M allocations: 478.439 MiB, 6.42% gc time, 99.94% compilation time)
(FB) 285.336 μs (231 allocations: 1.86 MiB)
(B)  32.017971 seconds (64.42 M allocations: 3.362 GiB, 3.73% gc time, 99.97% compilation time)
```
- GPU:
```julia
(F)  26.145839 seconds (57.83 M allocations: 3.002 GiB, 4.76% gc time, 44.62% compilation time)
(FB) 129.965 μs (555 allocations: 29.50 KiB)
(B)  44.136454 seconds (98.40 M allocations: 5.114 GiB, 5.47% gc time, 90.51% compilation time)
```

---

Results from `@code_warntype`

<details>
  <summary>Before</summary>

```julia
julia> @code_warntype m(x)
MethodInstance for (::Chain{Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}})(::Array{Float32, 4})
  from (c::Chain)(x) in Flux at /home/pxl-th/.julia/dev/Flux/src/layers/basic.jl:49
Arguments
  c::Chain{Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}}
  x::Array{Float32, 4}
Body::Any
1 ─ %1 = Base.getproperty(c, :layers)::Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}
│   %2 = Flux.Tuple(%1)::Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}
│   %3 = Flux.applychain(%2, x)::Any
└──      return %3
```
</details>

<details>
  <summary>After</summary>

```julia
julia> @code_warntype m(x)
MethodInstance for (::Chain{Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}})(::Array{Float32, 4})
  from (c::Chain)(x) in Flux at /home/pxl-th/.julia/dev/Flux/src/layers/basic.jl:49
Arguments
  c::Chain{Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}}
  x::Array{Float32, 4}
Body::Matrix{Float32}
1 ─ %1 = Base.getproperty(c, :layers)::Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}
│   %2 = Flux.Tuple(%1)::Tuple{Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Conv{2, 4, typeof(relu), Array{Float32, 4}, Vector{Float32}}, Flux.var"#312#314", Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, typeof(softmax)}
│   %3 = Flux.applychain(%2, x)::Matrix{Float32}
└──      return %3
```
</detail>


DhairyaLGandhi:
I believe you'll need to use BenchmarkTools for the numbers. Which Julia version was the `Any` inferred with?

DhairyaLGandhi:
We probably don't want to have small Unions be part of the dispatch.

src/conv.jl
```diff
@@ -48,22 +48,27 @@ export conv, conv!, ∇conv_data, ∇conv_data!, ∇conv_filter, ∇conv_filter!
...
+Apply convolution filter `w` to input `x`. `x` and `w` are 3d/4d/5d tensors
+in 1d/2d/3d convolutions respectively.
 """
-function conv(x, w::AbstractArray{T, N}; stride=1, pad=0, dilation=1, flipped=false, groups = 1) where {T, N}
-    stride = expand(Val(N-2), stride)
-    pad = expand(Val(N-2), pad)
-    dilation = expand(Val(N-2), dilation)
-    cdims = DenseConvDims(x, w; stride=stride, padding=pad, dilation=dilation, flipkernel=flipped, groups = groups)
-    return conv(x, w, cdims)
+function conv(
```
DhairyaLGandhi, line 54: Best to use the existing function signature style across the PR

src/conv.jl
```diff
@@ -48,22 +48,27 @@ export conv, conv!, ∇conv_data, ∇conv_data!, ∇conv_filter, ∇conv_filter!
...
+    x, w::AbstractArray{T, N}; stride=1, pad=0, dilation=1,
+    flipped=false, groups=1,
```
DhairyaLGandhi, line 56: We also should use spaces around default values

DhairyaLGandhi (changes requested):
Could we move the syntax to match the current some? It's harder to look for the most important changes when there are unrelated changes in a PR. The more to the point the changes, the better

DhairyaLGandhi:
Re note: I believe in most cases that means that the inference is actually going fine.

pxl-th:
> Which Julia version was the `Any` inferred with?

I'm using 1.7.1. But issue is [present](https://github.com/FluxML/Flux.jl/issues/1178#issuecomment-759636910) on previous 
 versions as well.

In the PR description I've posted `@code_warntype` for the model, but here's also for the single convolutional layer.

**Before:**

```julia
julia> c = Conv(ones(Float32, 2, 2, 1, 3), ones(Float32, 3));
julia> x = zeros(Float32, 28, 28, 1, 5);
julia> @code_warntype c(x)
MethodInstance for (::Conv{2, 4, typeof(identity), Array{Float32, 4}, Vector{Float32}})(::Array{Float32, 4})
  from (c::Conv)(x::AbstractArray) in Flux at /home/pxl-th/.julia/dev/Flux/src/layers/conv.jl:163
Arguments
  c::Conv{2, 4, typeof(identity), Array{Float32, 4}, Vector{Float32}}
  x::Array{Float32, 4}
Locals
  #198::Flux.var"#198#199"
  cdims::DenseConvDims{2}
  b::Array{Float32, 4}
  σ::typeof(identity)
Body::Any
1 ─ %1  = Base.getproperty(c, :σ)::Core.Const(identity)
│   %2  = Base.getproperty(c, :bias)::Vector{Float32}
│   %3  = Core.tuple(%2)::Tuple{Vector{Float32}}
│         (#198 = %new(Flux.:(var"#198#199")))
│   %5  = #198::Core.Const(Flux.var"#198#199"())
│   %6  = Base.getproperty(c, :stride)::Tuple{Int64, Int64}
│   %7  = Flux.length(%6)::Core.Const(2)
│   %8  = Flux.ntuple(%5, %7)::Core.Const((1, 1))
│   %9  = Core.tuple(Flux.:(:), 1)::Core.Const((Colon(), 1))
│   %10 = Core._apply_iterate(Base.iterate, Flux.reshape, %3, %8, %9)::Array{Float32, 4}
│         (σ = %1)
│         (b = %10)
│   %13 = (:stride, :padding, :dilation, :groups)::Core.Const((:stride, :padding, :dilation, :groups))
│   %14 = Core.apply_type(Core.NamedTuple, %13)::Core.Const(NamedTuple{(:stride, :padding, :dilation, :groups)})
│   %15 = Base.getproperty(c, :stride)::Tuple{Int64, Int64}
│   %16 = Base.getproperty(c, :pad)::NTuple{4, Int64}
│   %17 = Base.getproperty(c, :dilation)::Tuple{Int64, Int64}
│   %18 = Base.getproperty(c, :groups)::Int64
│   %19 = Core.tuple(%15, %16, %17, %18)::Tuple{Tuple{Int64, Int64}, NTuple{4, Int64}, Tuple{Int64, Int64}, Int64}
│   %20 = (%14)(%19)::NamedTuple{(:stride, :padding, :dilation, :groups), Tuple{Tuple{Int64, Int64}, NTuple{4, Int64}, Tuple{Int64, Int64}, Int64}}
│   %21 = Core.kwfunc(Flux.DenseConvDims)::Core.Const(Core.var"#Type##kw"())
│   %22 = Base.getproperty(c, :weight)::Array{Float32, 4}
│         (cdims = (%21)(%20, Flux.DenseConvDims, x, %22))
│   %24 = σ::Core.Const(identity)
│   %25 = Base.getproperty(c, :weight)::Array{Float32, 4}
│   %26 = Flux.conv(x, %25, cdims)::AbstractArray{yT, 4} where yT
│   %27 = Base.broadcasted(Flux.:+, %26, b)::Any
│   %28 = Base.broadcasted(%24, %27)::Any
│   %29 = Base.materialize(%28)::Any
└──       return %29
```

**After:**

```julia
julia> c = Conv(ones(Float32, 2, 2, 1, 3), ones(Float32, 3));
julia> x = zeros(Float32, 28, 28, 1, 5);
julia> @code_warntype c(x)
MethodInstance for (::Conv{2, 4, typeof(identity), Array{Float32, 4}, Vector{Float32}})(::Array{Float32, 4})
  from (c::Conv)(x::AbstractArray) in Flux at /home/pxl-th/.julia/dev/Flux/src/layers/conv.jl:163
Arguments
  c::Conv{2, 4, typeof(identity), Array{Float32, 4}, Vector{Float32}}
  x::Array{Float32, 4}
Locals
  #198::Flux.var"#198#199"
  cdims::DenseConvDims{2, 2, 2, 4, 2}
  b::Array{Float32, 4}
  σ::typeof(identity)
Body::Array{Float32, 4}
1 ─ %1  = Base.getproperty(c, :σ)::Core.Const(identity)
│   %2  = Base.getproperty(c, :bias)::Vector{Float32}
│   %3  = Core.tuple(%2)::Tuple{Vector{Float32}}
│         (#198 = %new(Flux.:(var"#198#199")))
│   %5  = #198::Core.Const(Flux.var"#198#199"())
│   %6  = Base.getproperty(c, :stride)::Tuple{Int64, Int64}
│   %7  = Flux.length(%6)::Core.Const(2)
│   %8  = Flux.ntuple(%5, %7)::Core.Const((1, 1))
│   %9  = Core.tuple(Flux.:(:), 1)::Core.Const((Colon(), 1))
│   %10 = Core._apply_iterate(Base.iterate, Flux.reshape, %3, %8, %9)::Array{Float32, 4}
│         (σ = %1)
│         (b = %10)
│   %13 = (:stride, :padding, :dilation, :groups)::Core.Const((:stride, :padding, :dilation, :groups))
│   %14 = Core.apply_type(Core.NamedTuple, %13)::Core.Const(NamedTuple{(:stride, :padding, :dilation, :groups)})
│   %15 = Base.getproperty(c, :stride)::Tuple{Int64, Int64}
│   %16 = Base.getproperty(c, :pad)::NTuple{4, Int64}
│   %17 = Base.getproperty(c, :dilation)::Tuple{Int64, Int64}
│   %18 = Base.getproperty(c, :groups)::Int64
│   %19 = Core.tuple(%15, %16, %17, %18)::Tuple{Tuple{Int64, Int64}, NTuple{4, Int64}, Tuple{Int64, Int64}, Int64}
│   %20 = (%14)(%19)::NamedTuple{(:stride, :padding, :dilation, :groups), Tuple{Tuple{Int64, Int64}, NTuple{4, Int64}, Tuple{Int64, Int64}, Int64}}
│   %21 = Core.kwfunc(Flux.DenseConvDims)::Core.Const(Core.var"#Type##kw"())
│   %22 = Base.getproperty(c, :weight)::Array{Float32, 4}
│         (cdims = (%21)(%20, Flux.DenseConvDims, x, %22))
│   %24 = σ::Core.Const(identity)
│   %25 = Base.getproperty(c, :weight)::Array{Float32, 4}
│   %26 = Flux.conv(x, %25, cdims)::Array{Float32, 4}
│   %27 = Base.broadcasted(Flux.:+, %26, b)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{4}, Nothing, typeof(+), Tuple{Array{Float32, 4}, Array{Float32, 4}}}
│   %28 = Base.broadcasted(%24, %27)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{4}, Nothing, typeof(identity), Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{4}, Nothing, typeof(+), Tuple{Array{Float32, 4}, Array{Float32, 4}}}}}
│   %29 = Base.materialize(%28)::Array{Float32, 4}
└──       return %29
```

pxl-th:
> I believe you'll need to use BenchmarkTools for the Fand B numbers too.

I've used `@time` to get the measurements for the first run, which includes compilation. I believe `@btime` runs several times and does not include compilation times in its results (correct me if I'm wrong).

DhairyaLGandhi:
That's right, it would be helpful to see runtime differences.

pxl-th:
**Before:**

**CPU**
- Forward `@btime`: 
- Backward `@btime`: 

**GPU**
- Forward `@btime`: 
- Backward `@btime`: 

**After:**

**CPU**
- Forward `@btime`: 282.443 μs (231 allocations: 1.86 MiB)
- Backward `@btime`: 2.285 ms (2015 allocations: 5.79 MiB)

**GPU**
- Forward `@btime`: 
- Backward `@btime`:

pxl-th:
# Runtime differences

Timings above include runtime differences `(FB)`, but here's also for the backward pass.
Using `@btime` for the code in the PR description.

## **Before:**

**CPU**
- Forward: `313.752 μs (387 allocations: 1.86 MiB)`
- Backward: `2.211 ms (2453 allocations: 5.81 MiB)`

**GPU**
- Forward: `161.414 μs (621 allocations: 33.09 KiB)`
- Backward: `1.278 ms (3676 allocations: 459.58 KiB)`

## **After:**

**CPU**
- Forward: `282.443 μs (231 allocations: 1.86 MiB)`
- Backward: `2.285 ms (2015 allocations: 5.79 MiB)`

**GPU**
- Forward: `126.859 μs (555 allocations: 29.50 KiB)`
- Backward: `1.202 ms (3518 allocations: 447.41 KiB)`

src/conv.jl
pxl-th: Reverted to the existing style.

pxl-th:
> We probably don't want to have small Unions be part of the dispatch.

Maybe for now it is ok, since the runtime is not affected. But in the future PRs this can be removed, once other `ConvDims` are fixed

src/dim_helpers/DenseConvDims.jl
```diff
@@ -1,87 +1,103 @@
 export DenseConvDims
 
-"""
-    DenseConvDims
+struct DenseConvDims{N, K, S, P, D}
+    input_size::NTuple{N, Int}
```
ToucheSir, line 4:
```suggestion
struct DenseConvDims{N, K, S, P, D, F} <: ConvDims{N, S, P, D, F}
    input_size::NTuple{N, Int}
```
With this and my other suggestion on `flipkernel`, I think it should be possible to keep `DenseConvDims` a subtype of `ConvDims` while still maintaining type stability?

src/dim_helpers/DenseConvDims.jl
```diff
@@ -1,87 +1,103 @@
...
+    channels_out::Int
+    groupcount::Int
 
-# Convenience wrapper to create DenseConvDims objects
-function DenseConvDims(x_size::NTuple{M}, w_size::NTuple{M};
-                       stride=1, padding=0, dilation=1, flipkernel::Bool=false, groups = 1) where M
+    stride::NTuple{S, Int}
+    padding::NTuple{P, Int}
+    dilation::NTuple{D, Int}
+    flipkernel::Bool
```
ToucheSir, line 14:
```suggestion
    flipkernel::F
```
See comment thread above.

src/dim_helpers/DenseConvDims.jl
ToucheSir: Alternatively:
```suggestion
struct DenseConvDims{N, K, S, P, D} <: ConvDims{N, S, P, D, Bool}
```
I would favour the first version over this, however.

pxl-th: If we are ok with these parameters having different meaning.
For example in `DenseConvDims` parameter `S` is just the size of the stride tuple, while in `ConvDims` it is the stride tuple itself.

ToucheSir: Ah, I missed that. How about `struct DenseConvDims{N, K, S, P, D, F} <: ConvDims{Dims{N}, Dims{S}, Dims{P}, Dims{D}, F}` (`Dims{N} === NTuple{N, Int}`)?

pxl-th: If I understand you correctly, you still want to encode only type information?
Because right now (before this PR) parameters in  `ConvDims` and `DenseConvDims` are not types, but actual values.
So for the `S` parameter they would be not `NTuple{2, Int}` but `(2, 2)`, for example.

```julia
julia> x = zeros(Float32, 28, 28, 1, 5);
julia> w = rand(Float32, 3, 3, 1, 1);
julia> cdims = DenseConvDims(x, w)
DenseConvDims: (28, 28, 1) * (3, 3) -> (26, 26, 1), stride: (1, 1), pad: (0, 0, 0, 0), dil: (1, 1), flip: false, groups: 1
julia> typeof(cdims)
DenseConvDims{2, (3, 3), 1, 1, 1, (1, 1), (0, 0, 0, 0), (1, 1), false}
```

Therefore structurally, new `DenseConvDims` cannot have the same meaning for its type's parameters, because it does not encode actual values (like `(2, 2)` for the stride), but only type information (like size of the tuple).

ToucheSir: This is what I get for not properly reading through the codebase first :facepalm:, thanks for clarifying. Please disregard all my comments and LMK when this is ready to go!

pxl-th: Not a problem! After getting done with this PR I plan to work on fixing type-stability issues with other conv dims (that would include pooling and depthwise dims). So these `Union`s will not be there for long.

ToucheSir:
Since this is a breaking change, we can also tolerate the union since there will be follow-up PRs for other subtypes of ConvDims as well. Just need to make sure the union is gone before cutting a release :)

pxl-th:
Fixed type-inference for pooling and depthwise convolution as well.
Might be a lot, but this way we don't have to worry about multiple PRs and `Union`s.

**Side note:** wanted to make resnet model fully inferrable, but turns out, `BatchNorm` is also not inferrable:

```julia
julia> b = BatchNorm(1);
julia> x = zeros(Float32, 28, 28, 1, 1);
julia> @code_warntype b(x)
MethodInstance for (::BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}})(::Array{Float32, 4})
  from (BN::BatchNorm)(x) in Flux at /home/pxl-th/.julia/dev/Flux/src/layers/normalise.jl:269
Arguments
  BN::BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}
  x::Array{Float32, 4}
Locals
  #284::Flux.var"#284#285"{Array{Float32, 4}, Int64}
  affine_shape::NTuple{4, Int64}
  reduce_dims::Vector{Int64}
  N::Int64
Body::Any
1 ─       Core.NewvarNode(:(#284))
│         Core.NewvarNode(:(affine_shape))
│         Core.NewvarNode(:(reduce_dims))
│         Core.NewvarNode(:(N))
│   %5  = Flux.ndims(x)::Core.Const(4)
│   %6  = (%5 - 1)::Core.Const(3)
│   %7  = Flux.size(x, %6)::Int64
│   %8  = Base.getproperty(BN, :chs)::Int64
│   %9  = (%7 == %8)::Bool
└──       goto #3 if not %9
2 ─       goto #4
3 ─ %12 = Base.AssertionError("size(x, ndims(x) - 1) == BN.chs")::Any
└──       Base.throw(%12)
4 ┄       (N = Flux.ndims(x))
│   %15 = (N::Core.Const(4) - 2)::Core.Const(2)
│   %16 = (1:%15)::Core.Const(1:2)
│         (reduce_dims = Base.vcat(%16, N::Core.Const(4)))
│   %18 = Flux.:(var"#284#285")::Core.Const(Flux.var"#284#285")
│   %19 = Core.typeof(x)::Core.Const(Array{Float32, 4})
│   %20 = Core.typeof(N::Core.Const(4))::Core.Const(Int64)
│   %21 = Core.apply_type(%18, %19, %20)::Core.Const(Flux.var"#284#285"{Array{Float32, 4}, Int64})
│         (#284 = %new(%21, x, N::Core.Const(4)))
│   %23 = #284::Core.PartialStruct(Flux.var"#284#285"{Array{Float32, 4}, Int64}, Any[Array{Float32, 4}, Core.Const(4)])
│         (affine_shape = Flux.ntuple(%23, N::Core.Const(4)))
│   %25 = (:reduce_dims, :affine_shape)::Core.Const((:reduce_dims, :affine_shape))
│   %26 = Core.apply_type(Core.NamedTuple, %25)::Core.Const(NamedTuple{(:reduce_dims, :affine_shape)})
│   %27 = Core.tuple(reduce_dims, affine_shape::Core.PartialStruct(NTuple{4, Int64}, Any[Core.Const(1), Core.Const(1), Int64, Core.Const(1)]))::Core.PartialStruct(Tuple{Vector{Int64}, NTuple{4, Int64}}, Any[Vector{Int64}, Core.PartialStruct(NTuple{4, Int64}, Any[Core.Const(1), Core.Const(1), Int64, Core.Const(1)])])
│   %28 = (%26)(%27)::Core.PartialStruct(NamedTuple{(:reduce_dims, :affine_shape), Tuple{Vector{Int64}, NTuple{4, Int64}}}, Any[Vector{Int64}, Core.PartialStruct(NTuple{4, Int64}, Any[Core.Const(1), Core.Const(1), Int64, Core.Const(1)])])
│   %29 = Core.kwfunc(Flux._norm_layer_forward)::Core.Const(Flux.var"#_norm_layer_forward##kw"())
│   %30 = (%29)(%28, Flux._norm_layer_forward, BN, x)::Any
└──       return %30
```

pxl-th:
Here's another example with more layers:

```julia
x = device(zeros(Float32, 224, 224, 1, 5))
y = device(zeros(Float32, 10, 5))
m = device(Chain(
    Conv((3, 3), 1 => 2, relu),
    MaxPool((2, 2)),
    Conv((3, 3), 2 => 2, relu),
    MaxPool((2, 2)),
    Conv((3, 3), 2 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    MaxPool((2, 2)),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    Conv((3, 3), 3 => 3, relu),
    MaxPool((2, 2)),
    x -> reshape(x, :, size(x, 4)),
    Dense(147, 10), softmax))
θ = params(m)

@time m(x)
BenchmarkTools.@btime $m($x)

@time gradient(θ) do
    Flux.crossentropy(m(x), y)
end
BenchmarkTools.@btime gradient($θ) do
    Flux.crossentropy($m($x), $y)
end
```

Timings in the order they appear in code above.

**Before:**

- **CPU:**

```julia
8.336650 seconds (33.78 M allocations: 1.800 GiB, 9.91% gc time, 99.63% compilation time)
4.162 ms (803 allocations: 36.60 MiB)
59.726230 seconds (71.67 M allocations: 3.815 GiB, 2.51% gc time, 99.84% compilation time)
33.088 ms (4586 allocations: 107.94 MiB)
```

- **GPU:**

```julia
26.857145 seconds (68.45 M allocations: 3.557 GiB, 5.80% gc time, 56.44% compilation time)
309.557 μs (1225 allocations: 70.75 KiB)
67.174837 seconds (104.17 M allocations: 5.416 GiB, 3.66% gc time, 93.56% compilation time)
2.277 ms (6496 allocations: 1.11 MiB)
```

**After:**

- **CPU:**

```julia
2.510704 seconds (9.98 M allocations: 550.919 MiB, 7.70% gc time, 99.62% compilation time)
5.438 ms (446 allocations: 36.58 MiB)
42.624773 seconds (65.89 M allocations: 3.536 GiB, 3.14% gc time, 99.80% compilation time)
36.277 ms (3650 allocations: 107.93 MiB)
```

- **GPU:**

```julia
22.459811 seconds (56.46 M allocations: 2.928 GiB, 6.11% gc time, 51.20% compilation time)
263.607 μs (1066 allocations: 56.62 KiB)
59.634783 seconds (99.78 M allocations: 5.186 GiB, 4.05% gc time, 92.60% compilation time)
2.193 ms (6125 allocations: 1.11 MiB)
```

ToucheSir (commented):
RE `@inferred`, I noticed the tests pass now so is that no longer an issue? It's likely you were running into https://github.com/JuliaLang/julia/issues/23749, so I'd recommend Cthulhu.jl as the most reliable way to debug inference.

RE BatchNorm, the revised implementation in https://github.com/FluxML/Flux.jl/pull/1509 may be a starting point (if not a complete solution) for fixing inference. Another approach could be to first create a type stable functional version in NNlib (simultaneously addressing https://github.com/FluxML/NNlib.jl/issues/19) and then retrofitting the Flux layer on top of that.

src/dim_helpers/DenseConvDims.jl
```diff
@@ -1,87 +1,81 @@
...
     # Next, check that the spatial dimensions match up
-    @assert x[1:M-2] == input_size(cdims) DimensionMismatch("Data input spatial size ($(x[1:M-2]) vs. $(input_size(cdims)))")
+    @assert x[1:M-2] == cdims.input_size DimensionMismatch("Data input spatial size ($(x[1:M-2]) vs. $(input_size(cdims)))")
     @assert y[1:M-2] == output_size(cdims) DimensionMismatch("Data output spatial size ($(y[1:M-2]) vs. $(output_size(cdims)))")
-    @assert w[1:M-2] == kernel_size(cdims) DimensionMismatch("Kernel spatial size ($(w[1:M-2]) vs. $(kernel_size(cdims)))")
+    @assert w[1:M-2] == cdims.kernel_size DimensionMismatch("Kernel spatial size ($(w[1:M-2]) vs. $(kernel_size(cdims)))")
 
     # Check the groups match
-    @assert channels_in(cdims) % groupcount(cdims) == 0 DimensionMismatch("Groups ($(groupcount(cdims))) should be divisble by input channels $(channels_in(cdims))")
+    @assert cdims.channels_in % cdims.groupcount == 0 DimensionMismatch("Groups ($(groupcount(cdims))) should be divisble by input channels $(channels_in(cdims))")
```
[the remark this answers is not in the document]
pxl-th: Nevermind, probably remnants of something I've tried to do. Using methods now

src/dim_helpers/ConvDims.jl
```diff
@@ -6,34 +6,54 @@ export ConvDims
...
-abstract type ConvDims{N, S, P, D, F} end
+abstract type ConvDims{N} end
+
+@inline spatial_dims(::ConvDims{N}) where N = N
+@inline input_size(c::ConvDims) = c.input_size
+@inline kernel_size(c::ConvDims) = c.kernel_size
+
+@inline stride(c::ConvDims) = c.stride
+@inline padding(c::ConvDims) = c.padding
+@inline dilation(c::ConvDims) = c.dilation
```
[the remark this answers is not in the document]
pxl-th: Doesn't seem to affect much. Maybe slightly slower or maybe just noise.
But as for the default values, we have them in the constructors for other `...Dims`, so no sure they are needed here.
Otherwise we can add following defaults:

```julia
stride(c::ConvDims{N}) = NTuple{N, Int}(repeat(1, N))
padding(c::ConvDims{N}) = NTuple{N, Int}(repeat(0, N))
dilation(c::ConvDims{N}) = NTuple{N, Int}(repeat(1, N))
```

src/dim_helpers/ConvDims.jl
```diff
@@ -6,34 +6,54 @@ export ConvDims
...
+@inline flipkernel(c::ConvDims) = c.flipkernel
```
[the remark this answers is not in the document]
pxl-th: I've [tried](https://github.com/pxl-th/NNlib.jl/blob/fix-perf/src/impl/conv_im2col.jl#L8) adding `flipkernel` value to the type, not boolean like it was before, but `Val{true}` to keep type-stability.
However, the performance was pretty much the same. Maybe because previously the kernel size was also part of the type and now it has to [retrieve](https://github.com/FluxML/NNlib.jl/blob/master/src/impl/conv_im2col.jl#L6) this property every time from the struct.

pxl-th:
> RE @inferred, I noticed the tests pass now so is that no longer an issue? It's likely you were running into JuliaLang/julia#23749, > so I'd recommend Cthulhu.jl as the most reliable way to debug inference.

Strangely, it is working now... Added more inference tests.

src/dim_helpers/ConvDims.jl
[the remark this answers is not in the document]
ToucheSir: Ok, let's not worry about defaults then.

src/dim_helpers/ConvDims.jl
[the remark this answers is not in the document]
ToucheSir: Confirmed on my side as well, there's basically no difference with master so I think this was a dead end. Feel free to revert.

ToucheSir:
Ok, I think it's actually pooling, not `conv` that has regressed perf-wise:
```
# master
typeof(pdims) = PoolDims{2, (2, 2), (2, 2), (0, 0, 0, 0), (1, 1)}
  91.530 μs (13 allocations: 245.59 KiB)

# PR
typeof(pdims) = PoolDims{2, 2, 2, 4, 2}
  732.686 μs (7 allocations: 245.36 KiB)
```
A comparison of profiles between both showed https://github.com/FluxML/NNlib.jl/blob/v0.7.31/src/impl/pooling_direct.jl#L53 post-PR but not on master. My suspicion is that the loop is being unrolled, as demonstrated by this simple example:

```julia
function naive_pool(x::Vector{Float32}, ks::Int)
  @inbounds for x_i in 1:length(x) - ks
    m = x[x_i]
    for k_i in 1:ks
      m = max(m, x[x_i + k_i])
    end
    x[x_i] = m
  end
end

function naive_pool_val(x::Vector{Float32}, ks::Val{K}) where K
  ks = K
  for x_i in 1:length(x) - ks
    m = x[x_i]
    for k_i in 1:ks
      m = max(m, x[x_i + k_i])
    end
    x[x_i] = m
  end
end

x = rand(Float32, 1024)
@code_llvm naive_pool(x, 3) # not unrolled
@code_llvm naive_pool_val(x, Val(3)) # unrolled, multiple references to max
```
Since `conv_direct!` and `depthwiseconv_direct!` use similar inner loops, they may also be affected. WDYT about (re-)promoting the kernel shape and potentially input channels to type space? I wonder if doing that at the last possible moment (i.e. right before calling the impl function) would work...

ToucheSir:
I managed to shave off ~150μs by using `CartesianIndices` instead of the multi-range iterator, but that's still a good 6x slower than master. The conundrum is that Flux [doesn't keep track](https://github.com/FluxML/Flux.jl/blob/master/src/layers/conv.jl#L659-L674) of kernel size at the type level. If possible, I'd like to solicit @chriselrod's opinion on whether it's possible to bridge the performance gap while keeping type inference happy.

pxl-th:
Thanks for looking into it! I was suspecting pooling, but didn't have the chance to look at it then.
I've managed to slightly improve pooling performance by reducing number of things that we re-compute.

---
**Code:**
```julia
using BenchmarkTools
function main()
    x = rand(Float32, 224, 224, 3, 2)
    pdims = PoolDims(x, 3)

    y_max = maxpool(x, pdims)
    y_mean = meanpool(x, pdims)
    dy = ones(Float32, size(y_max)...)

    @btime maxpool($x, $pdims)
    @btime meanpool($x, $pdims)

    @btime ∇maxpool($dy, $y_max, $x, $pdims)
    @btime ∇meanpool($dy, $y_mean, $x, $pdims)

    nothing
end
```
---

**Master:**

```julia
62.949 μs (13 allocations: 128.97 KiB)
39.504 μs (13 allocations: 128.97 KiB)
1.060 ms (17 allocations: 1.15 MiB)
133.712 μs (17 allocations: 1.15 MiB)
```

**This PR:**

```julia
176.703 μs (7 allocations: 128.73 KiB)
154.341 μs (7 allocations: 128.73 KiB)
926.558 μs (11 allocations: 1.15 MiB)
303.342 μs (11 allocations: 1.15 MiB)
```

---

**[Model](https://github.com/FluxML/NNlib.jl/pull/370#issuecomment-1008196485):** 

**Master:**

```julia
4.162 ms (803 allocations: 36.60 MiB)
33.088 ms (4586 allocations: 107.94 MiB)
```

**This PR before latest optimization:**

```julia
5.438 ms (446 allocations: 36.58 MiB)
36.277 ms (3650 allocations: 107.93 MiB)
```

**This PR:**

```julia
4.540 ms (445 allocations: 36.58 MiB)
34.886 ms (3650 allocations: 107.93 MiB)
```

---

Also... do we need to multiply by `beta` [here](https://github.com/FluxML/NNlib.jl/blob/master/src/impl/pooling_direct.jl#L190)?
We do that for the padded regions. But since `beta` is not supported it doesn't matter right now.

src/impl/pooling_direct.jl
```diff
@@ -41,69 +41,80 @@ for name in (:max, :mean)
...
+                input_kh = ph + (kh - 1) * dil_h
+                input_kw = pw + (kw - 1) * dil_w
 
                 # This conditional will be optimized away at compile time
                 if $(name == :max)
-                    m = max(m, x[input_kw, input_kh, input_kd, c, batch_idx])
+                    xv = x[input_kw, input_kh, input_kd, c, batch_idx]
+                    if xv > m
+                        m = xv
+                    end
```
ToucheSir, line 67: Is this faster? If so, might `m = ifelse(xv > m, xv, m)` be even faster?

src/impl/pooling_direct.jl
```diff
@@ -41,69 +41,80 @@ for name in (:max, :mean)
...
+                ph = project(h, stride_h, pad_h_lo)
+                for w in w_region
+                pw = project(w, stride_w, pad_w_lo)
                 m = m_init
+
                 for kd in 1:kernel_d
-                    input_kd = project(d, stride_d, pad_d_lo) + (kd - 1) * dil_d
-                    if input_kd <= 0 || input_kd > depth
-                        # add here condition for handling options for paded value handling  
+                    input_kd = pd + (kd - 1) * dil_d
```
ToucheSir, line 93: Not necessary for this PR, but it would be nice if we could fold this calculation into the range somehow.

chriselrod:
> I managed to shave off ~150μs by using `CartesianIndices` instead of the multi-range iterator, but that's still a good 6x slower than master. The conundrum is that Flux [doesn't keep track](https://github.com/FluxML/Flux.jl/blob/master/src/layers/conv.jl#L659-L674) of kernel size at the type level. If possible, I'd like to solicit @chriselrod's opinion on whether it's possible to bridge the performance gap while keeping type inference happy.

As the kernel size isn't normally going to be changing within a model, I think we should store it at model definition in order to make `DenseConvDims` type stable, without giving up runtime performance.

ToucheSir:
Ok, the dims constructors will definitely need a rework then. https://github.com/FluxML/NNlib.jl/blob/c30ea9bf9d024adfeb99bf10fb8a1e91368ca8ea/src/dim_helpers/PoolDims.jl#L32 in particular is an absolute black hole for type instability, and I haven't yet figured yet how to make it behave...

pxl-th:
> I think we should store it at model definition in order to make DenseConvDims type stable, without giving up runtime performance.

Just to confirm that I understand correctly, that parameters would be in dims type as before, but they will be passed onto them from, for example, `MaxPool` from Flux instead of calculated in the dims constructor?

> Ok, the dims constructors will definitely need a rework then.
> 
> https://github.com/FluxML/NNlib.jl/blob/c30ea9bf9d024adfeb99bf10fb8a1e91368ca8ea/src/dim_helpers/PoolDims.jl#L32
> 
> in particular is an absolute black hole for type instability, and I haven't yet figured yet how to make it behave...

In this PR I did fix type-instability for `PoolDims` constructor, maybe some of it can be used in the rework.

DhairyaLGandhi:
Pinging @staticfloat for his eyes on the type stability as well.

pxl-th:
With the latest changes I've managed to nearly match the performance.
Backwards for the maxpooling is actually greatly improved.
Essentially I'm doing what @ToucheSir had suggested earlier is re-promote `kernel_size` to type right before calling the kernel.

**Master:**

```julia
62.949 μs (13 allocations: 128.97 KiB)
39.504 μs (13 allocations: 128.97 KiB)
1.060 ms (17 allocations: 1.15 MiB)
133.712 μs (17 allocations: 1.15 MiB)
```

**Latest commit:**

```julia
77.597 μs (9 allocations: 128.92 KiB)
74.611 μs (9 allocations: 128.92 KiB)
796.221 μs (13 allocations: 1.15 MiB)
124.385 μs (13 allocations: 1.15 MiB)
```

**[Model](https://github.com/FluxML/NNlib.jl/pull/370#issuecomment-1008196485):**

First evaluation timings are slightly slower than they were, but still much faster than on master.
I think this is because more optimizations are applied to the function... idk :)
Alternatively, because it gets compiled every time it is called with different `kernel_size`, which is not a case for the given model, though, but still can cause increased compilation times.

**Master:**

```julia
8.336650 seconds (33.78 M allocations: 1.800 GiB, 9.91% gc time, 99.63% compilation time)
4.162 ms (803 allocations: 36.60 MiB)
59.726230 seconds (71.67 M allocations: 3.815 GiB, 2.51% gc time, 99.84% compilation time)
33.088 ms (4586 allocations: 107.94 MiB)
```

**Latest commit:**

```julia
3.406770 seconds (10.15 M allocations: 560.621 MiB, 7.04% gc time, 99.77% compilation time)
3.839 ms (453 allocations: 36.58 MiB)
58.345734 seconds (63.01 M allocations: 3.400 GiB, 2.49% gc time, 99.91% compilation time)
32.999 ms (3665 allocations: 107.93 MiB)
```

pxl-th:
With the latest changes I've managed to nearly match the performance.
Backwards for the maxpooling is actually greatly improved.
Essentially I'm doing what @ToucheSir had suggested earlier: re-promoting `kernel_size` to type right before calling the kernel.

**[Pooling benchmarking](https://github.com/FluxML/NNlib.jl/pull/370#issuecomment-1008798642)**:

- **Master:**

```julia
62.949 μs (13 allocations: 128.97 KiB)
39.504 μs (13 allocations: 128.97 KiB)
1.060 ms (17 allocations: 1.15 MiB)
133.712 μs (17 allocations: 1.15 MiB)
```

- **Latest commit:**

```julia
77.597 μs (9 allocations: 128.92 KiB)
74.611 μs (9 allocations: 128.92 KiB)
796.221 μs (13 allocations: 1.15 MiB)
124.385 μs (13 allocations: 1.15 MiB)
```

**[Model](https://github.com/FluxML/NNlib.jl/pull/370#issuecomment-1008196485):**

First evaluation timings are slightly slower than they were, but still much faster than on master.
I think this is because more optimizations are applied to the function... idk :)
Alternatively, because it gets compiled every time it is called with different `kernel_size`, which is not a case for the given model, though, but still can cause increased compilation times.

- **Master:**

```julia
8.336650 seconds (33.78 M allocations: 1.800 GiB, 9.91% gc time, 99.63% compilation time)
4.162 ms (803 allocations: 36.60 MiB)
59.726230 seconds (71.67 M allocations: 3.815 GiB, 2.51% gc time, 99.84% compilation time)
33.088 ms (4586 allocations: 107.94 MiB)
```

- **Latest commit:**

```julia
3.406770 seconds (10.15 M allocations: 560.621 MiB, 7.04% gc time, 99.77% compilation time)
3.839 ms (453 allocations: 36.58 MiB)
58.345734 seconds (63.01 M allocations: 3.400 GiB, 2.49% gc time, 99.91% compilation time)
32.999 ms (3665 allocations: 107.93 MiB)
```

src/impl/conv_im2col.jl
```diff
@@ -1,14 +1,14 @@
 ## This file contains im2col-backed implementations of convolution for 2d and 3d
 ## convolutions.  Expect to see a lot of indexing.
 
-# Helper functions for flipkernel-induced dyslexia
-@inline function kernel_index(w, h, d, cdims::ConvDims{N, S, P, D, false}) where {N, S, P, D}
+# Helper function for flipkernel-induced dyslexia
+function kernel_index(w, h, d, cdims::ConvDims)
+    if flipkernel(cdims)
+        return (w, h, d)
+    end
```
ToucheSir, line 8:
```suggestion
    flipkernel(cdims) && return (w, h, d)
```
Might as well save a couple of lines :)

ToucheSir:
Stupendous, thank you @pxl-th for taking this on! I have a couple of tiny non-behavioural suggestions, but other than that I think this is ready to go in :)

src/impl/conv_direct.jl
```diff
@@ -53,18 +53,16 @@ function conv_direct!(y::AbstractArray{yT,5}, x::AbstractArray{xT,5},
...
     stride_w, stride_h, stride_d = stride(cdims)
-    out_width, out_height, out_depth = output_size(cdims)
-    
-    # Create a method that, at compile-time, determines how we're going to index into `w`
-    kproj(k, M, cdims::ConvDims{N,S,P,D,true}) where {N, S, P, D} = k
-    kproj(k, M, cdims::ConvDims{N,S,P,D,false}) where {N, S, P, D} = M - k + 1
-    
+
+    # Create a method that determines how we're going to index into `w`.
+    kproj(k, M, cdims::DenseConvDims) = flipkernel(cdims) ? k : (M - k + 1)
```
staticfloat, line 61: I don't have the data to back this up, but I kind of thought it would be better if the compiler were able to, at compile time, determine which way `kproj()` should be calculated.... Is it really faster to compare the boolean every time rather than have an inlineable function?

src/impl/pooling_direct.jl
```diff
@@ -41,69 +51,80 @@ for name in (:max, :mean)
...
+                    end
                 elseif $(name == :mean)
                     m += x[input_kw, input_kh, input_kd, c, batch_idx]
                 else
                     error("Unimplemented codegen path")
                 end
             end
-            y[w, h, d, c, batch_idx] = alpha * m + beta * y[w, h, d, c, batch_idx]
+
+            y[w, h, d, c, batch_idx] = alpha * m # + beta * y[w, h, d, c, batch_idx]
```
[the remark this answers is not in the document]
ToucheSir: IIRC it was a reaction to this line: https://github.com/FluxML/NNlib.jl/blob/master/src/impl/pooling_direct.jl#L9.

src/impl/conv_direct.jl
ToucheSir: This was experimented with in https://github.com/FluxML/NNlib.jl/pull/370#discussion_r780782844, and the conclusion then was that there was no noticeable difference. Whether that's changed after the latest round of commits I'm not sure.

ToucheSir: GH doesn't jump to the link in the resolved review thread, so here's a screenshot: 
image

src/impl/pooling_direct.jl
pxl-th: My assumption was that if `xv < m`, then we can avoid assignment at all

src/dim_helpers/ConvDims.jl
```diff
@@ -6,34 +6,54 @@ export ConvDims
...
+function padding(::ConvDims) end
+function dilation(::ConvDims) end
+function flipkernel(::ConvDims) end
 
-# Hack to get rid of type parameters
 function basetype(::Type{C}) where {C <: ConvDims}
-    if C <: DepthwiseConvDims
-        return DepthwiseConvDims
-    elseif C <: DenseConvDims
+    if C <: DenseConvDims
```
DhairyaLGandhi, line 24: is it necessary to change this order?

src/conv.jl
```diff
@@ -86,8 +88,9 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
             function $(Symbol("$(name)$(backend)"))(
                             x::AbstractArray{xT,N}, w::AbstractArray{wT,N},
                             cdims::ConvDims; kwargs...) where {xT, wT, N}
-                y = similar(x, promote_type(xT, wT), output_size(cdims)...,
-                               channels_out(cdims), size(x,N))
+                y = similar(
+                   x, promote_type(xT, wT), output_size(cdims)...,
+                   channels_out(cdims), size(x,N))
```
DhairyaLGandhi, line 93: Best if we can revert the syntax changes that aren't actually code changes. It makes it easier to remain consistent across the package and focus the PRs on the relevant changes.

src/dim_helpers/ConvDims.jl
```diff
@@ -6,34 +6,54 @@ export ConvDims
...
 `im2col!()` to generate efficient code, and helpful to reduce the number of kwargs
 getting passed around.
-
-We don't want to specialize on things like image size/channel count, so we generally
-store those as fields, just for convenience, and to allow for non-breaking changes when
-we decide we _do_ want to specialize on those values.  We always want to specialize on
-things like stride, padding, dilation, and kernel flipping though.
 """
-abstract type ConvDims{N, S, P, D, F} end
+abstract type ConvDims{N} end
```
DhairyaLGandhi, line 10: Are we expecting the compilation improvements to mostly come from removing the type params here?

src/impl/pooling_direct.jl
```diff
@@ -41,69 +51,80 @@ for name in (:max, :mean)
...
-          
-            # Initialize `m` to `0.0`, or `typemin(T)` or whatever.
+
+        @inbounds for batch_idx in 1:size(x, 5), c in 1:out_c
+            for d in d_region
+            pd = project(d, stride_d, pad_d_lo)
+            for h in h_region
+            ph = project(h, stride_h, pad_h_lo)
+            for w in w_region
+            pw = project(w, stride_w, pad_w_lo)
```
DhairyaLGandhi, line 67: is this change different from what is already the case on master? It looks similar, but its harder to read.

pxl-th:
I just went on and added back other parameters as `Val{...}` and it now matches or surpasses the performance on master.
Backward passes for the pooling didn't see any improvements from this, so I've left them as is.

`pooling_direct`

- **Master:**
```julia
62.577 μs (13 allocations: 128.97 KiB)
39.254 μs (13 allocations: 128.97 KiB)
1.051 ms (17 allocations: 1.15 MiB)
123.622 μs (17 allocations: 1.15 MiB)
```

- **This PR:**
```julia
40.165 μs (13 allocations: 129.06 KiB)
  40.146 μs (13 allocations: 129.06 KiB)
  825.402 μs (13 allocations: 1.15 MiB)
  134.233 μs (13 allocations: 1.15 MiB)
```

`conv_direct`

- **Master:**
```julia
3.014 ms (15 allocations: 1.13 MiB)
10.432 ms (29 allocations: 3.43 MiB)
2.993 ms (26 allocations: 1.15 MiB)
```

- **This PR:**
```julia
2.760 ms (15 allocations: 1.13 MiB)
10.396 ms (23 allocations: 3.43 MiB)
2.881 ms (19 allocations: 1.15 MiB)
```


<details>
  <summary>Code:</summary>
  
```julia
function main()
    x = rand(Float32, 224, 224, 3, 2)
    w = rand(Float32, 3, 3, 3, 3)
    pdims = PoolDims(x, 3)
    cdims = DenseConvDims(x, w)

    y_max = maxpool(x, pdims)
    y_mean = meanpool(x, pdims)
    dy = ones(Float32, size(y_max)...)

    @btime maxpool($x, $pdims)
    @btime meanpool($x, $pdims)

    @btime ∇maxpool($dy, $y_max, $x, $pdims)
    @btime ∇meanpool($dy, $y_mean, $x, $pdims)

    dy_conv = conv_direct(x, w, cdims)

    @btime conv_direct($x, $w, $cdims)
    @btime ∇conv_filter_direct($x, $dy_conv, $cdims)
    @btime ∇conv_data_direct($dy_conv, $w, $cdims)

    nothing
end
```
</details>

---

> First evaluation is slightly slower than it was, but still much faster than on master.

Could've been just noise, because right now I'm seeing timings around `2.7`-`2.9` seconds.

src/impl/pooling_direct.jl
```diff
@@ -41,69 +51,80 @@ for name in (:max, :mean)
...
+                        # add here condition for handling options for paded value handling
                         continue
                     end
 
                     for kh in 1:kernel_h
-                        input_kh = project(h, stride_h, pad_h_lo) + (kh - 1) * dil_h
-                        if input_kh <= 0 || input_kh > height
-                            # add here condition for handling options for paded value handling  
+                        input_kh = ph + (kh - 1) * dil_h
+                        if input_kh < 1 || input_kh > height
```
DhairyaLGandhi, line 117: Is there a reason to change this conditional?

src/impl/conv_direct.jl
pxl-th: Latest commit reverts this change

src/impl/pooling_direct.jl
pxl-th: Actually no, reverted.

src/dim_helpers/ConvDims.jl
pxl-th: No, this was due to some intermediate experimentation. Reverted.

src/impl/pooling_direct.jl
pxl-th: On the master `project(d, stride_d, pad_d_lo)` gets computed every iteration, while here it is computed only when `d` changes.

src/dim_helpers/ConvDims.jl
pxl-th: By removing these type parameters (which were essentially runtime values) we make `Dims` as well as everything "above" that uses it type-stable.
Instead of storing these parameters in the `Dims` type we re-promote them to `Val{...}` and dispatch conv/pooling implementations on them. So in some sense, we moved these type parameters "down" the stack where they don't impact type-stability but still provide the same performance benefits.

CarloLucibello:
@ToucheSir do you have any further comments? otherwise looks good to go

ToucheSir (approved):

src/dim_helpers/ConvDims.jl
```diff
@@ -6,13 +6,19 @@ export ConvDims
...
+@inline spatial_dims(::ConvDims{N}) where N = N
+@inline groupcount(c::ConvDims) = 1
+
+function input_size(::ConvDims) end
+function kernel_size(::ConvDims) end
+
+function stride(::ConvDims) end
+function padding(::ConvDims) end
+function dilation(::ConvDims) end
+function flipkernel(::ConvDims) end
```
mcabbott, line 21: What are these methods for? Is nothing being used somewhere to mean "unknown"? I don't see `=== nothing` but did not look everywhere.

(It might also be less mysterious to write them `input_size(::ConvDims) = nothing` etc, if the are needed.)

src/dim_helpers/ConvDims.jl
```diff
@@ -81,10 +103,10 @@ function check_spdf(x_size::NTuple{N}, w_size::NTuple{N}, stride, padding, dilat
 
     # padding is kind of a special case; we allow it to be either 2-length or 4-length,
     # since we support asymmetrical padding
-    if length(ppadding) != 2*nd
+    if length(ppadding) != 2 * nd
         if length(ppadding) == nd
             # Do this repeat dance so that we get lo/hi symmetrical padding
-            ppadding = tuple(repeat(collect(ppadding), inner=2)...)
+            ppadding = NTuple{2 * nd, Int}(repeat(collect(ppadding), inner=2))
```
mcabbott, line 109: I doubt it really matters but seems a shame to make small arrays when you just need tuples. How easy would it be to avoid `collect`?

mcabbott (commented):
I haven't read all of this closely, but the numbers look good. Two questions:

src/dim_helpers/ConvDims.jl
pxl-th: Tried implementing it with tuple-peeling, but turned out to be slower.

src/dim_helpers/ConvDims.jl
pxl-th: I guess it was meant to be a "fallback" (or interface indication) if someone is using `ConvDims` only.
IMO it would make sense to either remove them completely or remove `(::ConvDims)` part.

We can leave `function input_size end` in code to indicate "interface" that dims should implement as was suggested in [JuliaLang Antipatterns](https://www.oxinabox.net/2020/04/19/Julia-Antipatterns.html#notimplemented-exceptions).

src/dim_helpers/ConvDims.jl
mcabbott: Do you mean this line narrowly, or the whole thing? For just this, one attempt is:
```
julia> pad2(pp, nd=length(pp)) = NTuple{2 * nd, Int}(repeat(collect(pp), inner=2));

julia> @btime pad2($((1,2,3,)))
  min 569.745 ns, mean 600.348 ns (6 allocations, 432 bytes)
(1, 1, 2, 2, 3, 3)

julia> pad3(pp) = ntuple(i -> pp[(i-1)÷2+1], 2*length(pp));

julia> @btime pad3($((1,2,3,)))
  min 1.458 ns, mean 1.555 ns (0 allocations)
(1, 1, 2, 2, 3, 3)
```

mcabbott: Much as I like 1-based indexing, I'm amazed that worked first time. But more fair timing might be:
```
julia> pad2(pp) = NTuple{2 * length(pp), Int}(repeat(collect(pp), inner=2));

julia> @btime pad2($((1,2,3,)))
  min 69.502 ns, mean 76.486 ns (2 allocations, 192 bytes)
(1, 1, 2, 2, 3, 3)
```

src/dim_helpers/ConvDims.jl
mcabbott: Ok. Zero-method stubs are better as documentation, `function input_size end`. Still a little bit of a puzzle, since you wonder whether they are being defined e.g. in the root module with methods added from elsewhere. IMO if you were ever going to extend this, you'd look at one of the working examples, which defines all of them together, to figure out what's needed.

src/dim_helpers/ConvDims.jl
pxl-th: Didn't think about using `ntuple` function... :) Using it right now.

pxl-th:
Pinging to see if there are still things that need to be addressed

darsnack:
Haven't followed this thread closely. Are we waiting to merge cause it is breaking?

ToucheSir:
Are there any PRs we want to flush out the queue before a breaking release? If not, I vote to merge.

pxl-th:
> Haven't followed this thread closely. Are we waiting to merge cause it is breaking?

It is breaking in the sense that `ConvDims` type parameters do not mean the same thing.
However method-wise it is non-breaking.

DhairyaLGandhi:
Let's merge this, and release after updates to the various parts of the ecosystem that specialise on the type params to avoid breakages.

DhairyaLGandhi:
Seems like there are unintentional breakages https://buildkite.com/julialang/nnlibcuda-dot-jl/builds/149#b44502db-331d-457c-898e-db25d0405df4

pxl-th:
> Seems like there are unintentional breakages https://buildkite.com/julialang/nnlibcuda-dot-jl/builds/149#b44502db-331d-457c-898e-db25d0405df4

I see it is now fixed, thanks @DhairyaLGandhi!

PR #372, opened 2022-01-12
Expanded usage of softmax in docs.

rkube:
Softmax acts on a vector instead of a scalar. It would be good for the docs to reflect the slightly different usage.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -29,7 +29,7 @@ a sketch of a convolutional model with a softmax classifier head can be implemen
 ```jldoctest `
 julia> X = randn(Float32, 4, 4, 3, 9);
 
-julia> model = Chain(Conv((4, 4), 3 => 8), x -> flatten(x), x -> softmax(x));
+julia> model = Chain(Conv((4, 4), 3 => 8), flatten, softmax);
 
 julia> size(model(X))
 (8, 9)
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -22,6 +22,20 @@ with additional manipulations enhancing numerical stability.
 For a matrix input `x` it will by default (`dims = 1`) treat it as a batch of vectors,
 with each column independent. Keyword `dims = 2` will instead treat rows independently, and so on.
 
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):
Thanks @rkube for the nice docs clarifications, keep 'em coming if you can!

DhairyaLGandhi (approved):
bors r+

Thanks!

src/softmax.jl
```diff
@@ -22,6 +22,20 @@ with additional manipulations enhancing numerical stability.
 For a matrix input `x` it will by default (`dims = 1`) treat it as a batch of vectors,
 with each column independent. Keyword `dims = 2` will instead treat rows independently, and so on.
 
+Since softmax by default acts on a vector instead of a scalar, a syntax different from scalar
```
mcabbott, line 25: This is great but should say "array" not vector, I think.

mcabbott: Maybe it's OK, thinking more. 

```suggestion
Since softmax by default acts on whole columns of array, not individual nubmers, a syntax different that for from ordinary scalar
```

src/softmax.jl
```diff
@@ -22,6 +22,20 @@ with additional manipulations enhancing numerical stability.
...
+activation functions is to be used when using it in a layer of a neural network. For example,
+a sketch of a convolutional model with a softmax classifier head can be implemented like this:
+
+```jldoctest `
+julia> X = randn(Float32, 4, 4, 3, 9);
+
+julia> model = Chain(Conv((4, 4), 3 => 8), flatten, softmax);
+
+julia> size(model(X))
+(8, 9)
```
mcabbott, line 35: Also, I think this sketch should go below the basic examples. 

`softmax` accepts any array & can be used without Flux. We should show what it does, and then expand by pointing at how to use it with other things.

mcabbott: Also, I'm surprised tests pass with this `jldoctest`, since this does not load Flux. Maybe they aren't being run?

https://github.com/FluxML/NNlib.jl/blob/master/test/runtests.jl#L10-L18

PR #371, opened 2022-01-11
Improve activation functions, make fast versions accessible

mcabbott:
This upgrades many of the activation functions. For instance, the new `gelu` is about 7x faster (when broadcast), and has worst-case error `nextfloat(true, 9)` instead of 7675 before. You can check my working below if so inclined.

It also defines a function which does `fast_act(tanh) == tanh_fast` etc, to allow Flux to automatically substitute the faster-than-Base variants. For all except `tanh` and `sigmoid`, however, there seems to be no point in keeping the slow versions around at all. 

The exception is `softplus`, for which a fast version is surely possible, but not worked out yet, so not included here.

Paging @oscardssmith, since we discussed things in https://github.com/FluxML/NNlib.jl/pull/345

<details>

```julia
using NNlib, BenchmarkTools, Statistics
# From the tests:
function countepsfrom(x::T, xtrue) where {T<:AbstractFloat}
    target = T(xtrue)
    for n in Iterators.flatten(zip(0:100, -1:-1:-100))
        nextfloat(x, n) === target && return n
    end
    return round(Int, (target - x) / eps(x))
end

mean_eps(f, g, xs) = mean(x -> abs(countepsfrom(f(x), g(big(x)))), xs)
worst_eps(f, g, xs) = maximum(x -> abs(countepsfrom(f(x), g(big(x)))), xs)
function find_worst(f, g, xs)
    c, i = findmax(x -> abs(countepsfrom(f(x), g(big(x)))), xs)
    c, xs[i]
end
# Also needed:
mean_abs(f, g, xs) = mean(x -> abs(f(x) - g(big(x))), xs) |> Float32
oftf(x, y) = oftype(float(x), y)

# In the same order as activations.jl:

########## ELU

elu_old(x, α=1) = ifelse(x ≥ 0, float(x), α * (exp(x) - 1))

deriv_elu(Ω, α=1) = ifelse(Ω ≥ 0, one(Ω), Ω + α)

elu_fast(x, α=1) = ifelse(x ≥ 0, float(x), @fastmath α * (exp(x) - 1))

#=

# No downside, just replace:

julia> @btime y .= elu_old.(x)   setup=(x=randn(Float32,1000); y=similar(x));
  6.175 μs (0 allocations: 0 bytes)

julia> @btime y .= elu_fast.(x)   setup=(x=randn(Float32,1000); y=similar(x));
  871.268 ns (0 allocations: 0 bytes)

julia> mean_abs(elu_old, elu_old, -4:0.01f0:4)
[...]

PR #375, opened 2022-01-12
Widen `tanh_fast` signature

mcabbott:
As noted here https://github.com/FluxML/NNlib.jl/pull/371#issuecomment-1011411925 defining as a fallback only `tanh_fast(::Real)` causes problems for Flux right now. And `tanh` of course makes sense for complex numbers.

But really, all the other activation functions only make sense for real numbers, since they often compare `x>0` etc. In the interests of sanity and sensible guard-rails, they should probably all have signature `::Real` not `::Any`. But can't be changed just yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ version = "0.7.33"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
-ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -14,7 +13,6 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Adapt = "2, 3.2"
-ChainRules = "1 - 1.17"
 ChainRulesCore = "0.9.45, 0.10, 1"
 Compat = "3.14"
 Requires = "0.5, 1.0"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NNlib"
 uuid = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
-version = "0.7.32"
+version = "0.7.33"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


mcabbott:
CI fails on unrelated things which worked yesterday?
```
Test Summary:        | Pass  Total
223
Activation Functions | 1329   1329
224
Test Summary:          | Pass  Total
225
Batched Multiplication | 3327   3327
226
[ Info: Skipping Convolutional fuzzing tests, set NNLIB_TEST_FUZZING=true to run them
227
[ Info: Skipping Depthwise Convolutional fuzzing tests, set NNLIB_TEST_FUZZING=true to run them
228
Grouped Convolutions: Error During Test at /home/runner/work/NNlib.jl/NNlib.jl/test/conv.jl:659
229
230
  type Nothing has no field method
231
  Stacktrace:
232
    [1] getproperty
233
      @ ./Base.jl:42 [inlined]
234
    [2] matching_cr_sig(t::IRTools.Inner.Meta, s::Nothing)
235
      @ Zygote ~/.julia/packages/Zygote/umM0L/src/compiler/chainrules.jl:72
236
    [3] has_chain_rrule(T::Type)
237
      @ Zygote ~/.julia/packages/Zygote/umM0L/src/compiler/chainrules.jl:54
238

```

darsnack:
Getting the same issues locally on a completely separate project (getting Metalhead pre-training running).

darsnack:
The only thing I can see that has changed in this timeframe is ChainRules.jl/ChainRulesCore.jl but those changes seem completely orthogonal to the error.

mcabbott:
Thanks, that's useful. If just pushed an upper bound to exclude ChainRules v1.18.0, today's version, just to see?

Project.toml
```diff
@@ -13,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Adapt = "2, 3.2"
-ChainRulesCore = "0.9.45, 0.10, 1"
+ChainRulesCore = "0.9.45, 0.10, 1-1.17"
```
darsnack, line 16:
```suggestion
ChainRulesCore = "0.9.45, 0.10, 1 - 1.17"
```

mcabbott: Oh sorry, was reading the docs while you were writing.

mcabbott: Oh I'm an idiot, that bounds the wrong package, sorry

mcabbott: OK, restricting CR fixes it. I believe the cause is https://github.com/JuliaDiff/ChainRules.jl/pull/559/files#diff-7a31ac47520a0ad235f15cead3f4c48fcb595dc7f2421ebbee8ffce23e2510f5R84 , probably repaired in https://github.com/JuliaDiff/ChainRules.jl/pull/562

mcabbott: ok, seems fixed!

mcabbott:
This probably shouldn't wait too long, as the current state breaks Flux's tests.

darsnack (approved):
Both this and https://github.com/FluxML/Flux.jl/pull/1830 are okay with me, though I think this is safer to merge quickly and get tests up again.

PR #378, opened 2022-01-21
Informative error message for softmax variants

### LINKED ISSUE #373: Add a friendly warning to `softmax`
An easy mistake is:
```julia
julia> Dense(2,3,softmax)([1,2])
ERROR: MethodError: no method matching similar(::Float32, ::Type{Float32})

julia> softmax(pi/2)
ERROR: MethodError: no method matching similar(::Float64, ::Type{Float64})
```
Since `softmax(::Number)` it's currently an error, we can add a method which prints a friendly error message when you do this. Of course mathematically it could have a meaning, but `only(softmax([pi/2]))` is not very useful.

Edit -- note that this package already has warnings for the reverse problem, of applying an activation function to a whole array:
```julia
julia> methods(gelu)
# 2 methods for generic function "gelu":
[1] gelu(x::AbstractArray, args...) in NNlib at /Users/me/.julia/packages/NNlib/tAGmA/src/activations.jl:663
[2] gelu(x) in NNlib at /Users/me/.julia/packages/NNlib/tAGmA/src/activations.jl:319

julia> gelu([1,2,3])
ERROR: Use broadcasting (`gelu.(x)`) to apply activation functions to arrays.
Stacktrace:
 [1] error(::String, ::String, ::String)
   @ Base ./error.jl:44
```
Also xref doc improvements in https://github.com/FluxML/NNlib.jl/pull/372

theabhirath:
This PR fixes #373 by adding a warning saying that `softmax` expects an array and not a number.

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -146,5 +146,5 @@ end
 # Informative error message if any of the softmax variants is called with a number
 for f in (:softmax, :logsoftmax, :softmax!, :logsoftmax!, :logsumexp)
     @eval $(f)(x::Number, args...) = 
-      error("(`", $(string(f)), "(x)`) called with a number, but it expects an array.")
+      error("`", $(string(f)), "(x)` called with a number, but it expects an array. Usually this is because a layer like `Dense(3,4,softmax)` is broadcasting it like an activation function; `softmax` needs to be outside the layer.")
 end
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -142,3 +142,9 @@ function logsumexp(x::AbstractArray; dims = :)
     max_ = maximum(x; dims = dims)
     max_ .+ log.(sum(exp.(x .- max_); dims = dims))
 end
+
+# Informative error message if any of the softmax variants is called with a number
+for f in (:softmax, :logsoftmax, :softmax!, :logsoftmax!, :logsumexp)
+    @eval $(f)(x::Number, args...) = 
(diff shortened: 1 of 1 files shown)


src/softmax.jl
```diff
@@ -142,3 +142,9 @@ function logsumexp(x::AbstractArray; dims = :)
     max_ = maximum(x; dims = dims)
     max_ .+ log.(sum(exp.(x .- max_); dims = dims))
 end
+
+# Informative error message if any of the softmax variants is called with a number
+for f in (:softmax, :logsoftmax, :softmax!, :logsoftmax!, :logsumexp)
+    @eval $(f)(x::Number, args...) = 
+      error("(`", $(string(f)), "(x)`) called with a number, but it expects an array.")
```
DhairyaLGandhi, line 149: Should this be a warning instead? `softmax` is doing its job, but calling it with a number is not terribly useful. We added a documentation update saying that correct usage of `softmax` is on an array already too, specifically addressing the `Number` case.

darsnack: It won't work right now unless you call it like `softmax([x])` (see the linked issue).

mcabbott, line 149: Maybe one less set of brackets? And slightly more explanation, since the user presumably did not intend to apply this to one number, so may not know what to change. Maybe something like this:
```suggestion
      error("`", $(string(f)), "(x)` called with a number, but it expects an array. Usually this is because a layer like `Dense(3,4,softmax)` is broadcasting it like an activation function; `softmax` needs to be outside the layer.")
```

Issue #349, opened 2021-08-16
`ForwardDiff` and `Conv`: Slow fallback implementation invoked for conv!


garibarba:
```julia
using Flux
using ForwardDiff

x = rand(Float32, 5, 5, 1, 1)
layer = Conv((3,3), 1 => 1, pad=(1, 1), stride=(1, 1))
f = x -> sum(layer(x))

ForwardDiff.gradient(f, x)
```
Outputs:
```
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
│   yT = ForwardDiff.Dual{ForwardDiff.Tag{var"#5#6", Float32}, Float32, 9}
│   T1 = ForwardDiff.Dual{ForwardDiff.Tag{var"#5#6", Float32}, Float32, 9}
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/CSWJa/src/conv.jl:291
```
although the result is correct.

Does this mean there is a slowdown? Or not necessarily?

garibarba:
Related to https://github.com/SciML/DiffEqSensitivity.jl/issues/473

ToucheSir:
There is probably a slowdown, but I'm not sure you can do much better with dual numbers since the fast path requires BLAS-compatible data types.

DhairyaLGandhi:
If ForwardDiff overloads the definition of `<op>(::AbstractArray{Dual}...)` it can take advantage of blas in cases where it makes sense.

PR #383, opened 2022-02-05
Silence some warnings

mcabbott:
This avoids: 
```
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
[183](https://github.com/FluxML/Flux.jl/runs/5079634118?check_suite_focus=true#step:6:183)
│   yT = ForwardDiff.Dual{Nothing, Float32, 8}
[184](https://github.com/FluxML/Flux.jl/runs/5079634118?check_suite_focus=true#step:6:184)
│   T1 = ForwardDiff.Dual{Nothing, Float32, 8}
[185](https://github.com/FluxML/Flux.jl/runs/5079634118?check_suite_focus=true#step:6:185)
│   T2 = Float32
[186](https://github.com/FluxML/Flux.jl/runs/5079634118?check_suite_focus=true#step:6:186)
└ @ NNlib ~/.julia/packages/NNlib/JQe1Z/src/conv.jl:288
[187](https://github.com/FluxML/Flux.jl/runs/5079634118?check_suite_focus=true#step:6:187)
```
You will still get the warning if you mix Float32 & Float64, but not for dual numbers. 

Xref #349. I'm not sure that's closed, as they still take the slow path. But it's not by mistake.

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -285,7 +285,7 @@ for front_name in (:conv, :∇conv_data, :∇conv_filter,
                         y::AbstractArray{yT,N}, in1::AbstractArray{T1,N},
                         in2::AbstractArray{T2,N}, cdims::ConvDims;
                         kwargs...) where {yT, T1, T2, N}
-            if yT <: AbstractFloat  # don't print warning for ForwardDiff.Dual
+            if yT == Float64  # warn for Float32 + accidental Float64, but don't print warning for ForwardDiff.Dual
                 @warn string("Slow fallback implementation invoked for ", $(string(front_name)), "!  ",
                           "You probably don't want this; check your datatypes.") yT T1 T2 maxlog=1
             end
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -285,8 +285,10 @@ for front_name in (:conv, :∇conv_data, :∇conv_filter,
                         y::AbstractArray{yT,N}, in1::AbstractArray{T1,N},
                         in2::AbstractArray{T2,N}, cdims::ConvDims;
                         kwargs...) where {yT, T1, T2, N}
-            @warn string("Slow fallback implementation invoked for ", $(string(front_name)), "!  ",
+            if yT <: AbstractFloat  # don't print warning for ForwardDiff.Dual
+                @warn string("Slow fallback implementation invoked for ", $(string(front_name)), "!  ",
                           "You probably don't want this; check your datatypes.") yT T1 T2 maxlog=1
+            end
             $(Symbol("$(front_name)_direct!"))(y, in1, in2, cdims; kwargs...)
         end
     end


ToucheSir:
I wonder if we should have an info message instead for non-AbstractFloat outputs. WDYT?

mcabbott:
I guess my take here is that these warnings are there in case you messed up your float eltypes, and things will become slow by mistake. If you call ForwardDiff.gradient on something, then you really did want to put dual numbers in there. It will be slow, but so will e.g. matrix multiplication. It's not a mistake though, and so I think a noisy warning is unhelpful.

darsnack:
Other comment: this should check for subtypes of [`G`](https://github.com/FluxML/NNlib.jl/blob/fd1b9397401099af48625dc1f7dc78c3fb313640/src/conv.jl#L176) which is what actually invokes the fast vs slow paths.

mcabbott:
The PR's take is that the warning shouldn't be for all slow paths, only for slow paths which are clearly a mistake.

I agree there's an argument for telling people that Dual numbers are slow. But they are slow in many things, e.g. matmul, and may still be what you want to do. Nested AD is a bit of a performance minefield and you may want to time different approaches, if Dual happens to be best for you, then I don't think we should make it slower and noisy. 

I'm not sure I follow re G. What case will this warning not catch which should be caught, or the reverse?

mcabbott:
Ok. But can this come later?

This PR wants to disable some warnings, on the assumption that there are use cases for non-floats, which will take the slow path (as they will for matmul!) and if you want that, it's rude for the package to shout at you. The motivating example is that Flux's tests push Dual numbers through here. 

You could disable some more, but with a more complicated rule, but perhaps that can be a follow-up PR? I'm not sure what cases this logic is trying to catch. If I have a mix of floats, should the warning depend on whether one or all of them is individually acceptable, does this make it more likely to be a mistake? If I'm doing exotic floats like BigFloat, then perhaps no warnings is ideal? But maybe we can wait until someone wants that to argue the precise rule? This PR makes no changes there.

src/conv.jl
```diff
@@ -285,8 +285,10 @@ for front_name in (:conv, :∇conv_data, :∇conv_filter,
                         y::AbstractArray{yT,N}, in1::AbstractArray{T1,N},
                         in2::AbstractArray{T2,N}, cdims::ConvDims;
                         kwargs...) where {yT, T1, T2, N}
-            @warn string("Slow fallback implementation invoked for ", $(string(front_name)), "!  ",
+            if yT <: AbstractFloat  # don't print warning for ForwardDiff.Dual
```
darsnack, line 288:
```suggestion
            if all(T -> T <: G, (yT, T1, T2)) && !(T1 == T2 == yT)  # don't print warning for ForwardDiff.Dual
```

darsnack (approved):
FWIW, I don't feel strongly enough that I wouldn't accept this PR, but I've left a suggestion to make it more concrete.

> This PR makes no changes there.

That's is my point: NNlib already has a precise rule, this PR doesn't follow that definition. We have a for-loop based convolution and a GEMM-based one. The latter is called when the destination and source types match, and they are all subtypes of `G` which is the union of supported GEMM types (defined in `src/gemm.jl`). I would think that the only time the "You probably don't want this" warning applies is if you *almost* passed the GEMM check but had mismatched types. So, if one of your arguments is `BigFloat`, then the current version of the PR prints a warning that's misleading...with `BigFloat` there is no way you could be faster.

src/conv.jl
mcabbott: My argument against this is mostly that it seems like complexity to define a precise border out in the desert where nobody has been yet. When someone has reason to go there, they'll then need to figure out whether this line was drawn exactly there for what reason... and what's `G`, and why isn't `!(T1 == T2 == yT)` already implied by dispatch... who built that fence... these are sort-of puzzles for future readers. I don't want to presume what "almost passed" means, almost by what metric... this would be much clearer once someone runs into an example.

So my goal was to write the minimal thing, move the line not to cover the case which is actually hit by the tests (Dual) while not removing it from what I think it intends to cover (accidental Float64). 

Anyway, too many words! 

I'd be happy to test `if yT <: Float64` instead, if that sounds better? Super-simple, no puzzle. A mix of Float32 + Float64 will produce this. Pure Float64 won't ever come here. Complex numbers you're on your own. 

Accidental promotion is much less of a problem now that ProjectTo fixes it for you, too.

mcabbott:
```suggestion
            if yT == Float64  # warn for Float32 + accidental Float64, but don't print warning for ForwardDiff.Dual
```

darsnack: Yes, I see your point. Both the `AbstractFloat` or `Float64` options are fine by me!

PR #384, opened 2022-02-07
trigger op specialization in scatter

CarloLucibello:
It is easy to forget about this [performance tip](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing)

With this PR we get a >100x speedup by triggering specialization on `op`  :exploding_head: 

```julia
using Random, NNlib, BenchmarkTools

n = 100
m = 500
idx = rand(1:n, m)
src = rand(m)
dst = zeros(m)

@btime NNlib.scatter!(+, d, $src, $idx)  setup=(d=copy($dst));
#  343.351 μs (3500 allocations: 226.56 KiB)  # master
#  2.830 μs (0 allocations: 0 bytes)                  # this PR
 ```

Related to https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/124  cc @learning-chip

### DIFF
--- a/src/scatter.jl
+++ b/src/scatter.jl
@@ -70,7 +70,7 @@ julia> NNlib.scatter!(*, fill(0.5, 2, 4), [1 10; 100 1000], [3,2])
  0.5  500.0  50.0  0.5
 ```
 """
-function scatter!(op, dst::AbstractArray, src::AbstractArray, idx::AbstractArray)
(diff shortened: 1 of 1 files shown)


mcabbott:
Does adding this to any of the `∇scatter` functions help, too?

CarloLucibello:
I didn't try but I think it wouldn't, it's already specialized on `op::Union{typeof(+),typeof(-)}` etc...

PR #389, opened 2022-02-15
Fix gradient of convolution for complex values

zsoerenm:
This fixes https://github.com/FluxML/Flux.jl/issues/1876

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -100,7 +100,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                             dy::AbstractArray{yT,N}, w::AbstractArray{wT,N},
                             cdims::C; kwargs...) where {yT, wT, N, C <: ConvDims}
                 dx = similar(dy, input_size(cdims)..., channels_in(cdims), size(dy, N))
-                return $(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...)
+                return conj($(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...))
             end
         end
     end
@@ -113,7 +113,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                         cdims::ConvDims; kwargs...) where {xT, yT, N}
             dw = similar(dy, kernel_size(cdims)..., channels_in(cdims) ÷ groupcount(cdims),
                                                     channels_out(cdims))
-            return $(Symbol("∇conv_filter$(backend)!"))(dw, x, dy, cdims; kwargs...)
+            return conj($(Symbol("∇conv_filter$(backend)!"))(dw, x, dy, cdims; kwargs...))
         end
     end
 
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -100,7 +100,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                             dy::AbstractArray{yT,N}, w::AbstractArray{wT,N},
                             cdims::C; kwargs...) where {yT, wT, N, C <: ConvDims}
                 dx = similar(dy, input_size(cdims)..., channels_in(cdims), size(dy, N))
-                return conj($(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...))
+                return $(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...)
             end
         end
     end
@@ -113,7 +113,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                         cdims::ConvDims; kwargs...) where {xT, yT, N}
             dw = similar(dy, kernel_size(cdims)..., channels_in(cdims) ÷ groupcount(cdims),
                                                     channels_out(cdims))
-            return conj($(Symbol("∇conv_filter$(backend)!"))(dw, x, dy, cdims; kwargs...))
+            return $(Symbol("∇conv_filter$(backend)!"))(dw, x, dy, cdims; kwargs...)
         end
     end
 
--- a/src/gemm.jl
+++ b/src/gemm.jl
@@ -35,7 +35,7 @@ for (gemm, elt) in gemm_datatype_mappings
                                beta::$(elt), C::Ptr{$elt})
             # Convert our compile-time transpose marker to a char for BLAS
             convtrans(V::Val{false}) = 'N'
-            convtrans(V::Val{true})  = $elt <: Complex ? 'C' : 'T'
+            convtrans(V::Val{true})  = 'C'
 
             if transA == Val(false)
                 lda = M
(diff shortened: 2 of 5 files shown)


src/gemm.jl
```diff
@@ -35,7 +35,7 @@ for (gemm, elt) in gemm_datatype_mappings
                                beta::$(elt), C::Ptr{$elt})
             # Convert our compile-time transpose marker to a char for BLAS
             convtrans(V::Val{false}) = 'N'
-            convtrans(V::Val{true})  = 'T'
+            convtrans(V::Val{true})  = $elt <: Complex ? 'C' : 'T'
```
staticfloat, line 38: I believe that the underlying `{S,D,C,Z}GEMM` routine should correctly interpret `'C'`, even if it's a real-valued input.  In the case of real-valued input, `'C'` should be equivalent to `'T'`.

staticfloat (approved):
Great work, @zsoerenm!

I'm thinking a test case as simple as a `1x1` complex convolution with known gradient should be sufficient, just to check that it gives the gradient in the correct direction!

mcabbott:
Re tests, it should be possible to test directly the complex gradient of a function `C^n x C^m -> R`. I think ChainRulesTestUtils will even handle this by finite differencing.

Alternatively, a path where you take real input, make the complex arrays, call conv, sum, real is something on which you could test Zygote against ForwardDiff etc, end-to-end.

The test in https://github.com/FluxML/Flux.jl/issues/1876#issue-1138637519 looks probably right, but is quite complicated, and without reading it very carefully I can't tell whether it's adopted the same convention for complex numbers that Zygote uses, and not (say) the conjugate.

staticfloat:
Because that linked test case compares against `Flux.gradient()`, I think it should natively compare with Zygote and ensure that the result is correct?

zsoerenm:
I have made more changes to the pull request. This is not finished, yet. I wanted to reach out to you as soon as possible as I'd like to start a little debate here:
I think that for complex values the weights should be multiplied by it's conjugate with the input x. Though, it does not really make a difference, because in the end this will just conjugate the internal weights of the neural network (as far as I can see).
However, taking the conjugate is more text book like (see e.g. https://dsp.stackexchange.com/questions/55388/for-complex-values-why-use-complex-conjugate-in-convolution).
Whats more, with this modification taking the derivative with `∇conv_filter_direct!`, which is based on conv, just _worked_ out of the box.
I have only tested this for 1x1 convolution kernel with multiple input channels, yet. Note the `vec(w)' * vec(x)` which feels more natural for me than `transpose(vec(w)) * vec(x)`. Also note that `DepthwiseConv` will need adaptions. I will add these and more tests, once we have resolved this debate.
I'm quite new to neural networks. I'm not 100% sure, if everything is correct here. Especially I'm unsure about `∇conv_data`. Taking the conjugate of the result or not does not have an effect on my toy example. In both cases the neural network converges to its expected result.

src/gemm.jl
zsoerenm: I thought so, too, but buildkite was consistently failing on the GPU. Therefore, I implemented the branch above.

src/impl/conv_im2col.jl
```diff
@@ -51,7 +51,7 @@ function conv_im2col!(
         im2col!(col_slice, view(x, :, :, :, :, batch_idx), cdims)
         GC.@preserve col_slice w y begin
             col_ptr = pointer(col_slice)
-            w_ptr = pointer(w)
+            w_ptr = pointer(copy(conj(w)))
             y_ptr = pointer(y, (batch_idx - 1)*M*N + 1)
             gemm!(Val(false), Val(false), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
```
zsoerenm, line 56: This change was a little awkward for me.
All dense convolution tests are green, if I simply change the last line here:
```julia
w_ptr = pointer(w)
y_ptr = pointer(y, (batch_idx - 1)*M*N + 1)
gemm!(Val(false), Val(true), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
```
Note the `Val(true)` here. I wondered, if `Val(false)` is actually correct?
The `gradtest`s are failing with this change, though. So I left it as above.

src/conv.jl
```diff
@@ -100,7 +100,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                             dy::AbstractArray{yT,N}, w::AbstractArray{wT,N},
                             cdims::C; kwargs...) where {yT, wT, N, C <: ConvDims}
                 dx = similar(dy, input_size(cdims)..., channels_in(cdims), size(dy, N))
-                return $(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...)
+                return conj($(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...))
```
zsoerenm, line 103: I'm unsure about this change. My toy neural network example works with and without this change.

zsoerenm: Ah, my loss function is real in the end, due to `abs2`. This might be the reason, why it does not have an effect ;)

src/impl/conv_im2col.jl
ToucheSir: Interesting, looking at CI I wonder if this fixed the flaky gradtests we have when nthreads > 1!

staticfloat:
Okay, having looked through this a bit more thoroughly now with these changes, I see what you're proposing; you're saying that due to the definition of convolution, we should generally be taking the conjugate of the weights before convolving.

In a similar vein to how many neural network libraries treat correlation and convolution as the same (e.g. not flipping the kernel in time), I am also unsure if we actually need to do the convolution here.  It's pleasing, on an academic and technical level, since we now get the "textbook correct" results, and it certainly does make it easier to deal with non-ML fields (e.g. if the result of your ML training should correspond to some physical value).  However, it is (slightly) more work in some cases, as can be seen by how you have to `conj(x)` every now and then, which does eat up precious memory bandwidth and FLOPS.

I'm happy to let the NNlib maintainers weigh in on whether they think this is important enough to include, or if we should stick with only the first commit (along with the tests) and be happy to just state somewhere in the documentation that NNlib learns the conjugate of what convolution "should" learn.

src/impl/conv_im2col.jl
zsoerenm: No, what I meant to say is that with
```julia
gemm!(Val(false), Val(true), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
```
all gradtests are failing.

ToucheSir: Yes, sorry. I wasn't referring to the change you wrote, but to the diff here with `copy`.

zsoerenm:
Yes, to be consistent, the weights of the `Dense` layer should also be conjugated before multiplied with the input.
I guess that this change is not worth the trouble?

zsoerenm:
I have split the pull request.
This pull request simply fixes the convolution for complex values.
The other pull request https://github.com/FluxML/NNlib.jl/pull/390 addresses the question whether the weights should conjugated before it is multiplied with the input.

staticfloat (approved):
Great, I think this change is pretty non-controversial, and the tests look good to me.  Thanks so much @zsoerenm!

foldfelis:
Hi, is this PR also solve [this issue](https://github.com/FluxML/Flux.jl/issues/1655)?

ToucheSir:
It does not. Supporting GPU will require quite a bit more effort.

PR #390, opened 2022-02-17
Conjugate weight before multiplying it with input

zsoerenm:
This is a follow up of https://github.com/FluxML/NNlib.jl/pull/389
The question is whether the weights should be conjugated before they are multiplied by the input.
This is more text-book like (see e.g. https://dsp.stackexchange.com/questions/55388/for-complex-values-why-use-complex-conjugate-in-convolution).

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -100,7 +100,7 @@ for backend in (Symbol(), :_direct, :_im2col, :_nnpack)
                             dy::AbstractArray{yT,N}, w::AbstractArray{wT,N},
                             cdims::C; kwargs...) where {yT, wT, N, C <: ConvDims}
                 dx = similar(dy, input_size(cdims)..., channels_in(cdims), size(dy, N))
-                return $(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...)
+                return conj($(Symbol("$(name)$(backend)!"))(dx, dy, w, cdims; kwargs...))
             end
         end
     end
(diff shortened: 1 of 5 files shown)


staticfloat:
So if I'm understanding correctly, this would conjugate before convolution, which is more textbook correct and would make it slightly easier when comparing to non-Flux SciML workloads that deal with complex numbers; if we were to make this change however, I think we would need to make an effort to be consistent in our other workloads as well.  I'm not entirely sure if we need to, for example, calculate the forward pass of `Dense` as `conj(W) * x` to be consistent here; it seems to me that if the `Dense` layer is simply modeling matrix multiplication, there is no expectation of a conjugation upon `W`, and so it's fine to simply learn the weights that minimize the loss for a plain `W * x`.

On the other hand, I think it's better if we strive to match the existing mathematical norms as much as possible when it doesn't affect speed.

zsoerenm:
Uh, I think this pull request needs more discussions before being merged. If we go this way, I think we should also change the Dense layer to `conj(W) * x` to be consistent. Shall I go ahead and create a pull request for Flux and write some tests?

zsoerenm:
I wouldn't call this a breaking change. For real-valued numbers nothing has changed. It only affects convolutions with complex-valued numbers and those did not produce correct results before the commit https://github.com/FluxML/NNlib.jl/pull/389 was merged.

mcabbott:
One vote to revert for now.

I'm not so sure what's gained by following one convention not the other. For Flux at least, the weight is initially is arbitrary, so it shouldn't matter. Is there a non-Flux use which wants this?

Conjugating at every step costs a copy. The CPU `conv` already copies quite a few times, so maybe this is lost in the noise, but seems a bit wasteful. Maybe it could be pushed down to the adjoint BLAS routines everywhere instead?

Dense `conj(W) * x` likewise costs a copy. The layer could be changed to store the transpose, and do `W' * x`. But I still wonder what we gain?

ToucheSir:
That's odd, I'm pretty sure I didn't touch this PR at all and only merged #389? You can verify that none of the changes specific to this PR are in master. Perhaps it's because #389 is exactly 2 commits ahead of this one?

mcabbott:
Oh right, the changes here do not seem to be on master. So github is just confused, maybe because the first 4 commits here are the same as those squashed in merging https://github.com/FluxML/NNlib.jl/pull/389 .

PR #393, opened 2022-02-21
Simplify `softmax`, test second derivatives

mcabbott:
This should make softmax twice-differentiable. The code in `∇softmax` is simple enough not to bother writing a gradient by hand, all that's needed is not to take the mutating path. (Wasn't there an issue about this?)

While I'm less sure this is a good idea, it also removes `∇softmax!` entirely. That should mean that for CuArrays, it will avoid the slow path described here: https://github.com/FluxML/NNlibCUDA.jl/issues/30

Third, it also moves the old `∇softmax(Δ, x)` deprecations.jl, and adds a warning. I see this is still in use by Tracker, https://github.com/FluxML/Tracker.jl/blob/master/src/lib/array.jl#L474, but that can be fixed.

Finally, it adds a note about use with Flux, thus closes #372.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,1 +1,33 @@
-### v0.8 Deprecations
+### Deprecated while v0.7 was latest
+
+# Old 2-arg version recomputing forward
+function ∇softmax(Δ, x; dims = 1)
+    Base.depwarn("`∇softmax(Δ, x)` without `y = softmax(x)` argument is deprecatcated, as this is inefficient", :∇softmax)
+    ∇softmax(Δ, x, softmax(x, dims); dims)
+end
+∇softmax!(Δ, x; dims = 1) = Δ .= ∇softmax(Δ, x; dims)
+∇softmax!(out, Δ, x; dims = 1) = out .= ∇softmax(Δ, x; dims)
+
+# Old 2-arg version recomputing forward
+function ∇logsoftmax(Δ, x; dims = 1)
+    Base.depwarn("`∇logsoftmax(Δ, x)` without `y = logsoftmax(x)` argument is deprecatcated", :∇logsoftmax)
+    ∇logsoftmax(Δ, x, logsoftmax(x; dims); dims)
+end
+∇logsoftmax!(Δ, x; dims = 1) = Δ .= ∇logsoftmax(Δ, x; dims)
+∇logsoftmax!(out, Δ, x; dims = 1) = out .= ∇logsoftmax(Δ, x; dims)
+
+
+### Deprecated while v0.8 was latest
+
+function ∇softmax!(out::AbstractArray, Δ::AbstractArray, 
+                    x::AbstractArray, y::AbstractArray; dims = 1)
+    Base.depwarn("`∇softmax!` is deprecatcated", :∇softmax!)
+    out .= Δ .* y
+    out .= out .- y .* sum(out; dims = dims)
+end
+
+function ∇logsoftmax!(out::AbstractArray, Δ::AbstractArray,
+                    x::AbstractArray, y::AbstractArray; dims = 1) 
+    Base.depwarn("`∇softmax!` is deprecatcated", :∇softmax!)
+    out .= Δ .- sum(Δ, dims = dims) .* exp.(y)
+end
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -3,7 +3,7 @@
 
 function ∇softmax(Δ, x; dims = 1)
     # This 2-arg version recomputes the forward pass, which is slow.
-    # Removed in 0.7, but only prints a warning during 0.8:
+    # Removed from use in 0.7, but only prints a warning during 0.8:
     Base.depwarn("`∇softmax(Δ, x)` without `y = softmax(x)` argument is deprecated, as this is inefficient, please use `∇softmax_data(dy, y)`", :∇softmax)
     ∇softmax(Δ, x, softmax(x; dims); dims)
 end
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -2,7 +2,7 @@
 
 # Old 2-arg version recomputing forward
 function ∇softmax(Δ, x; dims = 1)
-    Base.depwarn("`∇softmax(Δ, x)` without `y = softmax(x)` argument is deprecatcated, as this is inefficient", :∇softmax)
+    Base.depwarn("`∇softmax(Δ, x)` without `y = softmax(x)` argument is deprecated, as this is inefficient", :∇softmax)
     ∇softmax(Δ, x, softmax(x; dims); dims)
 end
 ∇softmax!(Δ, x; dims = 1) = Δ .= ∇softmax(Δ, x; dims)
@@ -10,7 +10,7 @@ end
 
 # Old 2-arg version recomputing forward
 function ∇logsoftmax(Δ, x; dims = 1)
-    Base.depwarn("`∇logsoftmax(Δ, x)` without `y = logsoftmax(x)` argument is deprecatcated", :∇logsoftmax)
+    Base.depwarn("`∇logsoftmax(Δ, x)` without `y = logsoftmax(x)` argument is deprecated", :∇logsoftmax)
     ∇logsoftmax(Δ, x, logsoftmax(x; dims); dims)
 end
 ∇logsoftmax!(Δ, x; dims = 1) = Δ .= ∇logsoftmax(Δ, x; dims)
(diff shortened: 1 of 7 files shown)


src/softmax.jl
```diff
@@ -55,33 +73,32 @@ function softmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
...
-∇softmax(Δ, x, y; dims = 1) = ∇softmax(unthunk(Δ), x, y, dims = dims)
-
-# Can introduce at the end of deprecation cycle of ∇softmax!(out, Δ, x; dims = 1)
-# ∇softmax!(Δ, x, y; dims = 1) = ∇softmax!(Δ, Δ, x, y; dims = dims)
-
-function ∇softmax!(out::AbstractArray, Δ::AbstractArray, 
-                    x::AbstractArray, y::AbstractArray; dims = 1)
-    out .= Δ .* y
-    out .= out .- y .* sum(out; dims = dims)
+function ∇softmax(dy::AbstractArray{T}, x::AbstractArray, y::AbstractArray{S}; dims = 1) where {T,S}
```
mcabbott, line 79: The only reason this accepts `x` is to disambiguate from previous signature `∇softmax(Δ, x; dims = 1)`. Maybe we should allow you to pass anything (or `nothing`) for `x`, so as not to close over it.
```suggestion
function ∇softmax(dy::AbstractArray{T}, x, y::AbstractArray{S}; dims = 1) where {T,S}
```
Or maybe we should just give this a new name & a sensible signature?

```suggestion
function softmax∇x(dy::AbstractArray{T}, y::AbstractArray{S}; dims = 1) where {T,S}
```

CarloLucibello: I agree with the change but since we export `∇softmax` that would be breaking.

CarloLucibello: ok I saw you put a deprecation in place, fine

src/deprecations.jl
```diff
@@ -1 +1,47 @@
-### v0.8 Deprecations
+### Deprecated while v0.7 was latest
```
CarloLucibello, line 1: aren't you deprecating these now in 0.8 lifecycle?

src/softmax.jl
```diff
@@ -55,33 +73,33 @@ function softmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
     else
         @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 1, 0), exp(x - max_))
     end
-    out ./= sum(out; dims = dims)  # could re-use max_ when dims != (:) and eltype(x) == T.
+    y = out ./= sum(out; dims = dims)
```
CarloLucibello, line 76:
```suggestion
    out ./= sum(out; dims = dims)
    return out
```

src/softmax.jl
CarloLucibello: maybe we should remove the exports of gradients at some point since we define the rrules. I don't know if there are any consumers of those functions (e.g. Nabla.jl, ...).

src/deprecations.jl
mcabbott: They did not gain messages nor move to this file, but were no longer the preferred way once the versions re-using `y` were added. I guess that's all I want to capture. Nothing can be removed until 0.9.

src/softmax.jl
mcabbott: Oh I didn't notice it was exported.

I think we should probably chase down most users before merging. Tracker still uses the 2-arg ∇softmax...

src/softmax.jl
mcabbott: My thinking here, and with `dx = if` just below, is essentially to comment that we are calling the input x and the output y. Maybe it's too weird?

src/softmax.jl
ToucheSir: If the question is about naming schemes, I like what we have now with `∇conv_filter` and `∇conv_data`. That would be a tad more verbose than `softmax∇x`, but it would save some guesswork on what the positioning of each nabla means.

mcabbott: I just thought that `∇softmax_x` seemed a bit ugly, maybe `∇softmax_data` is better... the main goal was just not to re-use the existing name.

src/softmax.jl
mcabbott: removed

src/softmax.jl
```diff
@@ -95,38 +105,29 @@ It is semantically equivalent to the following:
...
 function logsoftmax!(out::AbstractArray{T}, x::AbstractArray; dims = 1) where {T}
-    max_ = maximum(x; dims = dims)
+    max_ = maximum(x; dims)
     if all(isfinite, max_)
         out .= x .- max_
     else
         @. out = ifelse(isequal(max_,Inf), ifelse(isequal(x,Inf), 0, -Inf), x - max_)
     end
-    log_ = log.(sum(exp, out; dims = dims))
+    @fastmath log_ = log.(sum(exp, out; dims))
```
mcabbott, line 119: This `@fastmath` is worth about a factor of 2:
```
julia> @btime softmax(x)   setup=(x=rand(Float32, 10, 1000));
  min 102.041 μs, mean 110.310 μs (6 allocations, 47.30 KiB)
  min 60.375 μs, mean 69.655 μs (6 allocations, 47.30 KiB)

julia> @btime logsoftmax(x)   setup=(x=rand(Float32, 10, 1000));
  min 126.166 μs, mean 132.883 μs (7 allocations, 51.36 KiB)
  min 67.875 μs, mean 74.167 μs (7 allocations, 51.36 KiB)
```
Maybe it's worth it? Is there an obvious accuracy check we should try out?

It shouldn't change `NaN` behaviour, since `maximum` will see any, and propagate them.

src/softmax.jl
```diff
@@ -95,52 +105,52 @@ It is semantically equivalent to the following:
...
+end
+
+function rrule(::typeof(logsumexp), x; dims = :)
+    # The gradient is `softmax`, but both compute `tmp` so it's worth saving.
+    max_ = maximum(x; dims)
+    @fastmath tmp = exp.(x .- max_)
+    @fastmath y = max_ .+ log.(sum(tmp; dims))
+    logsumexp_pullback(dy) = (NoTangent(), unthunk(dy) .* tmp ./ sum(tmp; dims))
+    return y, logsumexp_pullback
 end
```
mcabbott, line 154: There was no rule for logsumexp. Should we write one? 
```
julia> @btime logsumexp(x; dims=1)   setup=(x=rand(Float32, 10, 1000));
  min 62.500 μs, mean 67.966 μs (7 allocations, 51.36 KiB)  # before
  min 37.166 μs, mean 43.271 μs (7 allocations, 51.36 KiB)  # after, effect of @fastmath

julia> @btime Zygote.gradient(logsumexp, $(rand(100)));
  min 2.250 μs, mean 2.620 μs (27 allocations, 5.64 KiB)  # before
  min 6.417 μs, mean 6.973 μs (63 allocations, 7.58 KiB)  # just @fastmath, no rrule, strange?
  min 1.017 μs, mean 1.133 μs (7 allocations, 1.84 KiB)   # after
```

src/softmax.jl
ToucheSir: I had a look through the historical issues around `softmax` and couldn't find anything not already covered by the test suite.

ToucheSir: Actually, the integration test errors in https://github.com/FluxML/Tracker.jl/blob/master/test/tracker.jl#L35 and https://github.com/FluxML/Tracker.jl/blob/master/test/tracker.jl#L38 look to be new.

mcabbott: I thought those were failing on an earlier commit, without fastmath. 

Now I made https://github.com/FluxML/Tracker.jl/pull/127, which locally passes with or without this PR. We should merge that, then check that the downstream tests here pass with it, I think.

mcabbott:
I think this should be ready to go.

GPU tests are these: https://github.com/FluxML/NNlibCUDA.jl/blob/master/test/softmax.jl. They print warnings. If we had https://github.com/FluxML/NNlib.jl/issues/300 then they could be updated here in one PR...

I did however turn on Julia 1 buildkite tests (previously only 1.6), and include 1.6 on GHA (previously 1 and nightly).

ToucheSir (approved):
Feel free to merge at your leisure once the remaining stable CI passes.

mcabbott:
Quite a few CI failures, sadly:

Nightly: "ERROR: LoadError: UndefVarError: libblas_name not defined" despite #396, still waiting for binaries I guess.

GPU 1.6 (and GHA 1.7 2 threads): Convolutions "AutoDiff: spatial_rank=3   |   17     1       1     19"

ToucheSir:
Those are expected and I believe all are fixed on master :)

Issue #234, opened 2020-09-06
`Conv` is 2x slow than pytorch `Conv` on cpu


yiyuezhuo:
As @ToucheSir suggested, the Discourse [post](https://discourse.julialang.org/t/conv-is-2x-slow-than-pytorch-conv-on-cpu/39802/8) is opened here as well. Also I have used `MKL.jl` as the suggestion.

Example:
```julia
using LinearAlgebra: BLAS
BLAS.vendor()  # :mkl

using BenchmarkTools
using Flux

conv = Conv((7,7), 3 => 64; stride=2, pad=3)
dummy = randn(Float32, 224, 224, 3, 2);
conv(dummy)  # ignore compile time
@benchmark conv(dummy) 

#=
BenchmarkTools.Trial: 
  memory estimate:  19.29 MiB
  allocs estimate:  45
  --------------
  minimum time:     5.384 ms (0.00% GC)
  median time:      5.825 ms (0.00% GC)
  mean time:        6.245 ms (5.38% GC)
  maximum time:     7.804 ms (17.60% GC)
  --------------
  samples:          800
  evals/sample:     1
=#

cor = Flux.CrossCor((7,7), 3 => 64; stride=2, pad=3)
dummy = randn(Float32, 224, 224, 3, 2);
cor(dummy)  # ignore compile time
@benchmark cor(dummy) 
#=
BenchmarkTools.Trial: 
  memory estimate:  19.29 MiB
  allocs estimate:  51
  --------------
  minimum time:     5.518 ms (0.00% GC)
  median time:      5.806 ms (0.00% GC)
  mean time:        6.042 ms (4.94% GC)
  maximum time:     8.333 ms (13.14% GC)
  --------------
  samples:          827
  evals/sample:     1
=#

@benchmark conv($dummy) 
#=
BenchmarkTools.Trial: 
  memory estimate:  19.29 MiB
  allocs estimate:  45
  --------------
  minimum time:     5.503 ms (0.00% GC)
  median time:      5.786 ms (0.00% GC)
  mean time:        6.026 ms (4.99% GC)
  maximum time:     8.327 ms (12.01% GC)
  --------------
  samples:          830
  evals/sample:     1
=#
```

It seems that `@benchmark` with `PyCall` will cause memory leak, so here're `@time`s with `PyCall` multiple times manually, but it's still clear that the 2x gap exists.

```julia
using PyCall
const torch = pyimport_conda("torch", "torch")
[...]

PR #395, opened 2022-02-22
Disable BLAS threading within `conv_im2col!` etc.

mcabbott:
Looking at https://github.com/FluxML/NNlib.jl/issues/234, the implementation of `im2col!` seems to have 3 nested multi-threading operations: a `@spawn`, a `@threads` loop over the batch dim, and then BLAS threads. That might not be optimal. 

This PR finds some 30% speedups by keeping just the `@threads` loop. But could use more testing, etc. Especially for someone to try on a newer Intel machine with MKL.

Before:
```julia
julia> using Flux, BenchmarkTools, LinearAlgebra

julia> x = randn(Float32, 224, 224, 3, 32);  # more batch dim than https://github.com/FluxML/NNlib.jl/issues/234

julia> @btime copy($x);
  min 2.411 ms, mean 2.712 ms (2 allocations, 18.38 MiB)  # intel E5-2603 first line, 1.7 + OpenBLAS
  min 1.389 ms, mean 1.962 ms (2 allocations, 18.38 MiB)  # M1 mac second line always, 1.8 + Accelerate

julia> c = Conv((7,7), 3 => 64; stride=2, pad=3);

julia> cdims = Flux.DenseConvDims(x, c.weight; stride = c.stride, padding = c.pad, dilation = c.dilation, groups = c.groups);

julia> Threads.nthreads()
4

julia> BLAS.set_num_threads(4)

julia> @btime Flux.conv($x, $c.weight, $cdims);
  min 142.908 ms, mean 208.501 ms (32 allocations, 126.14 MiB)
  min 64.856 ms, mean 107.744 ms (66 allocations, 126.14 MiB)

julia> BLAS.set_num_threads(1)

julia> @btime Flux.conv($x, $c.weight, $cdims);
  min 120.732 ms, mean 350.911 ms (32 allocations, 126.14 MiB)  # faster min, slower mean, just a bad run?
  min 52.465 ms, mean 93.704 ms (67 allocations, 126.14 MiB)    # faster

# With the example from https://github.com/FluxML/NNlib.jl/issues/234 exactly:

julia> BLAS.set_num_threads(4)

julia> @btime conv($dummy);
  min 10.225 ms, mean 14.212 ms (37 allocations, 40.39 MiB)  # xeon
  min 5.835 ms, mean 8.560 ms (82 allocations, 40.39 MiB)    # M1

julia> BLAS.set_num_threads(1)

julia> @btime conv($dummy);
  min 13.030 ms, mean 15.990 ms (37 allocations, 40.39 MiB)
  min 7.011 ms, mean 8.411 ms (82 allocations, 40.39 MiB)

julia> size(dummy)  # batch size 2, hence only 2 threads not 4 used by @threads loop
(224, 224, 3, 2)
```
After:
```julia
julia> @btime Flux.conv($x, $c.weight, $cdims);
  min 112.312 ms, mean 140.835 ms (32 allocations, 126.14 MiB)  # was min 142.908 ms, mean 208.501 ms
  min 52.618 ms, mean 79.594 ms (32 allocations, 126.14 MiB)    # was min 64.856 ms, mean 107.744 ms

julia> @btime conv($dummy);
  min 13.061 ms, mean 16.524 ms (37 allocations, 40.39 MiB)  # slower, when batch size < nthreads. Perhaps test that?
  min 7.397 ms, mean 9.666 ms (37 allocations, 40.39 MiB)
```

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -196,11 +196,22 @@ for (front_name, backend) in (
                                  C_in = channels_in(cdims) ÷ groupcount(cdims),
                                  C_out = channels_out(cdims) ÷ groupcount(cdims))
 
-            Threads.@sync for (xc, wc) in zip(x_cs, w_cs)
+            if nthreads() > 1
+                th = BLAS.get_num_threads()
+                BLAS.set_num_threads(1)
+                # conv_im2col! has a loop with @threads, and benchmarks show that this is usually
+                # faster without BLAS multithreading, and without @spawn in the zip(x_cs, w_cs) loop.
+            end
+
+            for (xc, wc) in zip(x_cs, w_cs)
                 x = @view in1[ntuple(i -> i == 4 ? xc : Colon(), 5)...]
                 w = @view in2[ntuple(i -> i == 5 ? wc : Colon(), 5)...]
                 y = @view out[ntuple(i -> i == 4 ? wc : Colon(), 5)...]
-                Threads.@spawn $(Symbol("$(front_name)_$(backend)!"))(y, x, w, cdims2; kwargs...)
+                $(Symbol("$(front_name)_$(backend)!"))(y, x, w, cdims2; kwargs...)
+            end
+
+            if nthreads() > 1
+                BLAS.set_num_threads(th)
             end
 
             return out


darsnack:
Shouldn't this also test a convolution with groups > 1? That's what the spawn was for.

mcabbott:
Maybe, got an example?

In this case the `@spawn` is pure loss of course. Ideally I think we want exactly one multi-threading loop, but where to put it (without too much complexity) seems messy. Maybe making the batch loop use `@spawn` might make them compose better.

CarloLucibello:
Shall we go on with this? It could test also the grouped conv in https://github.com/FluxML/Flux.jl/pull/1921#issuecomment-1086576955

I can benchmark on a amd threadripper platform

mcabbott:
Benchmarks would be great. I stopped here because I thought we ought to check a whole range -- what if there is one group, or fewer channels than threads, etc... just timing one thing could lead you anywhere. 

Ideally we'd have multi-threading only on one outermost loop (and with some heuristic to decide when the problem is big enough). If one loop can't cover all cases then IIRC `@spawn` within `@spawn` composes better than `@threads`. But sorting all of this out started to seem like a big project.

ToucheSir:
@CarloLucibello thoughts on expanding https://github.com/FluxML/Flux.jl/blob/master/perf/conv.jl to use as a benchmarking suite?

PR #410, opened 2022-05-29
boil down upsampling code via dynamic dispatch

maxfreu:
This PR gets rid of some duplicate code which can be avoided simply by using dispatch on the different array dimensions. I think some sort of this was in discussion with @mcabbott when the initial code was written; I hope this improves things. Apart from that, I moved the `@inbounds` macro back into the threads scope, because it does not penetrate the `@threads` macro. The changes in the diff look bigger than they actually are, because I shifted some code up and down.

There is a corresponding PR in https://github.com/FluxML/NNlibCUDA.jl/pull/49 which can be merged synchronously. This is another case, where a subpackage would be nice, also for more stable CI.

### DIFF
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -105,27 +105,37 @@ can be directly specified with a keyword argument.
 The size of the output is equal to
 `(scale*S1, S2, S3)`, where `S1, S2, S3 = size(x)`.
 """
-function upsample_linear(x::AbstractArray{<:Any,3}, scale::Real)
-    outsize = floor(Int, scale * Base.size(x)[1])
+# the user facing function
+function upsample_linear(x::AbstractArray{<:Any,N}, scale::NTuple{M,Real}) where {N,M}
+    M == N-2 || error("The scale argument should be an NTuple with length $(N-2), but it has length $M.")
+    outsize = ntuple(i -> floor(Int, scale[i] * Base.size(x, i)), N-2)
     return upsample_linear(x; size=outsize)
 end
 
-function upsample_linear(x::AbstractArray{T,3}; size::Integer) where T
-    w,c,n = Base.size(x)
-    if w == size
+# convenience for single-number scale
+upsample_linear(x::AbstractArray{<:Any,N}, scale::Real) where N = upsample_linear(x, ntuple(_ -> scale, N-2))
+
+# this actually calls the upsamling kernel
+function upsample_linear(x::AbstractArray{T,N}; size::Union{Integer, NTuple{<:Any,Integer}}) where {T,N}
+    length(size) == N-2 || error("The scale argument should be an NTuple with length $(N-2), but it has length $(length(size)).")
+
+    if Base.size(x)[1:N-2] == size
         return x
     end
-    y = similar(x, T, size, c, n)
-    return upsample_linear_wcn!(y, x)
+
+    y = similar(x, T, size..., Base.size(x)[end-1:end]...)
+    return upsample_linear_kernel!(y, x)
 end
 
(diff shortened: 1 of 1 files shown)


maxfreu:
Hmm, keeping the functions around would have probably worked by forwarding to `upsample_linear_kernel`. However, the internal changes are reflected in the corresponding PR in NNlibCUDA & user-facing functions were not touched. So all the wcn whcn stuff is gone and instead relies on dispatch. The downside is that the function names are all the same now, the upside is that a lot duplicate of code was deleted. 

@CarloLucibello Sorry for the mess; I feel guilty because I didn't point out enough that this would break interoperability with NNlibCUDA by saying "can be merged synchronously" instead of "must". I dont really know how to handle integration tests in such cases. Merge, don't tag a new version, test against the specific commit hash?

CarloLucibello:
Not your fault, we should just think about how to stop relying on internals

maxfreu:
> Perviously, `upsample_linear_wcn!` only accepted 3-arrays, and `upsample_bilinear_whcn!` only accepted 4-arrays. But am I correct in thinking that `upsample_linear_kernel!` will always return the same answer as the old functions did?

Yes, it will return the same result. I hope I can work on this in the evening and add the deprecation macro + the forward. Ideally I should be able to test agains my local NNlibCUDA versions with and without the corresponding patch.

As far as internals are concerned, I think there is no way around this in this case, is it? I mean the user facing functions are declared in NNlib, together with the CPU code. The GPU side defines the kernels and the code to launch them. If I don't rely on the NNlib internals, I would have to duplicate the user facing stuff on the NNlibCUDA side, which makes no sense. Or am I wrong?

mcabbott:
I guess that NNlibCUDA overloading `f!` even when only `f` is documented does in a sense make `f!` a public function. Even thought the packages are friends, the registry sees them as separate, and thus deleting `f!` is a breaking change. 

But changing the preferred name of `f!` with a deprecation is completely fine, non-breaking. I think that's all that this PR has done. (I mean in terms of the interface.) 

>  add the deprecation macro + the forward

The macro should be all that's needed, I think. 

> test agains my local NNlibCUDA versions with and without the corresponding patch.

But also, test the registered NNlibCUDA against the new NNlib. That's the case which the registry allows. 

>  I think it is even possible to delete tags

I think the tag is already deleted (where it matters) here: https://github.com/JuliaRegistries/General/pull/61389

maxfreu:
Actually, the deprecation notice doesn't help. In the case NNlib 0.8.6 (new), NNlibCUDA 0.2.3 (old) the old functions, e.g. `upsample_linear_wcn!`, are never called, neither on the CPU, nor on the GPU side. What happens instead, is that the CuArrays are sent down the same codepath as the normal arrays and hit the CPU kernel, which results in a scalar indexing error. The functions overloaded in NNlibCUDA are simply never called in NNlib anymore. Therefore, and thinking about it, it becomes clearer to me, that older versions of NNlibCUDA are not compatible with NNlib versions >= 0.8.6, which perfectly fits the definition of a breaking change.

mcabbott:
Oh right, I see, the macro isn't going to work. It defines a ! function with the old name, but adding methods to this is useless, as only the new ! function gets called by `gather`.

I guess you could add more hops to dispatch, to make sure that overloading either `upsample_linear_kernel!` or `upsample_linear_wcn!` will work. Then NNlibCUDA can move to overload the new function, and the old one can be removed here at the next breaking change. Might be a bit ugly. 

Otherwise adding an upper bound in the registry would still work. 

Releasing 0.9 just for this seems like overkill. Nothing downstream actually needs to care, and https://juliahub.com/ui/Packages/NNlib/A7zdE/0.8.6?page=2 says that 25 different people are going to have to update their bounds.

PR #414, opened 2022-06-01
partly revert changes to stay compatible w NNlibCUDA

maxfreu:
Hi! Sorry once again for the mess I caused. This PR here cleans it up a bit and implements the suggestion from @mcabbott https://github.com/FluxML/NNlib.jl/pull/410#issuecomment-1142419044 to add another step of dispatch. Most of the code reductions still apply. The newly added block of forwarding can be removed at the next breaking change and the functions be renamed. NNlibCUDA can freely overload either e.g. `upsample_linear_kernel` or `upsample_linear_wcn`; the corresponding PR does exactly that and switches from the latter to the former. I tested both, NNlibCUDA v0.2.3 and the branch against this PR and both works.

### DIFF
--- a/src/upsample.jl
+++ b/src/upsample.jl
@@ -134,8 +134,17 @@ function upsample_linear(x::AbstractArray{T,<:Any}; size) where T<:Integer
     return round.(T, res)
 end
 
+# compatibility layer for old versions of NNlibCUDA
+# old versions overload upsample_linear_wcn, new versions overload upsample_linear_kernel
+upsample_linear_kernel!(y::AbstractArray{<:Any,3}, x::AbstractArray{<:Any,3}) = upsample_linear_wcn!(y,x)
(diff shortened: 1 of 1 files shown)


src/upsample.jl
```diff
@@ -134,8 +134,17 @@ function upsample_linear(x::AbstractArray{T,<:Any}; size) where T<:Integer
     return round.(T, res)
 end
 
+# compatibility layer for old versions of NNlibCUDA
+# old versions overload upsample_linear_wcn, new versions overload upsample_linear_kernel
```
mcabbott, line 138: Maybe say something like `can be removed from NNlib 0.9, i.e. revert https://github.com/FluxML/NNlib.jl/pull/414` in the comment?

mcabbott (approved):
This looks fine, not as messy as I feared. Thanks!

maxfreu:
This adds to the list of PRs that would have been easier with https://github.com/FluxML/NNlib.jl/issues/300, i.e. a subpackage. Then it would be just a minor version bump.

PR #423, opened 2022-06-19
Define activation functions taking arrays as input

### LINKED ISSUE #422: Activation functions have to be broadcasted by the user to act on arrays
Right now, NNlib activations act on scalars, meaning that they have to be explicitly broadcasted by the user (`x -> relu.(x)`, for example) to act on arrays. This shows up as an anonymous function inside of Flux models, often making error messages cryptic and difficult to track down. One obvious solution is to simply add these methods in NNlib (I'm happy to volunteer), but I'm not sure if there's a reason this isn't done already?

theabhirath:
An attempt to fix #422...hopefully this doesn't break anything

### DIFF
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -731,10 +731,9 @@ function softshrink(x, λ = 0.5)
     ifelse(hi > 0, ifelse(lo < 0, zero(hi), lo), hi)
 end
 
-# Provide an informative error message if activation functions are called with an array
+# Define broadcasts for activation functions on arrays
 for f in ACTIVATIONS
-    @eval $(f)(x::AbstractArray, args...) =
-      error("Use broadcasting (`", $(string(f)), ".(x)`) to apply activation functions to arrays.")
+  @eval $(f)(x::AbstractArray, args...) = $(f).(x, args...)
 end
 
 ## Faster, less accurate, versions of some.
--- a/test/activations.jl
+++ b/test/activations.jl
@@ -93,10 +93,14 @@ end
 @testset "Array input -> error" begin
     x = rand(5)
     for a in ACTIVATION_FUNCTIONS
-        @test_throws ErrorException a(x)
+        @test size(a(x)) == size(x)
+        grad = Zygote.gradient(p -> sum(a(p)), x)
+        @test size(grad[1]) == size(x)
     end
     for a in BINARY_ACTIVATIONS
-        @test_throws ErrorException a(x, 0.1)
+        @test size(a(x, 0.1)) == size(x)
+        grad = Zygote.gradient(p -> sum(a(p, 0.1)), x)
+        @test size(grad[1]) == size(x)
     end
 end
 


darsnack:
Convolution fuzzing tests are already failing on master

DhairyaLGandhi:
The canonical definition would almost always broadcast in the forwards
pass, which GPUCompiler usually catches and dispatches to the kernels with
array arguments. Can we get an idea of how we dispatch to these when most
forwards passes broadcast by default?

darsnack:
What do you mean? The canonical definition of what?

If it's the layer forward definitions, then this doesn't change those. All this does is allow `relu` directly in a `Chain` instead of `x -> relu.(x)`. And since the implementation is just broadcasting the function, it will hit the same paths.

DhairyaLGandhi:
The forward passes in many cases looks like 
```julia
act.(f(x))
```

Ie with the activation broadcasted over an object. In this case, the activation function provided has to be `relu`. When something like `x -> relu.(x)` is provided with the above form of the forwards pass, the anonymous function is already broadcasted and generates

```julia
(x -> relu.(x)).(f(x))
```

Thus what actually happens is that the anonymous function actually receives a scalar and works anyway since numbers are iterable. I think in most cases it doesn't make much difference but that is what is happening unless the compiler can optimise the extra broadcast away. In AD, we would actually have to see both the outer broadcast and the inner broadcast and generate pullback for both (this hopefully isn't the case with a lens based system that can interleave optimisation and compilation, but is the case elsewhere).

To be clear, I'm in favour of picking points of optimisation and simplifications, i just wanted to clarify that it's mostly useful for cases when the activation function sees an iterable of arrays (like a tuple or vector of array), and if there are specific advantages to automatically calling broadcasted operations (say for fusion) then perhaps overloading `broadcasted` would give the compiler more hints.

CarloLucibello (approved):

theabhirath:
Could we get a patch release with this? Would be helpful 😅

mcabbott:
I missed this, but before we release, is it a good idea? 

It means `Chain(Dense(2 => 2), relu)(rand(2, 2))` will work, instead of giving an explanatory error. But `Chain(Dense(2 => 2), tanh)(rand(2, 2))` will do something quite different, and using your own function `f` will probably not work. 

Another level at which this "do what I mean" could be implemented is to make the `Chain` constructor replace any lonely activation functions with something to broadcast them. Perhaps with a warning. Then when you enter such a chain at the REPL, at least you know that it's been fixed up somehow.

mcabbott:
> this is the assumed understanding

I agree it's what's normally meant, it just seems a bit contrary to Julia's normal behaviour. 

My `tanh` example is meant to highlight that some functions already have matrix definitions, which are quite different. Any function you write yourself out of `exp`, `log` etc. will also tend to have such a definition. Although in practice getting a square matrix input is going to be unlikely.

My `Chain` constructor idea may indeed not be a great one. If it's noisy, then it could be an earlier place to remind people "you need to broadcast that!" than waiting until a runtime error. There would be less expectation that such a "training wheels" feature also apply to `Parallel` etc.

darsnack:
> My tanh example is meant to highlight that some functions already have matrix definitions, which are quite different.

Good point, and actually these cases should probably reverted as piracy?

ToucheSir:
The nuclear option is to do what PyTorch does and make vectorized activation layer types/constructors, then not export any of the activation functions themselves so that users are incentivized to use the former. That gets around some of the confusion with `act.(x)` vs `act(x)` not always being equivalent, but historically we've not wanted to go down this route.

PR #425, opened 2022-07-20
export `grid_sample`

scheidan:
The motivation is to include the nice `grid_sample` function in the docs (otherwise it is very difficult to find). 
Maybe there is also a way to do so without exporting it.

### DIFF
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -79,6 +79,7 @@ include("gather.jl")
 include("scatter.jl")
 include("utils.jl")
 include("sampling.jl")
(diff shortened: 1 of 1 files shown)


ToucheSir:
There is indeed, but it's a little different since NNlib doesn't have a dedicated docs site. You'd have to add `grid_sample` to the list of functions on https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/nnlib.md.

scheidan:
Closed in favor of https://github.com/FluxML/Flux.jl/pull/2022

PR #426, opened 2022-07-22
Move `ctc_loss` from Flux to NNlib

mcabbott:
Copied without changes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NNlib"
 uuid = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
-version = "0.8.8"
+version = "0.8.9"
 
 [deps]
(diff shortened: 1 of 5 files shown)


ToucheSir (approved):
Tests pass so LGTM. I did notice though that we don't document the CTC loss at all currently :grimacing:. Thoughts on negotiating that if we start moving more loss functions over to NNlib?

mcabbott:
Didn't look into that... just trying to do the minimal change as step one.

I guess all losses could be moved, but most are just a few lines; this one needs CUDA which is the motivation here.

maetshju:
There is a docstring already for `ctc_loss`, but it looks like I didn't put a reference for the function in the actual docs anywhere when the PR was merged into Flux. I would be happy to revise the docs for it if desired.

Issue #432, opened 2022-08-13
Nested AD failure with `logσ` after https://github.com/JuliaDiff/ChainRules.jl/pull/644


ToucheSir:
Found while creating https://github.com/FluxML/Zygote.jl/pull/1285. Doing some printf debugging near https://github.com/FluxML/Zygote.jl/blob/be5b47fad5fc9c0a3e22f239ec8517df60ffdf4c/src/lib/broadcast.jl#L196:
```
summary(y∂b) = "5-element Vector{Tuple{Float64, Zygote.ZBack{NNlib.var\"#sigmoid_fast_pullback#142\"{Float64, ProjectTo{Float64, NamedTuple{(), Tuple{}}}}}}}"
args = (Base.Broadcast.Broadcasted(-, ([-0.8, -0.1, 0.2, 0.4, 1.3],)),)
```
This indicates https://github.com/FluxML/NNlib.jl/blob/cb30e19c63274791ace4b002844f09086a885e74/src/activations.jl#L849 is working, but for some reason the broadcast isn't being materialized before hitting the rule. Could this be something similar to https://github.com/dfdx/Yota.jl/issues/121#issuecomment-1211361086?

cc @mcabbott

mcabbott:
The CR rule for `.+` is lazier, doesn't materialize, so that things can fuse. But the mystery is why that should be called here, instead of Zygote's own rules.

ToucheSir:
I think hitting the `rrule` first for unary `-` makes sense, since the call chain goes `broadcasted(f, arg) -> broadcasted(style, f, arg)` and Zygote only has a rule for the latter in this case. What confused me at first was why the subsequent `broadcast(sigmoid_fast, ...)` _is_ hitting Zygote, but that also makes sense because https://github.com/FluxML/NNlib.jl/blob/v0.8.9/src/activations.jl#L880 only dispatches on arrays. Should we change that?

mcabbott:
Ah right, sorry I assumed there was a rule, but only binary `-` has one:
https://github.com/FluxML/Zygote.jl/blob/master/src/lib/broadcast.jl#L81-L82

So the simple solution is to define such a rule in Zygote. And perhaps audit the other fused rules to look for others with no Zygote equivalent: 
https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/broadcast.jl#L149-L231

PR #433, opened 2022-08-15
Widen activation broadcast rules

mcabbott:
In something like `relu.(x .+ b)`, the broadcast rule for `.+` in Zygote always makes an array, but the rule in ChainRules makes a `Broadcasted` to permit fusion. In that case the rule for `relu.(x::Numeric)` here won't currently be called; so this PR widens its signature. As suggested here: https://github.com/FluxML/NNlib.jl/issues/432#issuecomment-1214444303


ToucheSir (approved):
Is there any way to test the Broadcasted case? Otherwise LGTM.

mcabbott:
I added a test. I think it'll also fail with sufficiently old CR.jl, but maybe that's OK.

PR #436, opened 2022-10-13
Add basic benchmark harness

ToucheSir:
A revamp of @dfdx's work in #304. The main changes here are running benchmarks and rendering the report on Buildkite instead of GHA, because the former gives us runners with >2 threads.

### PR Checklist

- ~~Tests are added~~
- ~~Documentation, if applicable~~

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,6 +40,21 @@ steps:
       NNLIB_TEST_CUDA: true
     timeout_in_minutes: 60
 
+  - label: "Benchmarks"
+    plugins:
+      - JuliaCI/julia#v1:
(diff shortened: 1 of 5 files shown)


ToucheSir:
Unfortunately it's not possible to compare against master because no such benchmarking code exists there, but here is an example of what the output would look like: https://buildkite.com/julialang/nnlib-dot-jl/builds/501#annotation-default. Posting a GH comment would be far more difficult because of secret management and is best left for a future PR.

darsnack (approved):
This looks good to me, we can add more to the suite or solve the commenting issue separately.

Also, the documentation deployment issues should be resolved, and we can merge this to test that.

PR #438, opened 2022-10-22
Create root level index.html

Saransh-cpp:
I am not sure why this wasn't generated by `Documenter.jl` on the first run, but the [documentation](https://fluxml.ai/NNlib.jl/) is down because of this. Working demo - https://saransh-cpp.github.io/NNlib.jl/ (automatically redirects to `/dev`)

### PR Checklist

- [ ] Tests are added


ToucheSir:
I'm surprised this needs to be added manually to the gh-pages branch. Will it survive subsequent Documenter rebuilds?

Saransh-cpp:
Documenter should have added this file on the first build instead of us manually adding it. If it doesn't survive the subsequent rebuilds, we might have to report it as a bug.

ToucheSir (approved):

PR #441, opened 2022-10-25
Remove threading from all `∇*conv_filter` and re-enable old tests

ToucheSir:
This should hopefully fix persistent threading-related issues on CI, While also slightly reducing memory usage when taking gradients. The fix applied is the same as https://github.com/FluxML/NNlib.jl/pull/235. I've taken the opportunity to not over-allocate memory for convs and depthwise convs at the same time.

Fixes https://github.com/FluxML/NNlib.jl/issues/359, which has been causing CI failures for the past two years...

### DIFF
--- a/src/dim_helpers/ConvDims.jl
+++ b/src/dim_helpers/ConvDims.jl
@@ -77,6 +77,25 @@ function im2col_dims(c::ConvDims)
     )
 end
 
+"""
+    ∇filter_im2col_dims(c::ConvDims)
+
+Like [`im2col_dims`](@ref), but saves some memory because multiple (Julia) threads are
+not required for the filter gradient calculation.
+
+Note: in the future, this may return `Dims{2}` instead of `Dims{3}`.
+"""
+function ∇filter_im2col_dims(c::ConvDims)
+    return (
+        # Output size
+        prod(output_size(c)),
+        # Size of single dotproduct within convolution
+        prod(kernel_size(c))*channels_in(c),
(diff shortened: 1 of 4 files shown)


ToucheSir:
Ok, a couple of findings:
1. BLAS threads are not a culprit, since setting them == 1 still repros this easily.
2. `julia -t 2` is enough to repro this consistently. What I don't understand is why Buildkite and running locally are way more consistent than on GHA despite all 3 using 2 threads. Maybe resource contention on the latter leads to less effective parallelism?
3. At least on paper, threads should not be writing to overlapping memory regions for depthwise convs. Is it possible that there's a bug?
4. On the opposite end, could it be that the error comes from how FiniteDifferences interacts with threading? I don't understand its internals at all, but it feels to me like the Zygote version should give fully deterministic results.

Going to sleep on this and try seeing if both gradients are self-consistent across multiple runs tomorrow. If anybody has any insights on this, please do post them!

ToucheSir:
So used to seeing broken CI here that the green looks unnatural! This is good to go, barring any objections about the threading removal.

CarloLucibello:
Very nice to see all greens here!
Should we open an issue tracking where we stand with threaded convolutions (nowhere I guess at this point) and lay down some possible future plans? E.g. we might reconsider wrapping again NNPACK but in a more robust and maintainable way, explore oneDNN or others

CarloLucibello (approved):

PR #347, opened 2021-08-09
Improve a few activation functions

mcabbott:
This makes `leakyrelu` 3x quicker, and `relu` 10% quicker. I'm sure these weren't bottlenecks for anyone, but still. Also upgrades gradients to use `InplaceableThunk`.

PR #443, opened 2022-11-18
Slightly faster softplus

Sleort:
A few percent faster `softplus` using `relu(x)` instead of `max(x, 0)`. Ref. https://github.com/FluxML/NNlib.jl/pull/347.

### DIFF
--- a/src/activations.jl
+++ b/src/activations.jl
@@ -615,7 +615,7 @@ julia> softplus(16f0)
 16.0f0
 ```
 """
-softplus(x) = log1p(exp(-abs(x))) + max(x, 0)
+softplus(x) = log1p(exp(-abs(x))) + relu(x)
 
 """
     logcosh(x)


CarloLucibello:
makes sense, thanks!

ToucheSir:
It would've been nice to try the new benchmarking apparatus on this, but oh well. @Sleort if you notice any other low-hanging fruit, maybe we can try then.

mcabbott:
BTW, it is possible to go much faster here:
```julia
julia> @btime y .= softplus.(x)   setup=(x=randn(Float32,1000); y=similar(x));
  15.416 μs (0 allocations: 0 bytes)

julia> @btime y .= g8.(x)   setup=(x=randn(Float32,1000); y=similar(x));
  123.042 μs (8489 allocations: 132.64 KiB)
```
where it appears I used `exp(x)` for large negative x, then a polynomial correction, then a rational -- I don't remember but perhaps this seemed too complicated:
```
@inline f7(x) = evalpoly(x, 
    (0.6931472f0, 0.33953527f0, 0.09100373f0, 0.016568843f0, 0.0013059091f0, 1.8928275f-6)
)/evalpoly(x, 
    (1.0f0, 0.21118937f0, 0.11348734f0, 0.0145165585f0, 0.0014029884f0)
)

@inline f8(x) = evalpoly(x, 
    (0.99217963f0, 0.0027701536f0, -0.00039373388f0, 2.8039218f-5, -9.995612f-7, 1.4259708f-8)
)

@inline g8(x) = begin ax = abs(x)
    eax = @fastmath exp(-ax)
    ifelse(x < -15, eax, ifelse(x < -10, eax * f8(ax), eax * f7(ax) + relu(x)))
end
```

PR #447, opened 2022-12-24
Add `lppool` implementation

skyleaworlder:
About FluxML/Flux.jl#1431: LPPool1d and LPPool2d.

### PR Checklist

Related tests and documentation will come soon.

- [x] Tests are added

### DIFF
--- a/src/impl/pooling_direct.jl
+++ b/src/impl/pooling_direct.jl
@@ -1,14 +1,14 @@
 # Pooling is so similar, we abstract over meanpooling and maxpooling, simply replacing
 # the inner loop operation and a few initialization parameters.
-for name in (:max, :mean)
+for name in (:max, :mean, :lp)
     @eval function $((Symbol("$(name)pool_direct!")))(
                     y::AbstractArray{T, 5}, x::AbstractArray{T, 5},
-                    pdims::PoolDims; alpha::T=T(1), beta::T=T(0)) where T
+                    pdims::PoolDims; alpha::T=T(1), beta::T=T(0), kwargs...) where T
         $((Symbol("$(name)pool_direct!")))(
             y, x, pdims,
             Val(kernel_size(pdims)), Val(channels_out(pdims)),
             Val(padding(pdims)), Val(dilation(pdims)), Val(stride(pdims));
-            alpha, beta)
+            alpha, beta, kwargs...)
         return y
     end
 
@@ -17,7 +17,7 @@ for name in (:max, :mean)
         pdims::PoolDims,
         # kernel size, channels out, padding, dilation, stride
         ::Val{K}, ::Val{C}, ::Val{P}, ::Val{D}, ::Val{S};
-        alpha::T=T(1), beta::T=T(0),
+        alpha::T=T(1), beta::T=T(0), kwargs...
     ) where {T, K, C, P, D, S}
         @assert beta == T(0) "beta not supported yet"
         check_dims(size(x), size(y), pdims)
@@ -41,10 +41,15 @@ for name in (:max, :mean)
             alpha = alpha / prod(K)
         end
 
+        p = if $(name != :lp) 0 else
+            !haskey(kwargs, :p) && error("lppool must pass p")
+            kwargs[:p]
+        end
+
         # Each loop, we initialize `m` to something, set that here.
         m_init = if $(name == :max)
             T <: AbstractFloat ? nextfloat(typemin(T)) : typemin(T)
-        elseif $(name == :mean)
+        elseif $(name == :mean) || $(name == :lp)
             T(0)
         else
             error("Unimplemented codegen path")
@@ -78,10 +83,14 @@ for name in (:max, :mean)
                     end
                 elseif $(name == :mean)
                     m += x[input_kw, input_kh, input_kd, c, batch_idx]
+                elseif $(name == :lp)
+                    m += x[input_kw, input_kh, input_kd, c, batch_idx]^p
                 else
                     error("Unimplemented codegen path")
                 end
             end
+
+            m = $(name == :lp) ? m^(1 / p) : m
 
             y[w, h, d, c, batch_idx] = alpha * m # + beta * y[w, h, d, c, batch_idx]
             end
@@ -128,12 +137,15 @@ for name in (:max, :mean)
                                 end
                             elseif $(name == :mean)
                                 m += x[input_kw, input_kh, input_kd, c, batch_idx]
+                            elseif $(name == :lp)
+                                m += x[input_kw, input_kh, input_kd, c, batch_idx]^p
                             else
                                 error("Unimplemented codegen path")
                             end
                         end
                     end
                 end
+                $(name == :lp) && (m = m^(1 / p))
                 y[w, h, d, c, batch_idx] = alpha * m # + beta * y[w, h, d, c, batch_idx]
                 end
                 end
@@ -159,7 +171,7 @@ for name in (:max, :mean)
                     dx::AbstractArray{T,5}, dy::AbstractArray{T,5},
                     y::AbstractArray{T,5}, x::AbstractArray{T,5},
                     pdims::PoolDims, ::Val{K}; # == kernel_size(pdims)
-                    alpha::T=T(1), beta::T=T(0)) where {T, K}
+                    alpha::T=T(1), beta::T=T(0), kwargs...) where {T, K}
         check_dims(size(x), size(dy), pdims)
 
         width, height, depth = input_size(pdims)
@@ -178,12 +190,20 @@ for name in (:max, :mean)
 
         # If we're doing mean pooling, we represent division by kernel size by rolling
         # it into the `alpha` multiplier.
-        if $(name == :mean)
+        if $(name == :mean) || $(name == :lp)
             alpha = alpha / prod(K)
+        end
+
+        p = if $(name != :lp) 0 else
+            !haskey(kwargs, :p) && error("lppool must pass p")
+            kwargs[:p]
         end
 
         # Start with the central region
         w_region, h_region, d_region = central_region
+        display(d_region)
+        display(h_region)
+        display(w_region)
         @inbounds for batch_idx in 1:size(x, 5), c in 1:out_c
             for d in d_region
             pd = project(d, stride_d, pad_d_lo)
@@ -226,6 +246,8 @@ for name in (:max, :mean)
                 elseif $(name == :mean)
                     # Either does meanpool :(
                     dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * alpha
+                elseif $(name == :lp)
+                    dx[input_kw, input_kh, input_kd, c, batch_idx] += (dy_idx^p * alpha)^(1/p)
                 else
                     error("Unimplemented codegen path")
                 end
@@ -286,6 +308,8 @@ for name in (:max, :mean)
                                 end
                             elseif $(name == :mean)
                                 dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * alpha #+ beta * dx[x_idxs...]
+                            elseif $(name == :lp)
+                                dx[input_kw, input_kh, input_kd, c, batch_idx] += (dy_idx^p * alpha)^(1/p)
                             else
                                 error("Unimplemented codegen path")
                             end
--- a/src/impl/pooling_direct.jl
+++ b/src/impl/pooling_direct.jl
@@ -201,9 +201,6 @@ for name in (:max, :mean, :lp)
 
         # Start with the central region
         w_region, h_region, d_region = central_region
-        display(d_region)
-        display(h_region)
-        display(w_region)
         @inbounds for batch_idx in 1:size(x, 5), c in 1:out_c
             for d in d_region
             pd = project(d, stride_d, pad_d_lo)
@@ -247,7 +244,8 @@ for name in (:max, :mean, :lp)
                     # Either does meanpool :(
                     dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * alpha
                 elseif $(name == :lp)
-                    dx[input_kw, input_kh, input_kd, c, batch_idx] += (dy_idx^p * alpha)^(1/p)
+                    grad = x[input_kw, input_kh, input_kd, c, batch_idx]^(p-1) * y_idx^(1-p)
+                    dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * grad
                 else
                     error("Unimplemented codegen path")
                 end
@@ -309,7 +307,8 @@ for name in (:max, :mean, :lp)
                             elseif $(name == :mean)
                                 dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * alpha #+ beta * dx[x_idxs...]
                             elseif $(name == :lp)
-                                dx[input_kw, input_kh, input_kd, c, batch_idx] += (dy_idx^p * alpha)^(1/p)
+                                grad = x[input_kw, input_kh, input_kd, c, batch_idx]^(p-1) * y_idx^(1-p)
+                                dx[input_kw, input_kh, input_kd, c, batch_idx] += dy_idx * grad
                             else
                                 error("Unimplemented codegen path")
                             end
--- a/src/impl/pooling_direct.jl
+++ b/src/impl/pooling_direct.jl
@@ -84,14 +84,14 @@ for name in (:max, :mean, :lpnorm)
                 elseif $(name == :mean)
                     m += x[input_kw, input_kh, input_kd, c, batch_idx]
                 elseif $(name == :lpnorm)
-                    # y = (∑ x^p)^(1/p), here to calculate (∑ x^p)
+                    # y = (∑ᵢ xᵢ^p)^(1 / p), here to calculate ∑ᵢ xᵢ^p
                     m += x[input_kw, input_kh, input_kd, c, batch_idx]^p
                 else
                     error("Unimplemented codegen path")
                 end
             end
 
-            # for lpnormpool, y = (∑ x^p)^(1/p)
+            # for lpnormpool, y = (∑ᵢ xᵢ^p)^(1 / p)
             m = $(name == :lpnorm) ? m^(T(1) / p) : m
 
             y[w, h, d, c, batch_idx] = alpha * m # + beta * y[w, h, d, c, batch_idx]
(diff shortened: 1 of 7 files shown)


mcabbott (commented):
Looks good, thanks! I have not checked the implementation closely but have a few comments in the meantime...

src/pooling.jl
```diff
@@ -166,7 +172,20 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
 end
 
 
-for pool in [:maxpool, :meanpool]
+"""
+    lppool(x, p::Number, k::NTuple; pad=0, stride=k)
+
+Perform Lp pool operation with `p`-norm and `window size `k` on input tensor `x`
+"""
```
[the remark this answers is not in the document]
ToucheSir: The PyTorch layer docstring would be a nice reference for this.

src/pooling.jl
```diff
@@ -166,7 +172,20 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
+function lppool(x, p::Number, k::NTuple{N, Integer}; pad=0, stride=k) where N
+    pad = expand(Val(N), pad)
+    stride = expand(Val(N), stride)
+    pdims = PoolDims(x, k; padding=pad, stride=stride)
```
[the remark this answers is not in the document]
skyleaworlder:
```suggestion
    ndims(x) == length(k) + 2 || error("lppool expects ndims(x) == length(k)+2,
                                        dimension of x is $(ndims(x)),
                                        length of k need $(ndims(x) - 2),
                                        but now it's $(length(k))")
    pdims = PoolDims(x, k; padding=pad, stride=stride)
```

I add size checker in new commit like this.

src/pooling.jl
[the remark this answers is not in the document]
skyleaworlder: I add 

> Can this explain a bit more?
> 
> * It should say that it requires `ndims(x) == length(k)+2`.
> * I think it should also say `lppool(x, 1, k) ./ prod(k) ≈ meanpool(x, k)`, and maybe `lppool(x, 2, k).^2 ./ prod(k) ≈ meanpool(x.^2, k)`.
> * What range of `p` is allowed? E.g. someone might expect `lppool(reshape(1:10.0,:,1,1), Inf, (3,))` to be maxpool, but it isn't... maybe anything which works for `norm` but not here should be an error?
> * Can it briefly say what types are allowed for `stride, pad`, or refer to fuller docs elsewhere?

I add docs for these aspects in new commit.

> The PyTorch layer docstring would be a good inspiration for this.

I wonder if it might be suitable to write super detailed docs here. Flux.jl would wrap pooling layer after all.

src/impl/pooling_direct.jl
```diff
@@ -41,10 +41,15 @@ for name in (:max, :mean)
             alpha = alpha / prod(K)
         end
 
+        p = if $(name != :lp) 0 else
+            !haskey(kwargs, :p) && error("lppool must pass p")
```
CarloLucibello, line 45:
```suggestion
            !haskey(kwargs, :p) && error("lppool needs keyword argument `p`")
```

src/impl/pooling_direct.jl
```diff
@@ -78,11 +83,17 @@ for name in (:max, :mean)
...
+                elseif $(name == :lp)
+                    # y = (∑ x^p)^(1/p), here to calculate (∑ x^p)
+                    m += x[input_kw, input_kh, input_kd, c, batch_idx]^p
                 else
                     error("Unimplemented codegen path")
                 end
             end
 
+            # for lppool, y = (∑ x^p)^(1/p)
+            m = $(name == :lp) ? m^(1 / p) : m
```
CarloLucibello, line 95:
```suggestion
            m = $(name == :lp) ? m^(T(1) / p) : m
```
We don't want to cast Float64 here. I think we should also put some type check or conversion on `p` somewhere above.

src/impl/pooling_direct.jl
```diff
@@ -178,10 +192,15 @@ for name in (:max, :mean)
 
         # If we're doing mean pooling, we represent division by kernel size by rolling
         # it into the `alpha` multiplier.
-        if $(name == :mean)
+        if $(name == :mean) || $(name == :lp)
             alpha = alpha / prod(K)
```
CarloLucibello, line 196: If the output is supposed to be `y = (∑ x^p)^(1/p)`, we shouldn't divide by prod(K), right?

src/pooling.jl
```diff
@@ -154,19 +168,69 @@ end
...
+
+For each element `x` in (k × k) window, lppool computes `(∑ x^p)^(1 / p)` as output.
+
+* When p = 1, lppool(x, p, k) ./ prod(k) ≈ meanpool(x, k)
+* When p = 2, lppool(x, p, k).^2 ./ prod(k) ≈ meanpool(x.^2, k)
+
+!!! warning
+
+    Theoretically, when `p -> ∞`, lppool(x, p, k) ≈ maxpool(x, k).
+    But it's not correct in julia. Given a normal valuable `x`,
```
CarloLucibello, line 208: "Given a normal valuable `x`" not clear

test/pooling.jl
```diff
@@ -248,6 +248,70 @@ meanpool_answer_dict = Dict(
     )
 )
 
+lppool_answer_dict = Dict(
```
CarloLucibello, line 251: where did you take these values? pytorch? or "manual"computation?

skyleaworlder: Part from pytorch `LPPool1d` and `LPPool2d` (normal & stride), and part from calculator (pad, because pytorch `LPPool` doesn't support pad)

src/impl/pooling_direct.jl
skyleaworlder: Oh yes, I forgot to delete this. Thank you!

src/pooling.jl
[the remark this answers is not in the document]
ToucheSir: Our docstrings for `meanpool` and `maxpool` are quite short. Given that LP-pooling is far less well known however, it warrants a bit more explanation.

src/pooling.jl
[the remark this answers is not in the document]
mcabbott: Great. It could also live in PoolDims, to be applied to all of them.

src/pooling.jl
```diff
@@ -154,19 +168,69 @@ end
...
+    But it's not correct in julia. Given an arbitrary Number `n`,
+    ```jldoctest
+    julia> n = 10
+    10
+
+    julia> ans^Inf
+    Inf
+
+    julia> ans^(1/Inf)
+    1.0
```
mcabbott, line 217: This isn't the way. It's not a warning box. And "it's not correct in julia" is simply false. 

Conceptually this function is like `norm(x,p)` on a moving window. The least surprising option is that it should accept all the same `p` values which `norm` does. 

There's a mathematically sensible meaning for when p becomes infinite, etc. But the implementation of `LinearAlgebra.norm`  doesn't fall out of floating point arithmetic like this comment suggest it should. It branches on `if p == -Inf` etc. 

We could duplicate all those branches here. And test (like LinearAlgebra does) that they all work. 

It's also fine to decide that nobody needs `lppool(x, -3.3)`. In that case, the two things needed here are (1) a comment like "accepts `0 < p < Inf`", no more, and (2) an error on `p` outside the accepted range. But no lesson on IEEE floats, no giant warning box.

mcabbott:
Re the name, we have `meanpool` and `maxpool`, corresponding to mapping `mean` and `maximum` (which is `reduce(max, ...)` over moving windows. 

This one maps `norm` over a moving window. Shouldn't it be `normpool`?

I realise this breaks with pytorch, but we aren't tied to copying them. Also, `lppool` is a weird name to read, how many ppp in a row, etc.

ToucheSir:
The PyTorch implementation uses "power-average pooling", so that's another option.

One concerning thing I did find while looking around for name ideas is that nobody seems to be using this? GitHub search turns up only a couple instances of `LPPool2d` that weren't just copies of the PyTorch repo, and I couldn't identify the source paper/presentation/document this operation was first introduced in.

src/pooling.jl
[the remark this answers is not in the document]
skyleaworlder: OK, I fix it!

skyleaworlder:
> The PyTorch implementation uses "power-average pooling", so that's another option.
> 
> One concerning thing I did find while looking around for name ideas is that nobody seems to be using this? GitHub search turns up only a couple instances of `LPPool2d` that weren't just copies of the PyTorch repo, and I couldn't identify the source paper/presentation/document this operation was first introduced in.

The paper proposes `LPPool` might be [Learned-Norm Pooling for Deep Feedforward and Recurrent Neural Networks](https://arxiv.org/pdf/1311.1780.pdf). In this paper, they claim it's better to consider $L_p$ as a hyperparameter instead of a fixed value. And about another problem, I must say I also haven't used this kind of pooling method, maxpool for me mostly. Actually, I just notice Flux.jl#1431 and feel like taking this to learn about Flux.

ToucheSir:
Nice find! Do you mind adding it as the primary reference in the docstring (see how this is done for Flux layers)? Also, is `lpnormpool` too much of a mouthful, and if so which parts of that name should stay since the paper uses both $L_p$ and "norm"?

skyleaworlder:
> Nice find! Do you mind adding it as the primary reference in the docstring (see how this is done for Flux layers)? Also, is `lpnormpool` too much of a mouthful, and if so which parts of that name should stay since the paper uses both $L_p$ and "norm"?

OK, I've changed the name in recent commit and add paper reference.

mcabbott:
The title does have "Norm Pooling" as they felt squeezing L_p in there too much?

Also LinearAlgebra's function could be called `lpnorm`, but it's not, they settled on just `norm`. That seems like the obvious thing to echo.

skyleaworlder:
I think lppool is ok or "lp" is needed because pytorch uses this. If normpool is used as name, users may spend time figuring out that normpool equals lppool in pytorch.

CarloLucibello:
having the more articulate name `lpnormpool` for a rarely used operator seems fine

CarloLucibello (approved):

src/pooling.jl
```diff
@@ -166,7 +180,32 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
 
 
-for pool in [:maxpool, :meanpool]
+"""
+    lpnormpool(x, p::Number, k::NTuple{N, Integer}; pad=0, stride=k)
+
+Perform Lp pool operation with value of the Lp norm `p` and window size `k` on input tensor `x`, also known as LPPool in pytorch.
+This pooling operator from [Learned-Norm Pooling for Deep Feedforward and Recurrent Neural Networks](https://arxiv.org/abs/1311.1780).
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
```
mcabbott, line 189: `[3, 5]` means something else in Julia. And the relationship is always, not usually:
```suggestion
* `x` and `k`: Expects `ndim(x) ∈ 3:5` and always `length(k) == ndim(x) - 2`
```

src/pooling.jl
```diff
@@ -166,7 +180,32 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
+* `p` is restricted to (0, Inf). Out-of-range `p` leads to undefined behavior.
```
mcabbott, line 190: It's an error not undefined, right? And `(0, Inf)` could mean a tuple, clearer as code?
```suggestion
* `p` is restricted to `0 < p < Inf`.
```

src/pooling.jl
```diff
@@ -166,7 +180,32 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
+* `pad`: See [`pad_zeros`](@ref) for details.
+* `stride`: Stride for each spatial axis. `k` as default if not present.
+
+For each element `x` in (k × k) window, lpnormpool computes `(∑ x^p)^(1 / p)` as output.
```
mcabbott, line 194: This `x` and `k` don't match the ones above. Perhaps:
```suggestion
For all elements `xᵢ` in a size-`k` window, lpnormpool computes `(∑ᵢ xᵢ^p)^(1 / p)` as output.
```

mcabbott (changes requested):

src/pooling.jl
```diff
@@ -141,9 +147,13 @@ expand(N, i::Integer) = ntuple(_ -> i, N)
 
 
 """
-    maxpool(x, k::NTuple; pad=0, stride=k)
+    maxpool(x, k::NTuple{N, Integer}; pad=0, stride=k)
 
 Perform max pool operation with window size `k` on input tensor `x`.
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
```
mcabbott, line 154: Same here

src/pooling.jl
```diff
@@ -154,9 +164,13 @@ end
 
 
 """
-    meanpool(x, k::NTuple; pad=0, stride=k)
+    meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k)
 
 Perform mean pool operation with window size `k` on input tensor `x`.
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
```
mcabbott, line 171: and here

src/pooling.jl
```diff
@@ -166,7 +180,32 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
+* `p` is restricted to (0, Inf). Out-of-range `p` leads to undefined behavior.
+* `pad`: See [`pad_zeros`](@ref) for details.
+* `stride`: Stride for each spatial axis. `k` as default if not present.
+
+For each element `x` in (k × k) window, lpnormpool computes `(∑ x^p)^(1 / p)` as output.
+
+* When p = 1, lpnormpool(x, p, k) ./ prod(k) ≈ meanpool(x, k)
+* When p = 2, lpnormpool(x, p, k).^2 ./ prod(k) ≈ meanpool(x.^2, k)
```
mcabbott, line 197: These are code, and don't really need bullet points:
```suggestion
Thus `lpnormpool(x, 1, k) ./ prod(k) ≈ meanpool(x, k)` and 
`lpnormpool(x, 2, k).^2 ./ prod(k) ≈ meanpool(x.^2, k).
```

src/pooling.jl
```diff
@@ -166,7 +180,32 @@ function meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k) where N
...
+"""
+    lpnormpool(x, p::Number, k::NTuple{N, Integer}; pad=0, stride=k)
+
+Perform Lp pool operation with value of the Lp norm `p` and window size `k` on input tensor `x`, also known as LPPool in pytorch.
+This pooling operator from [Learned-Norm Pooling for Deep Feedforward and Recurrent Neural Networks](https://arxiv.org/abs/1311.1780).
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
+* `p` is restricted to (0, Inf). Out-of-range `p` leads to undefined behavior.
+* `pad`: See [`pad_zeros`](@ref) for details.
+* `stride`: Stride for each spatial axis. `k` as default if not present.
```
mcabbott, line 192: It accepts one integer for stride right? 

```suggestion
* `stride`: Either a tuple the same length as `k`, or one integer for all directions. Default is `k`.
```

src/pooling.jl
```diff
@@ -154,9 +164,13 @@ end
...
 
 """
-    meanpool(x, k::NTuple; pad=0, stride=k)
+    meanpool(x, k::NTuple{N, Integer}; pad=0, stride=k)
 
 Perform mean pool operation with window size `k` on input tensor `x`.
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
+* `pad`: See [`pad_zeros`](@ref) for details.
+* `stride`: Stride for each spatial axis. `k` as default if not present.
```
mcabbott, line 173: Same here:
```suggestion
* `stride`: Either an integer, or a tuple the same length as `k`. Default is `k`.
```

src/pooling.jl
```diff
@@ -141,9 +147,13 @@ expand(N, i::Integer) = ntuple(_ -> i, N)
...
 
 """
-    maxpool(x, k::NTuple; pad=0, stride=k)
+    maxpool(x, k::NTuple{N, Integer}; pad=0, stride=k)
 
 Perform max pool operation with window size `k` on input tensor `x`.
+
+* `x` and `k`: Usually, ndim(x) ∈ [3, 5], length(k) ∈ [1, 3], s.t. ndim(x) == length(k) + 2
+* `pad`: See [`pad_zeros`](@ref) for details.
+* `stride`: Stride for each spatial axis. `k` as default if not present.
```
mcabbott, line 156: Same here:
```suggestion
* `stride`: Either an integer, or a tuple the same length as `k`. Default is `k`.
```

src/pooling.jl
skyleaworlder: Thanks for your patient review! I have updated this in new commit.

src/pooling.jl
skyleaworlder: Thanks for your patient review! I have updated this in new commit.

src/pooling.jl
skyleaworlder: Sorry, my bad. I have fixed this in recent commit.

src/pooling.jl
skyleaworlder: OK, now it looks more specific and solid.

src/pooling.jl
skyleaworlder: Sorry, my bad. I missed this before. And now I add this following this review.

skyleaworlder:
@mcabbott Thanks for your patient review! I resolved all points metioned above and re-requested a new review just now :)

skyleaworlder:
@mcabbott @ToucheSir @CarloLucibello Thanks for your patience and kindness again! I really learn alot from this PR.

src/pooling.jl
```diff
@@ -103,7 +109,7 @@ end
 # Finally, let's generate auto-allocating versions of all our functions, for all backends:
 for backend in (Symbol(), :_direct, :_nnpack)
     # First make auto-allocating versions of the basic pooling calls:
-    for name in (:maxpool, :meanpool)
+    for name in (:maxpool, :meanpool, :lpnormpool)
         @eval begin
             function $(Symbol("$(name)$(backend)"))(
```
mcabbott, line 114: Maybe worth noting that this `similar` means that lower-precision input gives the correct output without promotion, and that integer input fails here (unlike other pooling):
```
julia> NNlib.lpnormpool(ones(Float32, 4,1,1), 2.001, (2,), stride=1)
3×1×1 Array{Float32, 3}:
[:, :, 1] =
 1.4139687
 1.4139687
 1.4139687

julia> NNlib.lpnormpool(ones(Int, 4,1,1), 2.001, (2,), stride=1)
ERROR: InexactError: Int64(1.4139686415190424)

julia> NNlib.maxpool(ones(Int, 4,1,1), (2,), stride=1)  # is OK with integers
3×1×1 Array{Int64, 3}:
[:, :, 1] =
 1
 1
 1

julia> NNlib.meanpool(ones(Int, 4,1,1), (2,), stride=1)  # is OK with integers
3×1×1 Array{Int64, 3}:
[:, :, 1] =
 1
 1
 1
```

PR #450, opened 2023-01-02
Avoid `maximum` in `softmax`

mcabbott:
I think `max` is very careful about NaN & signed zero, and hence `maximum` is surprisingly slow. This uses the `@fastmath` version: 
```
julia> let x = randn(Float32, 100, 1000)
         @btime softmax($x)
         @btime _softmax($x)  # this PR
       end;
  min 1.344 ms, mean 1.422 ms (7 allocations, 398.91 KiB)
  min 647.875 μs, mean 707.579 μs (3 allocations, 394.73 KiB)
```

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -17,7 +17,7 @@ See also [`logsoftmax`](@ref).
 
 # Examples
 
-```jldoctest softmax; filter = r"[+-]?([0-9]*[.])?[0-9]+(f[+-]*[0-9])?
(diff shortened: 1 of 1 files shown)


CarloLucibello:
We could remove the doctest

CarloLucibello (approved):

Sleort:
The new `fast_maximum` isn't always faster than `maximum`, ref. https://github.com/FluxML/NNlib.jl/pull/456#issuecomment-1370942366 ...

PR #454, opened 2023-01-03
Add `dropout`

mcabbott:
This adds a `dropout` function.

At least on CPU, this should be about twice as fast as the one in Flux, and use less memory. There are some Flux PRs, but the code is messy, and I got lost & started from scratch.

It's not quite the fastest variant. The gradient can be faster if we believe that every zero in the output comes from the mask, never from the input. That may not be a terrible assumption -- if you don't have a `relu` before it, zeros are rare; if you do, then the gradient associated to a zero will be discarded anyway. But... perhaps it's best to be correct.

<details>

I can put a gist of these variations somewhere. But local benchmarks are:

```julia
let x = randn(Float32, 100, 100)
    @btime dropout($x, 0.3) # Flux
    @btime mydropout($x, 0.3)  # sans mask -- this PR
    @btime mydropout2($x, 0.3)  # with mask
end;
  # min 10.375 μs, mean 19.879 μs (13 allocations, 78.45 KiB)
  # min 5.292 μs, mean 10.736 μs (2 allocations, 39.11 KiB)
  # min 12.958 μs, mean 17.476 μs (5 allocations, 44.67 KiB)

let x = randn(Float32, 100, 100)
    @btime gradient(x -> sum(dropout(x, 0.3)), $x) # Flux
    @btime gradient(x -> sum(mydropout(x, 0.3)), $x)  # sans mask... but sadly wrong answers
    @btime gradient(x -> sum(mydropout2(x, 0.3)), $x)  # with boolean mask
    @btime gradient(x -> sum(mydropout3(x, 0.3)), $x)  # keep random numbers -- this PR
end;
  # min 28.292 μs, mean 49.702 μs (70 allocations, 158.55 KiB)
  # min 12.500 μs, mean 22.952 μs (20 allocations, 78.70 KiB)
  # min 26.916 μs, mean 37.291 μs (37 allocations, 84.69 KiB)
  # min 13.208 μs, mean 26.490 μs (23 allocations, 117.92 KiB)


let x = randn(Float32, 100, 100)
    @btime dropout($x, 0.3; dims=1)
    @btime mydropout($x, 0.3; dims=1)
    @btime mydropout2($x, 0.3; dims=1)
end;
  # min 1.363 μs, mean 5.047 μs (3 allocations, 39.59 KiB)
  # min 1.346 μs, mean 6.003 μs (3 allocations, 39.59 KiB)
  # min 1.675 μs, mean 5.735 μs (5 allocations, 39.72 KiB)

let x = randn(Float32, 100, 100)
    @btime gradient(x -> sum(dropout(x, 0.3; dims=1)), $x)
    @btime gradient(x -> sum(mydropout(x, 0.3; dims=1)), $x)
    @btime gradient(x -> sum(mydropout2(x, 0.3; dims=1)), $x)
    @btime gradient(x -> sum(mydropout3(x, 0.3; dims=1)), $x)
    @btime gradient(x -> sum(mydropout4(x, 0.3; dims=1)), $x)  # changed to same strategy as dims=:
end;
  # min 25.458 μs, mean 47.425 μs (108 allocations, 122.16 KiB)
  # min 8.722 μs, mean 17.797 μs (21 allocations, 79.19 KiB)
  # min 13.709 μs, mean 24.887 μs (33 allocations, 79.61 KiB)
  # min 8.680 μs, mean 18.442 μs (21 allocations, 79.19 KiB)
  # min 9.916 μs, mean 22.267 μs (21 allocations, 79.25 KiB)  # slightly slower, but simpler
```

</details>

Edit: timing things on the GPU, at first the `dims=:` case seemed slow, but now it's not...

<details>

```julia
# ] add https://github.com/mcabbott/NNlib.jl/tree/dropout

using NNlib, CUDA, BenchmarkTools, Zygote, Flux, ChainRulesCore, Random

NNlib._rng_from_array(::CuArray) = CUDA.default_rng()

NNlib.dropout(cu(ones(3,5)), 0.3)
NNlib.dropout(cu(ones(3,5)), 0.4; dims=1)

Flux.dropout(cu(ones(3,5)), 0.3)
Flux.dropout(cu(ones(3,5)), 0.4; dims=1)

# To try out the copy-the-RNG idea, only for gradients:

@eval Base.copy(r::CUDA.RNG) = $(Expr(:new, CUDA.RNG, :(r.seed), :(r.counter)))

using NNlib: _fast_broadcast!

mydropout6(args...; kw...) = dropout(args...; kw...)  # same forward as this PR
function ChainRulesCore.rrule(::typeof(mydropout6), rng::AbstractRNG, x::AbstractArray, p::Real; dims = :)
    dup_rng = copy(rng)   # save for backward pass
    y = dropout(dup_rng, x, p; dims)  # for other dims, could re-use buffer...
    val = convert(eltype(y), 1/(1-p))
    function back6(Δ)
        dims isa Colon || error("will be wrong")
        dx = rand!(dup_rng, similar(y))
        _fast_broadcast!(dx, unthunk(Δ)) do q, dy
            (q>p) * val * dy
        end
        (NoTangent(), NoTangent(), dx, NoTangent())
    end
    y, back6
end

ChainRulesCore.@non_differentiable CUDA.default_rng()

mydropout6(CUDA.default_rng(), cu(ones(3,5)), 0.3)
gradient(x -> sum(mydropout6(CUDA.default_rng(), x, 0.3)), cu(ones(3,5)))


##### Forward times & allocations

julia> let x = CUDA.rand(100, 1000)
           @btime CUDA.@sync Flux.dropout($x, 0.3)
           @btime CUDA.@sync NNlib.dropout($x, 0.3)
           println()
           @btime CUDA.@sync Flux.dropout($x, 0.3; dims=1)
           @btime CUDA.@sync NNlib.dropout($x, 0.3; dims=1)
       end;
  49.163 μs (94 allocations: 4.27 KiB)
  29.623 μs (48 allocations: 2.31 KiB)

  43.097 μs (85 allocations: 4.05 KiB)
  42.857 μs (85 allocations: 3.98 KiB)

# one pass strategy, for the last:
  39.118 μs (88 allocations: 5.33 KiB)


# First run?
  # 49.339 μs (94 allocations: 4.27 KiB)
  # 261.194 μs (41 allocations: 392.56 KiB)  # awful

  # 42.835 μs (85 allocations: 4.05 KiB)
  # 47.702 μs (77 allocations: 4.05 KiB)

julia> let x = CUDA.rand(1000, 10_000)
           CUDA.@time Flux.dropout(x, 0.3)
           CUDA.@time NNlib.dropout(x, 0.3)
           println()
           CUDA.@time Flux.dropout(x, 0.3; dims=1)
           CUDA.@time NNlib.dropout(x, 0.3; dims=1)
       end;
  0.000415 seconds (94 CPU allocations: 4.266 KiB) (2 GPU allocations: 76.294 MiB, 5.40% memmgmt time)
  0.000269 seconds (48 CPU allocations: 2.312 KiB) (1 GPU allocation: 38.147 MiB, 2.50% memmgmt time)

  0.000185 seconds (86 CPU allocations: 4.062 KiB) (2 GPU allocations: 38.151 MiB, 6.87% memmgmt time)
  0.000186 seconds (86 CPU allocations: 4.000 KiB) (2 GPU allocations: 38.151 MiB, 4.62% memmgmt time)

# one pass:
  0.000214 seconds (88 CPU allocations: 5.328 KiB) (2 GPU allocations: 38.151 MiB, 3.88% memmgmt time)

# Earlier:
  # 0.000474 seconds (94 CPU allocations: 4.266 KiB) (2 GPU allocations: 76.294 MiB, 2.89% memmgmt time)
  # 0.046926 seconds (41 CPU allocations: 38.149 MiB) (1 GPU allocation: 38.147 MiB, 0.02% memmgmt time)

  # 0.000227 seconds (86 CPU allocations: 4.062 KiB) (2 GPU allocations: 38.151 MiB, 12.16% memmgmt time)
  # 0.000235 seconds (78 CPU allocations: 7.641 KiB) (2 GPU allocations: 38.151 MiB, 4.83% memmgmt time)


##### Gradient times & allocations

julia> let x = CUDA.randn(100, 1000), rng = CUDA.default_rng()
           @btime CUDA.@sync gradient(x -> sum(Flux.dropout(x, 0.3)), $x)
           @btime CUDA.@sync gradient(x -> sum(NNlib.dropout(x, 0.3)), $x)
           @btime CUDA.@sync gradient(x -> sum(mydropout6($rng, x, 0.3)), $x)
           println()
           @btime CUDA.@sync gradient(x -> sum(Flux.dropout(x, 0.3; dims=1)), $x)
           @btime CUDA.@sync gradient(x -> sum(NNlib.dropout(x, 0.3; dims=1)), $x)
       end;
  225.455 μs (364 allocations: 16.27 KiB)
  133.620 μs (242 allocations: 11.19 KiB)
  146.186 μs (258 allocations: 12.05 KiB)

  342.355 μs (509 allocations: 22.77 KiB)
  165.137 μs (284 allocations: 12.70 KiB)


# First try, without mydropout6
  # 194.412 μs (364 allocations: 16.27 KiB)
  # 318.048 μs (235 allocations: 401.44 KiB)  # bad, but not because of alloc

  # 259.827 μs (509 allocations: 22.77 KiB)
  # 150.939 μs (276 allocations: 12.77 KiB)  # great



julia> let x = CUDA.randn(1000, 10_000), rng = CUDA.default_rng()
           CUDA.@time gradient(x -> sum(Flux.dropout(x, 0.3)), x)
           CUDA.@time gradient(x -> sum(NNlib.dropout(x, 0.3)), x)
           CUDA.@time gradient(x -> sum(mydropout6(rng, x, 0.3)), x)
           println()
           CUDA.@time gradient(x -> sum(Flux.dropout(x, 0.3; dims=1)), x)
           CUDA.@time gradient(x -> sum(NNlib.dropout(x, 0.3; dims=1)), x)
       end;
  0.116546 seconds (61.41 k CPU allocations: 2.759 MiB, 26.81% gc time) (7 GPU allocations: 190.735 MiB, 8.33% memmgmt time)
  0.078563 seconds (52.12 k CPU allocations: 2.933 MiB) (6 GPU allocations: 152.589 MiB, 0.07% memmgmt time)
  0.078514 seconds (42.42 k CPU allocations: 2.552 MiB) (5 GPU allocations: 114.442 MiB, 0.07% memmgmt time)

  0.090443 seconds (54.14 k CPU allocations: 2.910 MiB) (8 GPU allocations: 152.596 MiB, 0.07% memmgmt time)
  0.083492 seconds (44.09 k CPU allocations: 2.394 MiB) (6 GPU allocations: 114.445 MiB, 0.07% memmgmt time)

# First try, without mydropout6
  # 0.070830 seconds (35.09 k CPU allocations: 1.953 MiB) (7 GPU allocations: 190.735 MiB, 0.10% memmgmt time)
  # 0.149181 seconds (52.21 k CPU allocations: 41.074 MiB, 14.75% gc time) (6 GPU allocations: 152.589 MiB, 0.18% memmgmt time)

  # 0.087902 seconds (54.21 k CPU allocations: 2.912 MiB) (8 GPU allocations: 152.596 MiB, 0.07% memmgmt time)
  # 0.082117 seconds (44.14 k CPU allocations: 2.399 MiB) (6 GPU allocations: 114.445 MiB, 0.06% memmgmt time)


##### CPU, on cyclops

julia> let x = rand(100, 1000)
           @btime Flux.dropout($x, 0.3)
           @btime NNlib.dropout($x, 0.3)
           println()
           @btime Flux.dropout($x, 0.3; dims=1)
           @btime CUDA.@sync NNlib.dropout($x, 0.3; dims=1)
       end;
  491.842 μs (13 allocations: 1.53 MiB)
  297.053 μs (2 allocations: 781.30 KiB)

  139.822 μs (3 allocations: 782.17 KiB)
  143.676 μs (3 allocations: 782.17 KiB)

# one pass strategy, for the last:
  143.043 μs (3 allocations: 782.17 KiB)
```

</details>

### DIFF
--- a/src/dropout.jl
+++ b/src/dropout.jl
@@ -1,0 +1,152 @@
+using Random, ChainRulesCore
+
+"""
+    dropout([rng], A, p; dims=:)
+
+Returns an array in which each element of `A` is either replaced with zero,
+with probability `p`, or else multiplied by `1/(1-p)`.
+
+By default every element is treated independently.
+With `dims=1`, a choice is made for every value of the 1st index
+i.e. each row of a matrix is either zero or not.
+
+Optional first argument is the random number generator used.
+
+# Examples
+```
+julia> dropout(ones(2, 10), 1/5)
+2×10 Matrix{Float64}:
+ 1.25  1.25  0.0   1.25  1.25  1.25  1.25  1.25  1.25  1.25
+ 1.25  1.25  1.25  0.0   1.25  1.25  0.0   1.25  1.25  1.25
+
+julia> mean(dropout(ones(10^4, 5), 0.3), dims=1)
+1×5 Matrix{Float64}:
+ 0.996  1.00171  1.00629  0.998714  0.992429
+
+julia> dropout(ones(5, 5), 0.7, dims=1)  # whole row the same
+5×5 Matrix{Float64}:
+ 3.33333  3.33333  3.33333  3.33333  3.33333
+ 0.0      0.0      0.0      0.0      0.0
+ 0.0      0.0      0.0      0.0      0.0
+ 3.33333  3.33333  3.33333  3.33333  3.33333
+ 0.0      0.0      0.0      0.0      0.0
+
+julia> mean(dropout(ones(10^4, 5), 0.3, dims=1), dims=1)
+1×5 Matrix{Float64}:
+ 1.00571  1.00571  1.00571  1.00571  1.00571
+```
+"""
+dropout(A::AbstractArray, p::Real; dims = :) = dropout(_rng_from_array(A), A, p; dims)
+
+function dropout(rng::AbstractRNG, A::AbstractArray, p::Real; dims = :)
+    T = float(eltype(A))
+    0 <= p <= 1 || throw(ArgumentError("dropout expects a probability 0 <= p <= 1"))
+    if p > 0
+        dst = similar(A, T)
+        pT = convert(real(T), p)
+        _dropout!(rng, dst, A, pT, dims)
+    else
+        # Not so sure we want fast paths... this tries but doesn't guarantee type-stability,
+        # and the rrule does not have such a fast paths.
+        convert(AbstractArray{T}, A)
+    end
+end
+
+"""
+    dropout!(B, A, p; dims=:)
+
+This does exactly `B .= dropout(A, p; dims)`,
+or rather, it's the implementation of out-of-place [`dropout`](@ref).
+"""
+function dropout!(dst::AbstractArray, src::AbstractArray, p::Real; dims=:)
+    size(dst) == size(src) || throw(DimensionMismatch("dropout! expects output array the same size as input"))
+    0 <= p <= 1 || throw(ArgumentError("dropout expects a probability 0 <= p <= 1"))
+    if p > 0
+        rng = _rng_from_array(A)
+        pT = convert(real(eltype(dst)), p)
+        _dropout!(rng, dst, src, pT, dims)
+    else
+        copyto!(dst, src)
+    end
+end
+
+# This is the easy case in that we can safely use the output array for random numbers.
+function _dropout!(rng::AbstractRNG, dst::AbstractArray, src::AbstractArray, p::Real, dims::Colon)
+    val = convert(eltype(dst), 1/(1-p))
+    rand!(rng, dst)
+    # dst .= (dst.>p) .* val .* src  # hits a SIMD bug
+    _fast_broadcast!(dst, src) do q, x
+        (q>p) * val * x
+    end
+    dst
+end
+
+# For other dims, we we do need to allocate something.
+function _dropout!(rng::AbstractRNG, dst::AbstractArray, src::AbstractArray, p::Real, dims)
+    tmp = similar(dst, ntuple(d -> d in dims ? size(src,d) : 1, ndims(src)))
+    rand!(rng, tmp)
+    val = convert(eltype(dst), 1/(1-p))
+    # One-pass strategy:
+    # dst .= (tmp.>p) .* val .* src
+    # Two-pass strategy:
+    _fast_broadcast!(tmp) do q
+        (q>p) * val
+    end
+    dst .= tmp .* src
+end
+
+# The gradient needs to keep the random choices made, thus store at least a BitArray,
+# but the following way turns out to be faster & simpler:
+function ChainRulesCore.rrule(::typeof(dropout), rng::AbstractRNG, A::AbstractArray, p::Real; dims = :)
+    T = float(eltype(A))
+    val = convert(T, 1/(1-p))
+    keep = if dims isa Colon
+        similar(A, T) 
+    else
+        similar(A, T, ntuple(d -> d in dims ? size(A,d) : 1, ndims(A)))
+    end
+    rand!(rng, keep)
+    Y = @. (keep>p) * A * val
+    function dropout_back(Δ)
+        dY = unthunk(Δ)
+        dA = @. (keep>p) * dY * val
+        (NoTangent(), NoTangent(), dA, NoTangent())
+    end
+    return Y, dropout_back
+end
+
+"""
+    _fast_broadcast!(f, x, y, z...)
+
+This does `x .= f.(x, y, z...)`, but works around
+an issue with broadcasting that prevents SIMD in such cases.
+Can be removed once https://github.com/JuliaLang/julia/issues/43153 is fixed.
+
+Not intended for general use. Does not check sizes!
+"""
+function _fast_broadcast!(f::F, x::Array, yz...) where {F<:Function}
+    bc = Broadcast.instantiate(Broadcast.broadcasted(f, x, yz...))
+    @simd ivdep for I in eachindex(bc)
+        @inbounds x[I] = bc[I]
+    end
+    return x
+end
+function _fast_broadcast!(f::F, x::AbstractArray, yz...) where {F<:Function}
+    # CUDA does not suffer from this bug
+    broadcast!(f, x, x, yz...)
+end
+
+
+"""
+    _rng_from_array(x)
+
+Return the random number generator most appropriate for `x`:
+`CUDA.default_rng()` for `CuArray`,
+else `Random.default_rng()`
+"""
+_rng_from_array(::AbstractArray) = Random.default_rng()
+# _rng_from_array(::CuArray) = CUDA.default_rng()
+
+@non_differentiable _rng_from_array(::Any)
+
+
--- a/src/dropout.jl
+++ b/src/dropout.jl
@@ -72,23 +72,25 @@ end
 
 # This is the easy case in that we can safely use the output array for random numbers.
 function _dropout!(rng::AbstractRNG, dst::AbstractArray, src::AbstractArray, p::Real, dims::Colon)
-    val = convert(eltype(dst), 1/(1-p))
+    T = real(eltype(dst))
+    val = convert(T, 1/(1-p))
     rand!(rng, dst)
     ## This is what we want, but it hits a SIMD bug, solved by _fast_broadcast!
     # dst .= (dst.>p) .* val .* src
     _fast_broadcast!(dst, src) do q, x
-        (q>p) * val * x
+        ((real(q)>p) * val) * x
     end
     dst
 end
 
 # For other dims, we we do need to allocate something.
 function _dropout!(rng::AbstractRNG, dst::AbstractArray, src::AbstractArray, p::Real, dims)
-    tmp = similar(dst, ntuple(d -> d in dims ? size(src,d) : 1, ndims(src)))
+    T = real(eltype(dst))
+    tmp = similar(dst, T, ntuple(d -> d in dims ? size(src,d) : 1, ndims(src)))
     rand!(rng, tmp)
-    val = convert(eltype(dst), 1/(1-p))
+    val = convert(T, 1/(1-p))
     ## One-pass strategy -- faster on GPU
-    dst .= (tmp.>p) .* val .* src
+    dst .= ((tmp.>p) .* val) .* src
     ## Two-pass strategy -- slightly faster on some CPUs?
     # _fast_broadcast!(tmp) do q
     #     (q>p) * val
@@ -99,7 +101,7 @@ end
 # The gradient needs to keep the random choices made, thus store at least a BitArray,
 # but the following way turns out to be faster & simpler:
 function ChainRulesCore.rrule(::typeof(dropout), rng::AbstractRNG, A::AbstractArray, p::Real; dims = :)
-    T = float(eltype(A))
+    T = float(real(eltype(A)))
     val = convert(T, 1/(1-p))
     keep = if dims isa Colon
         similar(A, T) 
@@ -107,10 +109,10 @@ function ChainRulesCore.rrule(::typeof(dropout), rng::AbstractRNG, A::AbstractAr
         similar(A, T, ntuple(d -> d in dims ? size(A,d) : 1, ndims(A)))
     end
     rand!(rng, keep)
-    Y = @. (keep>p) * A * val
+    Y = @. ((keep>p) * val) * A
     function dropout_back(Δ)
         dY = unthunk(Δ)
-        dA = @. (keep>p) * dY * val
+        dA = @. ((keep>p) * val) * dY
         (NoTangent(), NoTangent(), dA, NoTangent())
     end
     return Y, dropout_back
--- a/src/dropout.jl
+++ b/src/dropout.jl
@@ -1,4 +1,3 @@
-using Random, ChainRulesCore
 
 """
     dropout([rng], A, p; dims=:)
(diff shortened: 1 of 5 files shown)


ToucheSir:
One trick for GPU is to save the RNG state and lazily generate the mask in a kernel instead of materializing it (see https://triton-lang.org/master/getting-started/tutorials/04-low-memory-dropout.html for an implementation). This works because the default RNG is counter-based and doesn't have any dependence on previous states. Wikipedia tells me Xoshiro does not fall into this category, but we'd likely be doing a linear traversal on CPU and allocations are cheap(er) anyhow. All this to say that as long as things aren't slower on GPU, we can always optimize them separately in NNlibCUDA.

mcabbott:
Ah that's an idea I did not consider, would probably pay on CPU too. 

Copying Xoshiro is certainly fast and appears to work. It's just some bits... where do you read that it has other state?
```
julia> let rs = [copy(Random.default_rng()) for _ in 1:10]
         @test rand(rs[1]) == rand(rs[2])
         @test randn(rs[3], 3) == randn(rs[4], 3)
         @test rand!(rs[5], similar(1:0, 10, 10)) == rand!(rs[6], similar(1:0, 10, 10))
         rand(rs[1])
       end
0.8655824473102265

julia> dump(copy(Random.default_rng()))
Xoshiro
  s0: UInt64 0x3c8c337c7236f4ab
  s1: UInt64 0x5758ee14bcde3f3c
  s2: UInt64 0xb7bbc6e271015391
  s3: UInt64 0xb76992934d86a2b9
```
What I'm not sure of is whether `copy(::AbstractRNG)` is guaranteed to work like this. Copying `MersenneTwister` takes a μs.

ToucheSir:
> where do you read that it has other state?

I was going off https://en.wikipedia.org/wiki/Counter-based_random_number_generator_(CBRNG)#Background. It's not that it has more state, but that you can't get the Nth random value without first calculating some previous N-x values. Whereas for the Philox RNG CUDA.jl uses, you just set the counter to however many steps forward you want and it gives you a value.

> What I'm not sure of is whether `copy(::AbstractRNG)` is guaranteed to work like this.

This was my concern as well. We could always use dispatch to have a fallback path for unknown RNGs that materializes the mask.

src/dropout.jl
```diff
@@ -0,0 +1,152 @@
...
+function ChainRulesCore.rrule(::typeof(dropout), rng::AbstractRNG, A::AbstractArray, p::Real; dims = :)
+    T = float(eltype(A))
+    val = convert(T, 1/(1-p))
+    keep = if dims isa Colon
+        similar(A, T) 
+    else
+        similar(A, T, ntuple(d -> d in dims ? size(A,d) : 1, ndims(A)))
+    end
+    rand!(rng, keep)
+    Y = @. (keep>p) * A * val
```
ToucheSir, line 109: If the mask idea doesn't work out, another idea for making this type stable while still allowing short-circuiting on 0 and 1 is to make a lazy mask wrapper which only creates the mask if necessary. Depending on how devious we want to be, we can even use `promote_op` or a custom `mask_type(xtype::Type{A}, ...) where A <: AbstractArray` to determine what type the wrapped mask should be (since we can't determine it in the short-circuiting case).

mcabbott:
> but that you can't get the Nth random value without first calculating some previous N-x values to advance the internal state

Ah but that's fine I think. We don't need to skip into the future. We only need that the two copies used forward and backward produce the same, starting from whatever we happen to start.

> We could always use dispatch to have a fallback path for unknown RNGs that materializes the mask.

Oh man, I'd be quite keen not to have to maintain two paths. (Already the no-gradient and gradient versions differ a bit here...)

ToucheSir:
Yes, the skipping is more relevant on GPU. IIRC there's a way to split a `Xoshiro` RNG for multithreading, so that's likely not an issue on CPU.

On divergent paths, I don't think it's too too bad. The dispatch split point could be on what Flux calls `dropout_mask`. We then override `broadcast(*, ::LowMemDropoutMask, x)` or introduce our own `apply_mask` function with a `x .* mask` fallback. The rrule is technically not required because AD will do `mask .* dy` automatically.

mcabbott:
I think we should do this, it's a step forward. 

The copy-the-RNG story does work, and save memory on the GPU. But needs some constructor methods, and perhaps consultation with experts about safety. And copying MerseinTwister on 1.6 is pretty slow... anyway we should kick the can down the road I think. This saves memory & time over Flux. It needs a 2-line NNlibCUDA PR, and some tests there.

src/dropout.jl
```diff
@@ -0,0 +1,156 @@
...
+    bc = Broadcast.instantiate(Broadcast.broadcasted(f, x, yz...))
+    @simd ivdep for I in eachindex(bc)
+        @inbounds x[I] = bc[I]
+    end
+    return x
+end
+function _fast_broadcast!(f::F, x::AbstractArray, yz...) where {F<:Function}
+    # CUDA does not suffer from this bug
+    broadcast!(f, x, x, yz...)
+end
```
ToucheSir, line 144: Move this to utils.jl?

src/dropout.jl
```diff
@@ -0,0 +1,156 @@
...
+function ChainRulesCore.rrule(::typeof(dropout), rng::AbstractRNG, A::AbstractArray, p::Real; dims = :)
+    T = float(real(eltype(A)))
+    val = convert(T, 1/(1-p))
+    keep = if dims isa Colon
+        similar(A, T) 
+    else
+        similar(A, T, ntuple(d -> d in dims ? size(A,d) : 1, ndims(A)))
+    end
+    rand!(rng, keep)
+    Y = @. ((keep>p) * val) * A
```
ToucheSir, line 112: Is doing `(keep>p) * val)` twice faster than doing https://github.com/FluxML/NNlib.jl/pull/454/files#diff-e84a46a10c6ff3b32f9a7f735fafee1838628a5261e84f981ec8abcf845547f2R80-R82 once? If so, that's unfortunate.

src/dropout.jl
```diff
@@ -0,0 +1,156 @@
...
+    end
+end
+
+"""
+    dropout!(B, A, p; dims=:)
+
+This does exactly `B .= dropout(A, p; dims)`,
+or rather, it's the implementation of out-of-place [`dropout`](@ref).
+"""
+function dropout!(dst::AbstractArray, src::AbstractArray, p::Real; dims=:)
```
ToucheSir, line 60: IMO if there is a single `dropout!` function, it should have an explicit `rng` parameter. Adding an overload that uses `_rng_from_array` like `dropout` would work too.

src/dropout.jl
```diff
@@ -0,0 +1,156 @@
...
+
+"""
+    _rng_from_array(x)
+
+Return the random number generator most appropriate for `x`:
+`CUDA.default_rng()` for `CuArray`, else `Random.default_rng()`
+"""
+_rng_from_array(::AbstractArray) = Random.default_rng()
+
+@non_differentiable _rng_from_array(::Any)
```
ToucheSir, line 155: Any reason not to copy https://github.com/FluxML/Flux.jl/blob/ee78ce3cefb027228413f8edace3c0385139d786/src/utils.jl#L36-L49 wholesale (minus the CUDA overload)?

ToucheSir (commented):
No objections from me. All the aforementioned tricks can be done as additive or at least non-breaking changes in future PRs.

src/dropout.jl
mcabbott: Just simplifying a bit, I couldn't figure out why there were so many functions. Why make a different choice on 1.6
```
julia> using Random

julia> Random.default_rng()
MersenneTwister(0x9687b6121c4ccb062f473c9c3c8bccc6)

julia> Random.GLOBAL_RNG
Random._GLOBAL_RNG()

julia> VERSION
v"1.6.0"
```
compared to master:
```
julia> using Random

julia> Random.default_rng()
TaskLocalRNG()

julia> Random.GLOBAL_RNG
Random._GLOBAL_RNG()

julia> VERSION
v"1.10.0-DEV.204"
```

src/dropout.jl
mcabbott: I think it was very close to a wash. This way lets you have just one `back` function, instead of two cases.

src/dropout.jl
mcabbott: Can do. But maybe I vote to move it once there's a second use?

src/dropout.jl
ToucheSir: I don't remember now, but based on https://github.com/FluxML/Flux.jl/pull/1849#issuecomment-1022687847 it might've been related to thread safety?

src/dropout.jl
mcabbott: ok I made two methods now, optional RNG.

src/dropout.jl
ToucheSir: Cthulhu tells me that `rand(...)` uses `default_rng()` on 1.6 as well and it returns a thread-local RNG, so maybe this was much ado about nothing. cc @darsnack if I've missed something though.

src/dropout.jl
ToucheSir: I don't quite get the point about two `back` functions. Is it not sufficient to just `s/((keep>p) * val)/keep/g`? Zygote might like having to generate fewer `getproperty` calls for captured variables too.

ToucheSir: On second thought, doing this would mean `dropout` is no longer doubly-differentiable. I retract my previous suggestion.

CarloLucibello (approved):

PR #455, opened 2023-01-03
implement dot_product_attention

### LINKED ISSUE #385: Multi-head attention?
Now that we're seeing models with attention layers in more than one downstream (e.g. FluxML/Metalhead.jl#105 and Transformers.jl), it may be time to consider pulling some building blocks into NNlib. CUDA.jl already wraps cuDNN's MHA too, see https://github.com/JuliaGPU/CUDA.jl/blob/27c87a6f261aa7964d797e8fe4bf33b46c1a185e/test/cudnn/multiheadattn.jl#L55.

### LINKED ISSUE #391: Batched multiplication support for ndims > 3
Currently NNlib.batchmul works with arrays with upto 3 dimensions. It would be nice if this could be upgraded to function similar to `numpy.matmul` or `torch.matmul` - this would help in a lot of models, especially some of the attention-based ones I'm working on 😅

### LINKED ISSUE #451: support arbitrary number of batch dimensions in batched_mul
### Motivation and description

use case is the multihead attention layer in https://github.com/FluxML/Flux.jl/pull/2146

### Possible Implementation

```julia
function NNlib.batched_mul(x::AbstractArray{T1,N}, y::AbstractArray{T2,N}) where {T1,T2,N}
  sz = size(x)[3:end]
  @assert sz == size(y)[3:end]
  x2 = reshape(x, size(x, 1), size(x, 2), :)
  y2 = reshape(y, size(y, 1), size(y, 2), :)
  z = NNlib.batched_mul(x2, y2)
  return reshape(z, size(z, 1), size(z, 2), sz...)
end
```

CarloLucibello:
Factored out from https://github.com/FluxML/Flux.jl/pull/2146

Fix #385 

In the process, also extends `batched_mul` to multiple batch dimensions. Fix #451 fix #391 

We may want to consider hooking cudnn in a later PR


src/attention.jl
```diff
@@ -0,0 +1,125 @@
...
+    # [α] = [kv_len, q_len, num_heads, batch_size]
+
+    if bias !== nothing
+        α = α .+ bias
+    end
+
+    if mask !== nothing
+        if mask === :causal
+            mask = make_causal_mask(α)
+        end
```
CarloLucibello, line 93: do we want this or we leave it to the users?

src/attention.jl
```diff
@@ -0,0 +1,125 @@
...
+```
+"""
+function dot_product_attention(q::AA{N}, k::AA{N}, v::AA{N}; num_heads=1, kws...) where N
+    batch_size = size(q)[3:end]
+    
+    batch_size == size(k)[3:end] == size(v)[3:end] || throw(ArgumentError("Batch dimensions have to be the same."))
+    size(q, 1) == size(k, 1) || throw(ArgumentError("First dimension in query and key has to be the same."))
+    size(k, 2) == size(v, 2)  || throw(ArgumentError("Second dimension in key and value has to be the same."))
+    
+    q, k, v = map(x -> reshape(x, size(x, 1), size(x, 2), :), (q, k, v))
```
mcabbott, line 39: Is this worth doing, or can we just start with the case taking only 3D arrays? 

If we do want to take 4D, then probably that should be a different method which reshapes before & after. Then e.g. CuArray can overload just the 3D one.

src/attention.jl
```diff
@@ -0,0 +1,125 @@
+const AA3{T} = AbstractArray{T,3}
+const AA4{T} = AbstractArray{T,4}
+const AA{N,T} = AbstractArray{T,N}
+
+"""
+    dot_product_attention(query, key, value; [bias, droput_fn, mask, num_heads])
```
mcabbott, line 6: Bikeshedding, but can we have simpler keyword names? I think the ideal is one word 3-7 chars, no underscores, no single letters. 
```suggestion
    dot_product_attention(query, key, value; [bias::Bool, dropout::Function, mask::Bool, heads::Int])
```
Not so sure types help.

src/batched/batchedmul.jl
```diff
@@ -42,6 +44,16 @@ This will be copied, as doing so is faster than `batched_mul_generic!`.
 Both this `copy` and `batched_mul_generic!` produce `@debug` messages,
 and setting for instance `ENV["JULIA_DEBUG"] = NNlib` will display them.
 """
+
+function batched_mul(x::AbstractArray{T1,N}, y::AbstractArray{T2,N}) where {T1,T2,N}
```
mcabbott, line 48: My vote is to make this an internal `_batched_mul_4` or something for now. Partly because I think explaining what does and doesn't work becomes more complicated with this method. And that doesn't have to be solved to add attention.

mcabbott (commented):
I think this mostly looks good, and is about the right level -- fairly simple-to-read implementation with no magic.

The thing to get right now seems to be: Does this function line up fairly nicely with what CUDA provides, so that an overload `dot_product_attention(::CuArray, ...)` can smoothly provide the same functionality? From a quick look 

* It seems that that wants a weight array which if I understand right would be steps before this function:  https://github.com/JuliaGPU/CUDA.jl/blob/8a4cbdee50c716ff642eb3d9268f1a7ea4c29eb0/lib/cudnn/src/multiheadattn.jl#L20 

* I'm not entirely sure what its masking options are, what am I missing e.g. here https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSetAttnDescriptor

* Would need to read slower to see where it does & doesn't include bias, and dropout. 

Lots of random comments, but probably you should ignore some on internal details as these can be fixed, but the overall interface is the question.

src/attention.jl
CarloLucibello: We should allow the possibility to pass arbitrary masks. Causal masks are by far the prominent ones, for autoregressive decoding, but I think other types of masks, e.g. enforcing some kind of locality, are not so rare.

CarloLucibello: NeuralAttentionlib has a non-allocating causal mask https://github.com/chengchingwen/NeuralAttentionlib.jl/blob/master/src/mask/dataless.jl#L7
I'm not sure masking is a part of the implementation it is worth optimizing. 

Maybe we can add another keyword, `causal=true` to be used as an alternative to `mask`?

src/attention.jl
```diff
@@ -0,0 +1,125 @@
...
+    
+    # The following permutedims and batched_mul are equivalent to
+    # @tullio α[j, i, h, b] := q[d, h, i, b] * k[d, h, j, b]
+    kt = permutedims(k, (3, 1, 2, 4))
+    qt = permutedims(q, (1, 3, 2, 4))
+    α = batched_mul(kt, qt)
+    # [α] = [kv_len, q_len, num_heads, batch_size]
+
+    if bias !== nothing
+        α = α .+ bias
```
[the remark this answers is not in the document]
CarloLucibello: yes, bias is used for score reshaping based on locality and on context features. 
Can also be used for masking, but it is more intuitive and convenient to pass a boolean mask instead of a array with infs.

src/batched/batchedmul.jl
CarloLucibello: It's a pity to not make things available. Maybe I can leave the previous docstring unchanged and add a new one for the new method?

src/attention.jl
mcabbott: No, this line does a bunch of reshaping that isn't necessary in the common case of 3D input. That's what seems confusing and I think would be better handled by the public function having two methods.

src/attention.jl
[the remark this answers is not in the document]
mcabbott: If it must exist then  it probably ought to be a positional argument. AD with keywords is asking for pain.

src/attention.jl
CarloLucibello: done

CarloLucibello:
Regarding CUDNN, the descriptor there says
"Weight matrices WQ,i , WK,i , WV,i and WO,i play similar roles, adjusting vector lengths in q , K,V inputs and in the multi-head attention final output. The user can disable any or all projections by setting qProjSize , kProjSize , vProjSize or oProjSize arguments to zero."
so we can deactivate the projections to match the api introduced in this PR. Conversely, we can add overloads here in the future if we want to consider also the projections. 

I still have to look into dropout, masks and biases in cudnn

CarloLucibello:
I think this is good to go

src/attention.jl
```diff
@@ -0,0 +1,139 @@
...
+    # [logits] = [kv_len, q_len, nheads, batch_size]
+
+    if bias !== nothing
+        logits = logits .+ bias
+    end
+
+    if mask !== nothing
+        if mask === :causal
+            mask = make_causal_mask(logits)
+        end
```
darsnack, line 107: I think a cleaner API would be to let the `mask` keyword be a function. The `nothing` case is `mask = identity` and the causal case is `mask = make_causal_mask` (which I feel should be just `causal_mask` to be succinct).

Is there a reason to construct the mask on the fly? The calling layer in Flux can probably make and store the mask once. Then the other option is to allow `nothing` or an array. Then the user passes in `mask = causal_mask(ntoken)`.

darsnack: You're right, it shouldn't be `identity`, it should be `trues_like` though I'd be okay with `nothing` in order skip computing a mask at all.

My comment about constructing on the fly was not a performance concern. I just think it is more intuitive to pass in exactly the mask array I want used. It's an easier rule to remember and also scalable to whatever masking scheme is desired.

mcabbott: The downside is that you have to make an array the right size. If you have several layers and want the same scheme for each, then perhaps it's a pain. Whereas a function like `trues_like` is told the size automatically.

darsnack: That's true, but generally the size of this matrix which is `# of tokens X # of tokens` is known ahead of time. Even so, I agree that not needing to pass in this info is cleaner.

I mostly wanted to avoid "symbol switches" for arguments.

mcabbott: Yes to avoiding symbols. I like I like this `mask = trues_like` proposal the best so far.

One question I haven't looked at is what format the CUDNN thing is going to want.

ToucheSir: Instead of saying `mask` is either an array or callable, could we say it should be either an array or marker type for which one can override some `init_mask(x, k, v)` function? This would allow us to shift the conditionals out of the attention functions, while still allowing for relatively terse syntax like `mask = CausalMask()` when users don't want to precompute their own. You could imagine nice party tricks like passing `mask = I`.

mcabbott: https://github.com/FluxML/NNlib.jl/pull/460 is a go at this masked softmax idea. 

With that, the default of no mask can in fact be `mask = Returns(true)` here, instead of `trues_like`. And the terse causal mask can be `const causal_mask = triu ∘ trues_like`, or a function equivalent to this (maybe it can be more efficient, not sure `triu`  works on CuArrays). No conditionals required.

chengchingwen: `triu` only works on `AbstractMatrix`, which is not sufficient for the attention.

CarloLucibello: For this first implementation, I prefer to keep it more minimalistic and just accept `nothing` or arrays (I will remove `:causal`)

CarloLucibello:
`:causal` remove, now we only accept array masks. Good to go?

src/attention.jl
```diff
@@ -0,0 +1,137 @@
...
+  return mask
+end
+
+trues_like(x::AbstractArray, sz=size(x)) = fill!(similar(x, Bool, sz), true)
+falses_like(x::AbstractArray, sz=size(x)) = fill!(similar(x, Bool, sz), false)
+
+split_heads(x, nheads) = reshape(x, size(x, 1) ÷ nheads, nheads, size(x)[2:end]...)
+join_heads(x) = reshape(x, :, size(x)[3:end]...)
+
+@non_differentiable make_causal_mask(x)
```
CarloLucibello, line 134:
```suggestion
@non_differentiable make_causal_mask(::Any...)
```

test/attention.jl
```diff
@@ -0,0 +1,76 @@
...
+    y, α2 = dot_product_attention(q2, k2, k2; nheads=2)
+    @test α ≈ α2
+end
+
+@testset "specific results" begin
+    q = k = v = reshape([1:12;], 4, 3, 1) ./ 12
+    y, α = dot_product_attention(q, k, v; nheads=2)
+    ytrue = [0.4297536645089624, 0.5130869978422957, 0.6137914555895531, 0.6971247889228864, 0.46431026790247376, 0.5476436012358071, 0.6478764227436047, 0.731209756076938, 0.49773020657887745, 0.5810635399122107, 0.6804545876711346, 0.763787921004468]
+    ytrue = reshape(ytrue, 4, 3, 1)
+    αtrue = [0.3138955704910261, 0.3329478654910607, 0.35315656401791323, 0.264431440679808, 0.32820631493296265, 0.4073622443872293, 0.21921458153690657, 0.31838021718955445, 0.4624052012735389, 0.2886914482847165, 0.33124273666190807, 0.3800658150533755, 0.24123865285082136, 0.3238934260675431, 0.43486792108163547, 0.19843756756539277, 0.31176110185581074, 0.4898013305787966]
```
ToucheSir, line 30: Any way we could drop a few digits from these by adding a tolerance to the `isapprox` tests below?

src/attention.jl
```diff
@@ -0,0 +1,137 @@
...
+    # [logits] = [kv_len, q_len, nheads, batch_size]
+
+    if bias !== nothing
+        logits = logits .+ bias
+    end
+
+    if mask !== nothing
+        neginf = typemin(eltype(logits))
+        logits = ifelse.(mask, logits, neginf)
+    end
```
ToucheSir, line 108: WDYT about making these internal methods which dispatch on nothing? That way there's zero control flow and Zygote is happy. The main question is whether the additional code + complexity introduced would be worth the compile and runtime reduction.

ToucheSir (commented):
Mostly LGTM and appears to give us enough surface area for cuDNN. Just a couple final questions:

jondeuce:
This looks great, I've long thought that there should be some basic attention mechanisms available here.

Before settling on the API for passing masks, I thought I should mention that "non-boolean" masking such as [Attention with Linear Biases (ALiBi)](https://arxiv.org/abs/2108.12409) have been used for some pretty substantial projects, e.g. [BigScience's BLOOM](https://huggingface.co/bigscience/bloom), and it would be useful to have this option.

This would only be a slight generalization from what is currently implemented here (see Fig. 3 in the [ALiBi paper](https://arxiv.org/pdf/2108.12409.pdf)), and could easily be incorporated by dispatching on the type of the mask:
1. `mask::AbstractArray{Bool}` acts like `ifelse.(mask, logits, -Inf)`
2. `mask::AbstractArray{<:Real}` acts like `logits .+ mask`

This would actually be almost exactly what PyTorch does as well, which is a nice bonus. From the [`torch.nn.MultiheadAttention` docs](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html):
>Binary, byte, and float masks are supported. For a binary mask, a `True` value indicates that the corresponding position is not allowed to attend. For a byte mask, a non-zero value indicates that the corresponding position is not allowed to attend. For a float mask, the mask values will be added to the attention weight.

Interesting that PyTorch appears to have the meaning of `true` and `false` reversed compared to what is implemented in this PR, but Keras has the same convention as this PR (see the [docs](https://keras.io/api/layers/attention_layers/attention/) and the [code](https://github.com/keras-team/keras/blob/db11e9674157c9125c9215b3bd4928675cf5af53/keras/layers/attention/base_dense_attention.py#L131)). Not sure which meaning is more natural :man_shrugging:.

CarloLucibello:
@jondeuce the attention bias in the PR already does what you suggest. Should we collapse `bias` into `mask` or keep the two separate? I was inspired by https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.dot_product_attention.html?highlight=attention
for the interface

jondeuce:
@CarloLucibello Ahh, yes that is a flexible approach too, and clearly covers `ALiBi`-style masks. It's funny, I did notice the bias being added, but my brain did not register that `bias` would be an additive mask, instead I thought of learnable biases like in `Dense` layers.

The more I think about it, the more I like the way you have it. They are orthogonal and mutually compatible ways to apply a mask, and I don't think combining them and then doing different operations based on the mask type is conceptually any simpler anyways.

ToucheSir (approved):

PR #456, opened 2023-01-04
`fast_maximum` also for `logsumexp`

Sleort:
In the spirit of #450 ...

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -141,13 +141,13 @@ Without `dims` keyword this returns a scalar.
 See also [`logsoftmax`](@ref).
 """
 function logsumexp(x::AbstractArray; dims = :)
-    max_ = maximum(x; dims)
+    max_ = fast_maximum(x; dims)
     @fastmath max_ .+ log.(sum(exp.(x .- max_); dims))
 end
 
 function rrule(::typeof(logsumexp), x; dims = :)
     # The gradient is `softmax`, but both compute `tmp` so it's worth saving.
-    max_ = maximum(x; dims)
+    max_ = fast_maximum(x; dims)
     @fastmath tmp = exp.(x .- max_)
     @fastmath y = max_ .+ log.(sum(tmp; dims))
     logsumexp_pullback(dy) = (NoTangent(), unthunk(dy) .* tmp ./ sum(tmp; dims))


CarloLucibello:
Can you make some benchmarks? just to gauge the impact of this

Sleort:
Hmm... It turns out that #450 actually is a performance regression in some cases. While we have
```julia
julia> let x = randn(Float32, 100, 1000)
           @btime old_softmax($x; dims=1)  #with maximum
           @btime softmax($x; dims=1)  #with fast_maximum
       end;
  1.155 ms (7 allocations: 398.91 KiB)
  504.926 μs (3 allocations: 394.73 KiB)
```
the ranking is reversed when I set `dims=:`:
```julia
julia> let x = randn(Float32, 100, 1000)
           @btime old_softmax($x; dims=:)
           @btime softmax($x; dims=:)
       end;
  659.604 μs (4 allocations: 390.70 KiB)
  691.194 μs (13 allocations: 390.92 KiB)
```
The same pattern can be seen for `logsumexp`:
```julia
julia> let x = randn(Float32, 100, 1000)
           @btime logsumexp($x; dims=1)  #with maximum
           @btime new_logsumexp($x; dims=1) #with fast_maximum
           @btime logsumexp($x; dims=:)
           @btime new_logsumexp($x; dims=:)
       end;
  833.593 μs (8 allocations: 402.97 KiB)
  616.084 μs (5 allocations: 402.86 KiB)
  578.131 μs (4 allocations: 390.70 KiB)
  654.552 μs (18 allocations: 391.02 KiB)
```
This is because `fast_maximum` is not always the fastest (!):
```julia
julia> let x = randn(Float32, 100, 1000)
           @btime maximum($x; dims=1)
           @btime fast_maximum($x; dims=1)
           @btime maximum($x; dims=:)
           @btime fast_maximum($x; dims=:)
       end;
  416.025 μs (4 allocations: 4.17 KiB)
  73.618 μs (1 allocation: 4.06 KiB)
  55.362 μs (0 allocations: 0 bytes)  #maximum(x; dims=:) is faster than
  111.392 μs (0 allocations: 0 bytes)  #... fast_maximuM(x; dims=:)
```

So maybe we should make `fast_maximum` dispatch on `dims`, like
```
fast_maximum(x; dims) = fast_maximum(x, dims)
fast_maximum(x::AbstractArray{T}, dims) where {T} = @fastmath reduce(max, x; dims, init = float(T)(-Inf))
fast_maximum(x::AbstractArray{T}, ::Colon) where {T} = maximum(x)
```
?

CarloLucibello:
> So maybe we should make fast_maximum dispatch on dims, like

seems worth doing, but in the meantime let's merge this and use fast_maximum consistently in the library

mcabbott:
In `softmax` I'm reasonably sure that fastmax doesn't change the NaN behaviour. Is this true of `logsumexp` too? 

I didn't see clear improvements in `logsumexp` when I tried, so I thought simpler to leave it alone.

PR #458, opened 2023-01-07
add `_rng_compat_array`

mcabbott:


CarloLucibello (approved):

src/dropout.jl
```diff
@@ -155,3 +157,6 @@ _rng_from_array(::AbstractArray) = Random.default_rng()
 
 @non_differentiable _rng_from_array(::Any)
 
+# This exists because `rand!(defalut_rng(), CUDA.rand(3))` ignores the RNG,
+# and Flux would prefer an error. NNlibCUDA will overloard it to produce that.
```
ToucheSir, line 161:
```suggestion
# This exists because `rand!(default_rng(), CUDA.rand(3))` ignores the RNG,
# and Flux would prefer an error. NNlibCUDA will overload it to produce that.
```

ToucheSir (approved):

src/dropout.jl
```diff
@@ -62,6 +63,7 @@ dropout!(B::AbstractArray, A::AbstractArray, p::Real; dims = :) = dropout!(_rng_
 function dropout!(rng::AbstractRNG, dst::AbstractArray, src::AbstractArray, p::Real; dims=:)
     size(dst) == size(src) || throw(DimensionMismatch("dropout! expects output array the same size as input"))
     0 <= p <= 1 || throw(ArgumentError("dropout expects a probability 0 <= p <= 1"))
+    _rng_compat_array(rng, A)
```
ToucheSir, line 66:
```suggestion
    _rng_compat_array(rng, src)
```
It also occurs to me that the docstring is not in line with the actual function signature. Any reason to use `A`/`B` instead of `dst`/`src` for the former?

ToucheSir:
Sorry, hit the review trigger a little early. Needs some typo fixes to make CI happy.

mcabbott:
Done the hacky way in https://github.com/FluxML/NNlib.jl/commit/9425e5060286891a012a7809ecd344682d30f403

src/dropout.jl
mcabbott: No real reason, just ended up this way.

PR #459, opened 2023-01-08
Faster `softmax`?

mcabbott:
This defines a `fast_softmax` which uses a low-accuracy `fast_exp`. It's about 5x faster on CPU.

On a GPU, the low-accuracy `exp` isn't faster at all. For small arrays, `fast_softmax` is faster, because it skips the `all(isfinite, max_)` check & thus avoids synchronisation. Thus https://github.com/FluxML/NNlibCUDA.jl/pull/63 should get all the benefit.

The alternative on CPU is to make an `Array` specialisation using LoopVectorization. That's not as quick as this `fast_exp` (about 2x slower for me) but several more digits of precision. This `fast_exp` is roughly Float16 precision, do we want that?

### DIFF
--- a/src/softmax.jl
+++ b/src/softmax.jl
@@ -91,6 +91,170 @@ end
 fast_maximum(x::AbstractArray{T}; dims) where {T} = @fastmath reduce(max, x; dims, init = float(T)(-Inf))
 
 """
+    fast_exp(x)
+
+For `x::Float32`, this is a much faster (about 20x)
+but much less accurate (about 0.1%) version of `exp`.
+All other real numbers call `@fastmath exp(x)`.
+
+Handles `Inf` but not `NaN`:
+```
+julia> xs = Tuple([0, 1, Inf32, -Inf32, NaN32]);
+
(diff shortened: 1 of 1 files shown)


ToucheSir:
It looks like there is a whole subset of literature for fast softmax approximations. I only read through https://arxiv.org/abs/2111.10770v1, but it has a nice list of prior art. Also of interest may be existing CPU-optimized softmax impls like [oneDNN](https://github.com/oneapi-src/oneDNN/blob/7f4088bd9a029114eb6b056ddf56e5e96f216fc0/src/cpu/cpu_softmax_list.cpp).

mcabbott:
Had not looked, but not surprised there's a literature by now! IIRC we dropped the NVidia one as it was slower than NNlib's. 

The immediate goal though is to make this part small compared to the matmul & permutations. Going by my times [here](https://github.com/FluxML/Flux.jl/pull/2146#issuecomment-1368594525) this gets us from roughly 50% to 10%.

Issue #463, opened 2023-01-20
Symmetric Padding


nikopj:
### Motivation and description

Different to reflection padding, as illustrated in ImageFiltering.jl: https://juliaimages.org/ImageFiltering.jl/stable/function_reference/#Boundaries-and-padding

nikopj:
Implemented by #465

PR #465, opened 2023-01-25
symmetric and circular padding

nikopj:
Implements and exports `pad_symmetric` and `pad_circular`, addressing #463. Implemented padding is consistent with [ImageFiltering.jl](https://juliaimages.org/ImageFiltering.jl/stable/function_reference/#Boundaries-and-padding).

### PR Checklist

- [x] Tests are added
- [x] Documentation

### DIFF
--- a/src/padding.jl
+++ b/src/padding.jl
@@ -169,7 +169,7 @@ of some length `2n` that specifies the left and right padding size
 for each of the dimensions in `dims`. If `dims` is not given, 
 it defaults to the first `n` dimensions.
 
-For integer `pad` input instead, it is applied on both sides
+If `pad` is an integer, it is applied on both sides
 on every dimension in `dims`. In this case, `dims` 
 defaults to the first `ndims(x)-2` dimensions 
 (i.e. excludes the channel and batch dimension). 
@@ -230,10 +230,10 @@ of some length `2n` that specifies the left and right padding size
 for each of the dimensions in `dims`. If `dims` is not given, 
 it defaults to the first `n` dimensions.
 
-For integer `pad` input instead, it is applied on both sides
+If `pad` is an integer, it is applied on both sides
 on every dimension in `dims`. In this case, `dims` 
 defaults to the first `ndims(x)-2` dimensions 
-(i.e. excludes the channel and batch dimension).
+(i.e. excludes the channel and batch dimension). 
 
 See also [`pad_repeat`](@ref), [`pad_symmetric`](@ref), [`pad_circular`](@ref), and [`pad_constant`](@ref).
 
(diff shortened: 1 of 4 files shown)


test/padding.jl
```diff
@@ -99,5 +99,53 @@ end
...
       
-  gradtest(x -> pad_repeat(x, (2,2,2,2)), rand(2,2,2))
+  # pad_reflect needs larger test input as padding must 
+  # be strictly less than array size in that dimension
+  gradtest(x -> pad_reflect(x, (2,2,2,2)), rand(3,3,3))
+end
+
+@testset "padding symmetric" begin
+  y = pad_symmetric(reshape(1:9, 3, 3), (2,2), dims=2)
+  @test y ==  [ 4  1  1  4  7  7  4
```
ToucheSir, line 109: Some strangeness with formatting in this file

src/padding.jl
```diff
@@ -277,8 +277,127 @@ function pad_reflect(x::AbstractArray{F,N}, pad::NTuple{2,Int};
...
+    pad_symmetric(x, pad::Int; [dims])
+
+Pad the array `x` reflecting its values symmetrically across the border, i.e. the border values of `x` are present in the padding values, in contrast to [`pad_reflect`](@ref).
+
+`pad` can a tuple of integers `(l1, r1, ..., ln, rn)`
+of some length `2n` that specifies the left and right padding size
+for each of the dimensions in `dims`. If `dims` is not given, 
+it defaults to the first `n` dimensions.
+
+For integer `pad` input instead, it is applied on both sides
```
ToucheSir, line 291: I don't quite parse what "instead" means in this sentence. Maybe leftover word from an earlier edit?

ToucheSir: Oh, it was copied from another docstring. I think "if `pad` is an integer" would be clearer, but the boilerplate we have around dims handling could use some wordsmithing in general. @CarloLucibello WDYT?

nikopj: Yeah, it was copied from the pad_repeat docstring. I agree "if pad is an integer" would be clearer.

test/padding.jl
nikopj: Is it the indentation/alignment? Or structuring of the test?

src/padding.jl
ToucheSir: Feel free to change as you see fit if it makes things clearer :)

test/padding.jl
ToucheSir: The former. I see existing tests for other padding functions are also poorly formatted, but if we fix these ones we can copy the formatting here to those later.

nikopj:
@ToucheSir let me know if this formatting is any better.

ToucheSir (approved):
lndeed it is, thanks!

Issue #369, opened 2021-12-27
Bugged fallback in grouped Convs


SimonCoste:
Hi, 

I defined a *grouped* convolution in Flux using `C = Conv((1,1), 2=>2, groups=2)`. When I feed non-float arrays to this convolutional layer, eg with `C(rand(10,10,2,1))`, I first get a `Slow fallback` warning, and then an `AssertionError: DimensionMismatch`, see the stacktrace below.  

This error should not be here, and is very misleading since it is by no means a `DimensionMismatch` problem - the dimensions are ok - but it is apparently linked to the datatypes : indeed, accordingly to the warning, the error disappears when I use `C(rand(Float32, 10,10,2,1))`. 

Classical (non-grouped) convolutions do not display these kind of errors. 

```julia
julia> C = Conv((1,1), 2=>2, groups=2)
Conv((1, 1), 1 => 2)  # 4 parameters

julia> C(rand(10,10,2,1))
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
│   yT = Float64
│   T1 = Float64
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/P9BhZ/src/conv.jl:291
ERROR: AssertionError: DimensionMismatch("Data input channel count (2 vs. 2)")
Stacktrace:
  [1] check_dims(x::NTuple{5, Int64}, w::NTuple{5, Int64}, y::NTuple{5, Int64}, cdims::DenseConvDims{3, (1, 1, 1), 2, 2, 2, (1, 1, 1), (0, 0, 0, 0, 0, 0), (1, 1, 1), false})
    @ NNlib ~/.julia/packages/NNlib/P9BhZ/src/dim_helpers/DenseConvDims.jl:73
  [2] conv_direct!(y::Array{Float64, 5}, x::Array{Float64, 5}, w::Array{Float32, 5}, cdims::DenseConvDims{3, (1, 1, 1), 2, 2, 2, (1, 1, 1), (0, 0, 0, 0, 0, 0), (1, 1, 1), false}; alpha::Float64, beta::Bool)
    @ NNlib ~/.julia/packages/NNlib/P9BhZ/src/impl/conv_direct.jl:51
  [3] conv_direct!
    @ ~/.julia/packages/NNlib/P9BhZ/src/impl/conv_direct.jl:51 [inlined]
[...]

PR #468, opened 2023-02-01
Fix conv with groups when falling in direct backend

gabrielpreviato:
Fixing https://github.com/FluxML/NNlib.jl/issues/369.

When the convolution filter type is not the same as the input array, **conv** uses the _direct_ backend
Probably this backed was not updated when grouped convolution was added to **conv**, which led to some false warnings about dimension mismatches given that the **check_dimension** function changed as well.

This PR adds the same logic used in _im2col_ backend to the _direct_ backed, so grouped convolutions also works for the _direct_ backend.

### What has changed
Now for both _conv_ and _depthwiseconv_ (and their forwarding definitions), we have the functions definitions of both im2col and direct backend in a for-loop.

The idea is that future features or corrections will always be implemented for both backends avoiding the problem that happened in https://github.com/FluxML/NNlib.jl/issues/369.

For this, the loop needs a new element, where we had the frontend and backend, we now have the frontend, backend, and signature. This signature is used for the parametric types of the AbstractArray that are used for the functions since there are some type restrictions for the im2col backend.

### PR Checklist

- [X] Tests are added
- [X] Documentation, if applicable

### DIFF
--- a/src/conv.jl
+++ b/src/conv.jl
@@ -292,8 +292,8 @@ end
 for (front_name, backend) in (
     # This maps from public, front-facing name, to internal backend name
     :conv                   => :direct,
-    :∇conv_data             => :direct, 
-    :∇conv_filter           => :direct,
+    # :∇conv_data             => :direct, 
+    # :∇conv_filter           => :direct,
 )
 
     # We only define 3d conv primitives, we reshape lower down to get 1d and 2d convolution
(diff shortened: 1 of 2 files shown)


test/conv.jl
```diff
@@ -725,6 +725,17 @@ end
     @test size(conv(x, w; stride = (1, 2), pad = (2, 3), dilation = (2, 2), flipped = true)) == (12, 7, 16, 10)
 end
 
+# https://github.com/FluxML/NNlib.jl/issues/369
+@testset "conv_wrapper with groups - not equal types that trigger direct backend" begin
+    x = rand(Float32, 10, 10, 32, 8)
+    w = rand(Float64, 2, 2, 16, 4)
+    g = 2
+    @test size(conv(x, w; groups=g)) == (9, 9, 4, 8)
```
mcabbott, line 733: Perhaps better to test that the values are the same as the other implementation; this will also test the sizes:
```suggestion
    @test conv(x, w; groups=g) ≈ conv(x, Float32.(w); groups=g)
```

gabrielpreviato: Updated this to all tests

src/conv.jl
```diff
@@ -329,6 +329,62 @@ for (front_name, backend) in (
...
+
+    Threads.@sync for (wc, xc, yc) in zip(dw_cs, x_cs, dy_cs)
+        x = @view in1[ntuple(i -> i == 4 ? xc : Colon(), 5)...]
+        dy = @view in2[ntuple(i -> i == 4 ? yc : Colon(), 5)...]
+        dw = @view out[ntuple(i -> i == 5 ? yc : Colon(), 5)...]
+        Threads.@spawn ∇conv_filter_direct!(dw, x, dy, cdims2; kwargs...)
+    end
+
+    return out
+end
```
ToucheSir, line 386: This duplication with the code above is a little unfortunate. Is there any way we could reuse some of it? Maybe loop over `(:direct, :im2col)`?

gabrielpreviato: Yes, I thought the same. I was trying to implement something like this, but was having some problems with the function signatures, since for the im2col all types must be the same and from the list of possible types. I'm not used to Julia's metaprogramming so I'm learning a little to try come up with a solution.

gabrielpreviato: Yes, I was trying to implement this, but I'm not used to Julia's metaprogramming, so I'm doing some catch up to try coming up with a solution.

I was having some trouble with the function signatures, since for im2col all types must be the same and part of G.

I tried initially have a third element in the for loop tuple with the signature types, but then it gets messy when doing the where statement.

gabrielpreviato (commented):
@ToucheSir @mcabbott not the most elegant solution, but this is the idea that I had for merging both backends with the type declaration differences. Open for suggestions.

src/conv.jl
```diff
@@ -290,6 +334,62 @@ for front_name in (:conv, :∇conv_data, :∇conv_filter,
...
+
+#     Threads.@sync for (wc, xc, yc) in zip(dw_cs, x_cs, dy_cs)
+#         x = @view in1[ntuple(i -> i == 4 ? xc : Colon(), 5)...]
+#         dy = @view in2[ntuple(i -> i == 4 ? yc : Colon(), 5)...]
+#         dw = @view out[ntuple(i -> i == 5 ? yc : Colon(), 5)...]
+#         Threads.@spawn ∇conv_filter_direct!(dw, x, dy, cdims2; kwargs...)
+#     end
+
+#     return out
+# end
```
CarloLucibello, line 391: these comment methods can be removed, then this PR looks good to go

ToucheSir (approved):
Looks good and tests pass. Will merge tomorrow if there are no objections.

ToucheSir:
This was a nice enhancement, thanks @gabrielpreviato !

PR #471, opened 2023-02-13
Merge NNlibCUDA as subpackage

ToucheSir:
As part of the process of turning NNlibCUDA into a package extension, we need to first move the source into this repo while preserving its history. Without doing so, we [wouldn't be able to](https://github.com/JuliaRegistries/General/pull/73409) re-register it as a subpackage. To that end, I've followed the procedure described in https://github.com/JuliaPlots/Plots.jl/pull/4419#issuecomment-1266604686 to place it under `ext/NNlibCUDA`. Once this is done and registered, https://github.com/FluxML/NNlib.jl/pull/445 can be redone on top of this foundation.

### PR Checklist

- [N/A] Tests are added
- [N/A] Documentation, if applicable

### DIFF
--- a/ext/NNlibCUDA/.buildkite/pipeline.yml
+++ b/ext/NNlibCUDA/.buildkite/pipeline.yml
@@ -1,0 +1,20 @@
+steps:
+  - label: "GPU integration with julia v1.6"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.6" 
+      - JuliaCI/julia-test#v1: ~
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
+    timeout_in_minutes: 60
+
+  - label: "GPU integration with julia v1"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.8"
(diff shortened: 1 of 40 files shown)


ext/NNlibCUDA/.buildkite/pipeline.yml
```diff
@@ -0,0 +1,20 @@
...
+
+  - label: "GPU integration with julia v1"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.8"
+      - JuliaCI/julia-test#v1: ~
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
+    timeout_in_minutes: 60
```
CarloLucibello, line 20: not blocking for this PR, cleanup can be done later, but subpackages' `.buildkite` and `.github` dirs should be removed?

CarloLucibello (approved):

ext/NNlibCUDA/.buildkite/pipeline.yml
ToucheSir: Yes, absolutely.

PR #473, opened 2023-02-16
Allow regular convolution for AMDGPU

pxl-th:
MIOpen only [supports](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/convolution.html) cross-correlation as its convolution.

To perform regular convolution we can reverse the kernel along first two dimensions (for 2D conv).
Doing this every time during `NNlib.conv` calls is expensive, therefore on NNlib level a user will need to manually flip convolution kernel before passing it as an argument.
On the Flux level this will be handled automatically during `Flux.gpu(conv)` conversion once.

Flux's part is done in this PR: https://github.com/FluxML/Flux.jl/pull/2189

- Bump AMDGPU compat, since it has important performance optimizations.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
 AMDGPUExt = "AMDGPU"
 
 [compat]
-AMDGPU = "0.4.7"
+AMDGPU = "0.4.8"
 Adapt = "2, 3.2"
 ChainRulesCore = "1.13"
 Requires = "0.5, 1.0"
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -76,6 +76,11 @@ pad_zeros
 
 `Flux`'s `Conv` and `CrossCor` layers use `NNlib.DenseConvDims` and `NNlib.conv` internally. 
 
+!!! AMDGPU MIOpen only supports cross-correlation as its convolution algorithm (`flipkernel=true`).
+    Flip convolution kernel manually before passing it to `NNlib.conv`
+    to perform regular convolution (`flipkernel=false`): `reverse(w; dims=(1, 2))`.
+    `Flux` handles this automatically, this is only required for direct calls.
+
 ```@docs
 conv
 ConvDims
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -76,9 +76,11 @@ pad_zeros
 
 `Flux`'s `Conv` and `CrossCor` layers use `NNlib.DenseConvDims` and `NNlib.conv` internally. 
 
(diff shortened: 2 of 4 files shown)


CarloLucibello:
I think for consistency we should have
```julia
NNlib.conv(x, w, cdims) == NNlib.conv(ROCArray(x), ROCArray(wd), cdims)
```
so `conv` should flip the kernel internally. 

On the Flux side instead, when the `Conv` layer is moved to gpu the kernel is flipped and `conv(..., flipkernel=true)` is called in the forward (which is what happens for the `CrossCor` layer). 
I have to say that this solution is not ideal. What happens when you want to store an optimizer state? It should be flipped as well going back and forth.

pxl-th:
> I think for consistency we should have
> 
> ```julia
> NNlib.conv(x, w, cdims) == NNlib.conv(ROCArray(x), ROCArray(wd), cdims)
> ```
> 
> so `conv` should flip the kernel internally.

Should we then define custom `rrule` for `ROCArray` case?
Otherwise if someone is using `conv` in `gradient` directly, the gradient will be incorrect.

pxl-th:
> I think for consistency we should have
> 
> ```julia
> NNlib.conv(x, w, cdims) == NNlib.conv(ROCArray(x), ROCArray(wd), cdims)
> ```
> 
> so `conv` should flip the kernel internally.

We then also need to defined custom `rrule` for `ROCArray` that will flip gradient back if someone is using `conv` directly in the `gradient` call.

pxl-th:
> I think for consistency we should have
> 
> ```julia
> NNlib.conv(x, w, cdims) == NNlib.conv(ROCArray(x), ROCArray(w), cdims)
> ```
> 
> so `conv` should flip the kernel internally.

Agree, this is better. Done.
I also print a warning once to warn the user that this is sub-optimal.

> I have to say that this solution is not ideal. What happens when you want to store an optimizer state? It should be flipped as well going back and forth.

If the user continues training on the AMDGPU, then it should be alright.
We should probably specify these differences in the Flux.jl documentation.

pxl-th:
When this PR is merged, can we also tag a release?
So that I can specify compat in this PR: https://github.com/FluxML/Flux.jl/pull/2189

CarloLucibello (approved):

CarloLucibello:
buildkite CUDA CI failed with a "libLLVMExtra not defined" error. Before tagging a new release, I'm waiting for the CI results from the last commit to see if that was a fluke

ToucheSir:
To be safe, we should figure out our extension naming policy per https://github.com/JuliaMath/ChangesOfVariables.jl/pull/13 before tagging.

PR #477, opened 2023-02-21
change AMDGPUExt to NNlibAMDGPUExt

CarloLucibello:
Non CUDA-related part of #476


CarloLucibello:
@pxl-th is AMDGPU buildkite supposed to fail?

pxl-th:
> @pxl-th is AMDGPU buildkite supposed to fail?

Yes, it is currently broken. This specific reason I'd say is because of incorrect ROCm version, but I may be wrong. And there are also `HSA_OUT_OF_RESOURCES` errors on CI sometimes, Julian is looking into that.

PR #481, opened 2023-03-04
make NNlibCUDA an extension

CarloLucibello:
replacement for #476


CarloLucibello:
now that 1.9 is out we aready to do this, will file a new PR

PR #483, opened 2023-03-17
[NNlibAMDGPUExt] Load MIOpen module only if it is available

pxl-th:
Since MIOpen artifact is only available on Julia 1.9 it is not included in AMDGPU by default.
Therefore we need to load it only when it is available.

Otherwise, it causes crashes for users that do not have MIOpen installed.
For example in AMDGPU tests, where we compare CPU with AMDGPU implementations and don't rely on NNlib AMDGPU support.

- Migrate to separate `Project.toml` for tests.

### PR Checklist

- [ ] Tests are added


ToucheSir:
1.6 CUDA failure is probably spurious, but the AMDGPU one does not seem to be.

pxl-th:
Segfault in CI, because it uses global ROCm installation which is of incompatible version.
I've switched to using artifacts, but since MIOpen is not installed it skips tests.

ToucheSir:
Can we install the MIOpen_jll on CI?

.buildkite/pipeline.yml
```diff
@@ -76,18 +76,23 @@ steps:
         push!(conf[\"targets\"][\"test\"], \"AMDGPU\");
         open(io -> TOML.print(io, conf), \"Project.toml\", \"w\");
         """
+      - julia --project=. -e """
+        using Pkg;
+        Pkg.develop(\"AMDGPU\");
```
ToucheSir, line 81: Another reason to switch to test/Project.toml sooner than later?

ToucheSir: Or now, thanks!

pxl-th:
AMDGPU CI is probably down...

.buildkite/pipeline.yml
```diff
@@ -76,18 +76,23 @@ steps:
         push!(conf[\"targets\"][\"test\"], \"AMDGPU\");
```
ToucheSir, line 76: CI is breaking on `[targets]` no longer existing in Project.toml.

pxl-th:
Should we merge it? Even though AMDGPU CI is down, should be safe to do so.

ToucheSir (approved):
Sounds good. I thought the CUDA jobs were also failing but must've not checked recently enough.

PR #486, opened 2023-04-05
Use KernelAbstractions.jl for upsample kernels

pxl-th:
Ref #479.

- Migrate upsample kernels to KernelAbstractions.jl (implementation migrated from NNlibCUDA).
  Successfully tested on `CPU`, `ROCBackend`, `CUDABackend`.
  From the user perspective, there are no API changes (except `align_corners` option).
  Backend is selected automatically using `KernelAbstractions.get_backend(x)`.

- Introduce NNlib test suite, which accepts backend as an argument and runs tests for respective device.
  For now, it contains only upsample tests, but as we move other kernels to KernelAbstraction, it makes sense to move the rest of the tests to it as well.
  This way, we don't need to rewrite tests for each backend.

  When backend is
  - `CPU`, gradtest checks Zygote gradient against FiniteDifferences (as was before).
  - **not** `CPU`, gradtest checks agains `CPU`.
  This is because AMDGPU.jl fails with FiniteDifferences.jl.
  Still, fow now this is fine, because test suite for `CPU` runs first.
```julia
AssertionError: ROCArray only supports bits types
  Stacktrace:
    [1] ROCVector{Tuple{Vector{Float32}, FiniteDifferences.var"#Real_from_vec#20"}}(buf::AMDGPU.Runtime.Mem.Buffer, dims::Tuple{Int64}; offset::Int64)
    ...
```
- Add `align_corners` option (copied from NNlibCUDA.jl, but now supports CPU as well).
- Remove upsample from `ext/NNlibCUDA`.

**Side note:** this PR is based on #483 since it introduces separate `Project.toml` for tests. 

- [x] Add benchmarks comparing performance before & after.

### PR Checklist

- [x] Tests are added
- [ ] Documentation, if applicable

PR #487, opened 2023-04-09
Use KernelAbstractions.jl for gather/scatter kernels

pxl-th:
Depends on #486.
Successfully tested on `CPU`, `CUDABackend`, `ROCBackend`.

Since LLVM does not support certain atomic operations (fmin, fmax, fmul, fdiv) tests for them are disabled when using such backends, which is currently only `ROCBackend`:
https://llvm.org/docs/LangRef.html#atomicrmw-instruction

fmin/fmax requires LLVM 15+: https://reviews.llvm.org/D127041

### TODO

- [x] Add `@inbounds`.

### PR Checklist

- [x] Tests are added


pxl-th:
It may be that LLVM on Julia 1.6 is too old and we cann't support all of the ops that we currently support for CUDA.
On 1.9-rc2 it fails for `-` op.

As an alternative, we can retain (for now) NNlibCUDA scatter.

ToucheSir:
FWIW the copy of NNlibCUDA in this repo doesn't do anything at present, so we need to make sure the overloads in the standalone NNlibCUDA repo still work.

pxl-th:
> FWIW the copy of NNlibCUDA in this repo doesn't do anything at present, so we need to make sure the overloads in the standalone NNlibCUDA repo still work.

IIC, if NNlibCUDA in this repo is not registered, then tests at [`using NNlibCUDA`](https://github.com/pxl-th/NNlib.jl/blob/99f6ed4098df95e21fcea0a4797780daf1d32fb9/test/runtests.jl#L96) line should import the standalone repo.
And I've [disabled](https://github.com/pxl-th/NNlib.jl/blob/99f6ed4098df95e21fcea0a4797780daf1d32fb9/test/runtests.jl#L57) `Gather` and `Scatter` tests from test suite for CUDA, because we [call](https://github.com/pxl-th/NNlib.jl/blob/99f6ed4098df95e21fcea0a4797780daf1d32fb9/test/runtests.jl#L98) them from NNlibCUDA tests.

pxl-th:
Not sure what's with CUDA on 1.8...
Will investigate in a bit

pxl-th:
Probably was some kind of hiccup, now the CI passes.

ToucheSir (approved):
Anything left before merging?

pxl-th:
> Anything left before merging?

I think no, should be good to go

PR #488, opened 2023-04-13
Add dispatch path for FP16 batched mul

pxl-th:
rocBLAS supports FP16 batched gemm, but NNlib was dispatching to generic implementation.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/ext/NNlibAMDGPUExt/batched_mul.jl
+++ b/ext/NNlibAMDGPUExt/batched_mul.jl
@@ -1,0 +1,24 @@
+function _blas_at(x)
+    Base.stride(x, 1) == 1 && return x, 'N'
+    Base.stride(x, 2) == 1 && return batched_transpose(x), 'T'
+    throw(ArgumentError("""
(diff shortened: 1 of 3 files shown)


ext/NNlibAMDGPUExt/batched_mul.jl
```diff
@@ -0,0 +1,24 @@
...
+function NNlib._batched_mul!(
+    ::Type{AT}, C, A, B, α::Float16, β::Float16,
+) where AT <: ROCArray{Float16}
+    blasA, transA = _blas_at(A)
+    blasB, transB = _blas_at(B)
+    NNlib._batched_gemm!(AT, transA, transB, α, blasA, blasB, β, C)
+    C
+end
+
+function NNlib._batched_gemm!(
```
ToucheSir, line 20: Now that we have two GPU backends and both of them override this function, I wonder if it should become a more formal part of the API. Something to think about at a later date though.

ToucheSir (approved):

PR #489, opened 2023-04-19
Cast FP16 -> FP32 in softmax

pxl-th:
Even though MIOpen does support FP16 softmax, it still gives `NaN` values for big arrays (e.g. 4096x4096).
Therefore, cast `FP16 -> FP32`, perform softmax & cast back.

Eagerly free intermediate `FP32` arrays.
Override `rrule` for `FP16` softmax.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ lib/NNlibCUDA/Manifest.toml
 benchmark/Manifest.toml
 benchmark/*.json
 benchmark/report.md
+LocalPreferences.toml
--- a/LocalPreferences.toml
+++ b/LocalPreferences.toml
@@ -1,0 +1,2 @@
+[AMDGPU]
+use_artifacts = false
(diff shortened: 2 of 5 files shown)


CarloLucibello:
Does the problem occur with `logsoftmax` as well?

pxl-th:
> Does the problem occur with `logsoftmax` as well?

Actually, this may be some kind of a bug in MIOpen.
This behaviour is only observed on RX6700[XT], not on RX6800 or RX6900.
Also switching softmax mode from fast to accurate get's rid of `NaN` values.
And this is done on AMDGPU.jl side.

pxl-th:
Turns out, it was a synchronization issue with MIOpen.
Fixed in https://github.com/JuliaGPU/AMDGPU.jl/pull/415.

PR #491, opened 2023-05-06
Add (log)softmax benchmarks

ToucheSir:
In anticipation of some upcoming optimization work on these functions.

### PR Checklist

- [ ] ~~Tests are added~~
- [ ] ~~Documentation, if applicable~~

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,13 +1,45 @@
 using BenchmarkTools
 using NNlib
+using NNlib.ChainRulesCore: rrule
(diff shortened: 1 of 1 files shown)


mcabbott (approved):
No objection, although I'm not entirely sure how the files in benchmark/ are used.

ToucheSir:
You can see the benchmark output at https://buildkite.com/julialang/nnlib-dot-jl/builds/843. The table to check is the one under the first "Results" heading, which is currently empty as expected because there should be no source changes on this PR.

PR #497, opened 2023-06-15
move from bad thread-local to task-local

IanButterworth:
The previous `threadid`-based buffering may have been susceptible to error due to task migration.

### DIFF
--- a/src/impl/conv_im2col.jl
+++ b/src/impl/conv_im2col.jl
@@ -45,12 +45,12 @@ function conv_im2col!(
     N = channels_out(cdims)
     K = prod(kernel_size(cdims))*channels_in(cdims)
 
-    parts = collect(Iterators.partition(axes(x, 5), ceil(Int, size(x, 5) / ntasks)))
+    parts = Iterators.partition(axes(x, 5), ceil(Int, size(x, 5) / ntasks))
 
-    @sync for task_n in eachindex(parts)
+    @sync for (task_n, part) in enumerate(parts)
         Threads.@spawn begin
             col_slice = col_slice = view(col, :, :, task_n) # col_slice is a task-local workspace
-            for batch_idx in parts[task_n]
+            for batch_idx in part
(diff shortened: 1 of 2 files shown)


IanButterworth:
The nightly failure appears unrelated and is present on master

src/impl/conv_im2col.jl
```diff
@@ -44,16 +45,20 @@ function conv_im2col!(
...
-        # col_slice is a thread-local workspace
-        col_slice = view(col, :, :, threadid())
-
-        im2col!(col_slice, view(x, :, :, :, :, batch_idx), cdims)
-        GC.@preserve col_slice w y begin
-            col_ptr = pointer(col_slice)
-            w_ptr = pointer(w)
-            y_ptr = pointer(y, (batch_idx - 1)*M*N + 1)
-            gemm!(Val(false), Val(false), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
+    parts = collect(Iterators.partition(axes(x, 5), ceil(Int, size(x, 5) / ntasks)))
```
ToucheSir, line 48: How much effort would it take to avoid these `collect`s?

CarloLucibello: I don't think it's worth putting effort into avoiding allocations here, this is just partitioning indexes

ToucheSir: My question was more why line 50 below couldn't be:
```julia
@sync for (task_n, part) in enumerate(parts)
```

IanButterworth: I had tried `pairs` which didn't work with Iterators, but `enumerate` is good. Thanks. Updated

IanButterworth:
Ready for review, I think

ToucheSir (approved):

CarloLucibello:
it is better to squash-merge in the future, also for backporting convenience

IanButterworth:
I could have also tidied up with a force push

ToucheSir:
We can set the default merge action to be a squash merge. I've wanted to do that for a while.

PR #498, opened 2023-06-15
backports v0.8.21

CarloLucibello:
cc @IanButterworth


CarloLucibello:
@IanButterworth do you happen to know how to tag (with JuliaRegistrator) a backport release from a non-master branch?

IanButterworth:
I think you can do the same. A commit is what's tagged, which branch it's on isn't relevant to a tag, I believe

PR #499, opened 2023-06-15
port batchnorm rrule from Flux

CarloLucibello:
This is moved without changes from Flux. Now it is a good time to end this piracy.

### DIFF
--- a/src/normalization.jl
+++ b/src/normalization.jl
@@ -2,3 +2,13 @@
 function batchnorm end
 
 function ∇batchnorm end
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):
Is it time to bring back https://github.com/FluxML/NNlib.jl/pull/452?

CarloLucibello:
> Is it time to bring back https://github.com/FluxML/NNlib.jl/pull/452?

yes it would be good

Issue #510, opened 2022-04-15
conv on complex numbers


ToucheSir:
It appears PyTorch does this without using any custom kernels, perhaps we could explore something similar: https://github.com/pytorch/pytorch/blob/e177d2cc44de656d85c31b98d67a50e6689e0ed1/aten/src/ATen/native/Convolution.cpp#L665-L702

CarloLucibello:
this was solved by #517

PR #517, opened 2023-06-30
cudnn complex convolution via gauss trick

nikopj:
Addresses #510 using the same method as in Pytorch (Gauss's trick, complex conv via 3 real convs). 

### PR Checklist

- [x] Tests are added
- [x] Documentation: `conv` docstring updated

### DIFF
--- a/ext/NNlibCUDACUDNNExt/conv.jl
+++ b/ext/NNlibCUDACUDNNExt/conv.jl
@@ -56,6 +56,16 @@ function cudnnConvolutionDescriptor(cdims::DenseConvDims, x::DenseCuArray{T}, pa
                                Cint(NNlib.groupcount(cdims)))
 end
 
+@inline function _complex!(y::DenseCuArray{T1}, yr::DenseCuArray{T2}, yi::DenseCuArray{T2}; bias=zero(T1), alpha=one(T1), beta=zero(T1), σ=identity) where {T1 <: CUDNNComplexFloat, T2<:CUDNNFloat}
+    # if y is from similar(), it may have NaNs, and beta*NaN will propagate.
+    if beta != 0
+        @. y = σ(alpha*(yr + im*yi) + bias + beta*y)
+    else
+        @. y = σ(alpha*(yr + im*yi) + bias)
+    end
+    return y
+end
+
 function conv!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T}, cdims::DenseConvDims;
                alpha=1, beta=0, algo=-1) where T<:CUDNNFloat
     if cudnnversion() < v"6"
@@ -84,13 +94,25 @@ function conv!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T}, cdims
     a = conv!(similar(real(y)), xr, wr, cdims; algo=algo)
     b = conv!(similar(a), xi, wi, cdims; algo=algo)
     c = conv!(similar(a), xr + xi, wr + wi, cdims; algo=algo)
-    # if y is from similar(), it may have NaNs, and beta*NaN will propagate.
-    if beta != 0
-        @. y = alpha*((a - b) + im*(c - a - b)) + beta*y
-    else
-        @. y = alpha*((a - b) + im*(c - a - b))
-    end
-    return y
+    return _complex!(y, a - b, c - a - b; alpha=alpha, beta=beta)
+end
+
+# (xr + im*xi) * w = xr*w + im*(xi*w)
+function conv!(y::DenseCuArray{T1}, x::DenseCuArray{T1}, w::DenseCuArray{T2}, cdims::DenseConvDims;
+               alpha=1, beta=0, algo=-1) where {T1<:CUDNNComplexFloat, T2<:CUDNNFloat}
+    xr, xi = reim(x)
+    yr = conv!(similar(real(y)), xr, w, cdims; algo=algo)
+    yi = conv!(similar(yr), xi, w, cdims; algo=algo)
+    return _complex!(y, yr, yi; alpha=alpha, beta=beta)
+end
+
+# x * (wr + im*wi) = x*wr + im*(x*wi)
+function conv!(y::DenseCuArray{T1}, x::DenseCuArray{T2}, w::DenseCuArray{T1}, cdims::DenseConvDims;
+               alpha=1, beta=0, algo=-1) where {T1<:CUDNNComplexFloat, T2<:CUDNNFloat}
+    wr, wi = reim(w)
+    yr = conv!(similar(real(y)), x, wr, cdims; algo=algo)
+    yi = conv!(similar(yr), x, wi, cdims; algo=algo)
+    return _complex!(y, yr, yi; alpha=alpha, beta=beta)
 end
 
 function conv_bias_act!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T},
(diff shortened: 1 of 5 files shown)


ext/NNlibCUDACUDNNExt/conv.jl
```diff
@@ -67,6 +68,23 @@ function conv!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T}, cdims
...
+    a = conv!(similar(real(y)), xr, wr, cdims; algo=algo)
+    b = conv!(similar(a), xi, wi, cdims; algo=algo)
+    c = conv!(similar(a), xr + xi, wr + wi, cdims; algo=algo)
+    # if y is from similar(), it may have NaNs, and beta*NaN will propagate.
+    if beta != 0
+        @. y = alpha*((a - b) + im*(c - a - b)) + beta*y
+    else
+        @. y = alpha*((a - b) + im*(c - a - b))
+    end
+    any(isnan.(abs.(y))) && @warn "abs(y) isnan"
```
CarloLucibello, line 84: we don't do these kinds of checks anywhere. I think we can remove it, or at least put it in the beta != 0 if branch

ext/NNlibCUDACUDNNExt/conv.jl
```diff
@@ -67,6 +68,23 @@ function conv!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T}, cdims
     cudnnConvolutionForward!(y, w, x, d; alpha, beta, z=y)
 end
 
+function conv!(y::DenseCuArray{T}, x::DenseCuArray{T}, w::DenseCuArray{T}, cdims::DenseConvDims;
```
CarloLucibello, line 71: can you write a comment explaining the trick or add some references?

ext/NNlibCUDACUDNNExt/conv.jl
nikopj: my mistake, left by accident from debugging.

ext/NNlibCUDACUDNNExt/conv.jl
nikopj: np, added a comment above `conv!` explaining.

CarloLucibello:
Great, looks good to go. Since this is an important feature, maybe we should document it somewhere? the `conv` docstring?

nikopj:
> Great, looks good to go. Since this is an important feature, maybe we should document it somewhere? the `conv` docstring?

I noticed one dissimilarity between the cpu and cuda versions was that the cuda version could only handle when both arguments are complex, whereas the cpu version can handle a mix (though with errors for some of the pullback (∇) functions). This is understandable when the mix of real and complex doesn't make sense. E.g. for `dx = ∇conv_data(dy, w)`, we can have (`w` real, `dy` complex) but not (`w` complex, `dy` real), b/c `w` is complex => `y` complex for `y = conv(x, w)`.

I've now added a few more functions to allow the cuda conv functions to also handle a mix of real and complex inputs (in the same way that the cpu version can), and an additional sentence to the `conv` docstring mentioning that real and complex inputs are allowed. I tried to make the implementation more clean by writing an inline function `_complex!` which handles the beta if-statement.

There is one test that fails for the mixed case (`∇conv_filter!`, non-zero beta, padding=1, all input sizes), though I can't seem to figure out why. I've left it commented out for now with a note. Minus this one bug, I think things are ready to go.

ToucheSir:
I would make sure the case with non-zero beta isn't a correctness issue. We have some handling of that for single `conv` calls, but some of the fast paths (e.g. https://github.com/FluxML/NNlibCUDA.jl/pull/53) may no longer hold if one chains a bunch of them with delegated arguments like this PR does.

nikopj:
Ok, I think the beta bug is an error with the CPU version. Heres a MWE for the mixed real-complex `∇conv_filter!` with non-zero beta:
```julia
using NNlib, CUDA, cuDNN

for T=(Float64, ComplexF64), beta=(0,1), flip=(false, true)
    @show T, beta, flip
    x_cpu = fill(T(1), 2, 1, 1)
    w_cpu = T.([1; -1;;;])
    x_gpu = CuArray(x_cpu)
    w_gpu = CuArray(w_cpu)
    cdims = NNlib.DenseConvDims(x_cpu, w_cpu; flipkernel=flip)
    y_cpu = fill(T(1), 1, 1, 1)
    y_gpu = CuArray(y_cpu)

    w_cpu_2 = NNlib.∇conv_filter!(copy(w_cpu), real(x_cpu), y_cpu, cdims, alpha=T(1), beta=T(beta))
    w_gpu_2 = NNlib.∇conv_filter!(copy(w_gpu), real(x_gpu), y_gpu, cdims, alpha=T(1), beta=T(beta))
    @show w_cpu_2
    @show w_gpu_2
    @show w_cpu_2 ≈ Array(w_gpu_2)
end
```
Output:
```
(T, beta, flip) = (Float64, 0, false)
w_cpu_2 = [1.0; 1.0;;;]
w_gpu_2 = [1.0; 1.0;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (Float64, 0, true)
w_cpu_2 = [1.0; 1.0;;;]
w_gpu_2 = [1.0; 1.0;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (Float64, 1, false)
w_cpu_2 = [2.0; 0.0;;;]
w_gpu_2 = [2.0; 0.0;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (Float64, 1, true)
w_cpu_2 = [2.0; 0.0;;;]
w_gpu_2 = [2.0; 0.0;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (ComplexF64, 0, false)
w_cpu_2 = [1.0 + 0.0im; 1.0 + 0.0im;;;]
w_gpu_2 = [1.0 + 0.0im; 1.0 + 0.0im;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (ComplexF64, 0, true)
w_cpu_2 = [1.0 + 0.0im; 1.0 + 0.0im;;;]
w_gpu_2 = [1.0 + 0.0im; 1.0 + 0.0im;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (ComplexF64, 1, false)
w_cpu_2 = [2.0 + 0.0im; 0.0 + 0.0im;;;]
w_gpu_2 = [2.0 + 0.0im; 0.0 + 0.0im;;;]
w_cpu_2 ≈ Array(w_gpu_2) = true
(T, beta, flip) = (ComplexF64, 1, true)
w_cpu_2 = [0.0 + 0.0im; 2.0 + 0.0im;;;]
w_gpu_2 = [2.0 + 0.0im; 0.0 + 0.0im;;;]
w_cpu_2 ≈ Array(w_gpu_2) = false
```
So only the last case fails with `(T=ComplexF64, beat=1, flipkernel=true)`. As all of the arguments have zero imaginary part, we should expect the answer to be the same as for `T=Float64`, which points to the CPU version being incorrect.

ToucheSir (approved):
I noticed that we already have GPU-related notes in https://github.com/FluxML/NNlib.jl/blob/v0.9.3/docs/src/reference.md?plain=1#L79. It would be good to add what @CarloLucibello mentioned there too. Otherwise this LGTM.

ToucheSir:
It's great to have this feature now, thanks!

PR #521, opened 2023-07-07
Update AMDGPU compat to 0.5

pxl-th:
- Don't use artifacts in CI, since 0.5 requires ROCm 5.3+ for which there are no artifacts available.

### PR Checklist

- [ ] Tests are added


pxl-th:
Is it fine if I also tag a release?
Want to update compat for Flux next.

CarloLucibello:
Hi Anton, I forgot to mention that the policy in the org for the prominent packages is that those with write access should merge their own PRs, but only after they get an approval from another contributor. Anyways, this PR looks good to me, you can go on and tag a release

pxl-th:
Understood. Thanks!

PR #530, opened 2023-09-09
Bump AMDGPU compat to 0.6

pxl-th:
And drop 0.4

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ NNlibCUDAExt = "CUDA"
 NNlibCUDACUDNNExt = ["CUDA", "cuDNN"]
 
 [compat]
-AMDGPU = "0.4.8, 0.5"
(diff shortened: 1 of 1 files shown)


pxl-th:
Maybe we should add CUDA/AMDGPU/etc. to `Project.toml` only if `get(ENV, "NNLIB_TEST_CUDA", "false") == "true"` with `Pkg.add("CUDA")`?
This would avoid potential incompatibilities between the two packages.

CarloLucibello:
> Maybe we should add CUDA/AMDGPU/etc. to Project.toml only if get(ENV, "NNLIB_TEST_CUDA", "false") == "true" with Pkg.add("CUDA")?

you suggest we use `Pkg.add(...)` inside the if blocks in `runtests.jl` file? We can do that.  Let's merge #531 first though.

PR #531, opened 2023-09-09
Simplify runtests.jl

CarloLucibello:
and skip cpu tests when testing on gpu. 

The diff for `runtests.jl` is probably difficult to understand, but basically now we have a single block of tests for each backend. 

Another change is that now "AMDGPU" is part of the test environment and we don't call `Pkg.develop("AMDGPU")`.


ToucheSir:
If this was any FluxML repo other than NNlib I'd immediately be on board, but historically we've relied on Buildkite to catch threading-related bugs in the conv kernels and elsewhere because the buildkite runners have more threads to work with. Accomodating that would be a good idea.

ToucheSir:
On a side note, should we try moving to https://github.com/JuliaTesting/ReTestItems.jl for parallel test running and possible future VS Code integration? I noticed GPUCompiler recently did.

CarloLucibello:
> If this was any FluxML repo other than NNlib I'd immediately be on board, but historically we've relied on Buildkite to catch threading-related bugs in the conv kernels and elsewhere because the buildkite runners have more threads to work with. Accomodating that would be a good idea.

I've reinstated cpu tests on CUDA machine

ToucheSir (approved):

PR #535, opened 2023-09-20
bump CUDA compat

CarloLucibello:
Why CompatHelper is no longer active in our repos?


ToucheSir:
It was posting issues not too long ago. Did se ever update it to make it no longer expire after a certain amount of time?

PR #536, opened 2023-09-24
Add EnzymeRule for conv!/gather!/scatter!/dropout!/pool!

wsmoses:
Trying this out, hopefully to serve as a template for other functions.


wsmoses:
I wasn't sure where to put the test but here's an example:

```julia
using Enzyme
using NNlib

w = randn(Float32, 3, 3, 5, 7);
dw = zero(w)
loss(w, x) = sum(conv(x, w))
x = randn(Float32, (3, 3, 5, 8));

Enzyme.autodiff(Reverse, loss, Duplicated(w, dw), Const(x));

@show dw
```

wsmoses:
Another note, right now there is almost certainly more copying of inputs than necessary.

In the future we should add some sort of (type?) annotation to denote that some memory is immutable, or perhaps have a copy-on-write wrapper (as proposed by @gaurav-arya)

Project.toml
```diff
@@ -16,13 +17,15 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [weakdeps]
 AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
```
CarloLucibello, line 20: We could depend just on `EnzymeCore` right?

wsmoses:
Sure.

CarloLucibello:
Even if EnzimeCore is a lean dependency I think it is better to put all under a NNlibEnzymeCoreExt extension.

wsmoses:
So I tried that originally, but the problem with that is then your cuda extension would itself need an enzymecore extension.

If it matters, it also mirrors the way chainrulescore is used presently.

CarloLucibello:
> So I tried that originally, but the problem with that is then your cuda extension would itself need an enzymecore extension.

this can be addressed having also a NNlibEnzymeCoreCUDA extension.

> If it matters, it also mirrors the way chainrulescore is used presently.

true but chainrulescore is AD framework agnostic and quite widespread, enzymecore is not. It is a very lean dependency, so if making it an extension proves to be too tricky I wouldn't oppose making it a hard one but I think we should avoid it if we can.

wsmoses:
@CarloLucibello my attempt to use extensions (current code) fails. However, without the extension the previous commit (https://github.com/FluxML/NNlib.jl/pull/536/commits/f430a9b2feb884fea65d0eea49de341c05336cd2) succeeds.

If you see how to fix the extension, we can use that, if not we can use the prior commit.

ext/NNlibEnzymeCoreExt/NNlibEnzymeCoreExt.jl
```diff
@@ -0,0 +1,374 @@
...
+
+using NNlib
+import EnzymeCore
+using Random
+
+for (name, dataname, filtername) in ((typeof(NNlib.conv!), NNlib.∇conv_data!, NNlib.∇conv_filter!),
+                                     (typeof(NNlib.depthwiseconv!), NNlib.∇depthwiseconv_data!, NNlib.∇depthwiseconv_filter!) )
+    @eval begin
+
+function EnzymeCore.EnzymeRules.augmented_primal(config, func::EnzymeCore.Const{$name}, ::Type{RT},
```
CarloLucibello, line 11: this block should be indented twice.

We can shorten the signature a bit:
```suggestion
function EnzymeRules.augmented_primal(config, func::EnzymeCore.Const{$name}, ::Type{RT},
```
and `using EnzimeCore.EnzymeRules` at the beginning.

src/NNlib.jl
```diff
@@ -39,6 +39,12 @@ is_nnpack_available() = false
   end
 end
 
+@static if !isdefined(Base, :get_extension)
+@init @require EnzymeCore="f151be2c-9106-41f4-ab19-57ee4f262869" begin
+    include("../ext/NNlibEnzymeCore/NNlibEnzymeCoreExt.jl")
+end
+end
```
CarloLucibello, line 46: we only support julia 1.9+, this is not needed
```suggestion
```

src/conv.jl
```diff
@@ -47,7 +47,7 @@
 Apply convolution filter `w` to input `x`. `x` and `w` are 3d/4d/5d tensors
 in 1d/2d/3d convolutions respectively. `x` and `w` may have real or complex element types.
 """
-function conv(x, w::AbstractArray{T, N}; stride = 1, pad = 0, dilation = 1, flipped = false, groups = 1) where {T, N}
+@inline function conv(x, w::AbstractArray{T, N}; stride = 1, pad = 0, dilation = 1, flipped = false, groups = 1) where {T, N}
```
CarloLucibello, line 50: maybe better to factor out this change to another PR

src/enzyme.jl
```diff
@@ -0,0 +1,367 @@
+import EnzymeCore
```
CarloLucibello, line 1: the extension seems to be working so this file should be removed, right?

ext/NNlibEnzymeCoreExt/NNlibEnzymeCoreExt.jl
```diff
@@ -0,0 +1,374 @@
...
+        end
+    end
+
+    return (nothing, nothing, nothing, nothing)
+end
+
+end
+end
+
+function EnzymeCore.EnzymeRules.augmented_primal(config, func::EnzymeCore.Const{typeof(NNlib.gather!)}, ::Type{RT}, dst::OutType, src, idx::EnzymeCore.Const) where {OutType, RT}
```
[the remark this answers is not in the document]
ToucheSir: Are you basing these comments off some JuliaFormatter config? Because we don't have one for NNlib at present. We should have one, but I didn't want to hold up any PRs just because we don't.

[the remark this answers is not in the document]
CarloLucibello: no just generic style comments

ToucheSir (approved):
This was a tour de force of a PR, thanks @wsmoses!

Issue #538, opened 2023-09-26
grad_conv! computes wrong answer


wsmoses:
```
julia> dx, dy, cache_w = ([0.17;;; 0.19;;; 0.23], [0.11;;; 0.13;;; 0.15], [1.0;;;])
([0.17;;; 0.19;;; 0.23], [0.11;;; 0.13;;; 0.15], [1.0;;;])

julia> NNlib.∇conv_data!(dx, dy, cache_w, cdims; alpha=1.0, beta=1.0)
1×1×3 Array{Float64, 3}:
[:, :, 1] =
 0.11

[:, :, 2] =
 0.13

[:, :, 3] =
 0.15

```

This is wrong. grad_conv should do dx = alpha * dx + beta * grad. Here the correct result should be the original dx .+ dy, but instead just has dy.

wsmoses:
I traced the issue down to a flawed `function ∇conv_data_im2col!`

PR #539, opened 2023-09-26
Fix grad conv im2col

wsmoses:
Fixes https://github.com/FluxML/NNlib.jl/issues/538

Required for https://github.com/FluxML/NNlib.jl/pull/536

### DIFF
--- a/src/impl/conv_im2col.jl
+++ b/src/impl/conv_im2col.jl
@@ -162,7 +162,7 @@ function ∇conv_data_im2col!(
                     col_ptr = pointer(col_slice)
                     gemm!(Val(false), Val(true), M, N, K, alpha, dy_ptr, w_ptr, T(0), col_ptr)
                 end
-                col2im!(view(dx, :, :, :, :, batch_idx), col_slice, cdims)
+                col2im!(view(dx, :, :, :, :, batch_idx), col_slice, cdims, beta)
             end
         end
     end
@@ -276,7 +276,7 @@ end
 
 
 """
-    col2im!(x, col, cdims)
+    col2im!(x, col, cdims, beta=0)
 
 Does the inverse of `im2col!()`, converting `col` back into a 3d image, used for backward
 passes, transposed convolutions, etc...
@@ -287,7 +287,7 @@ desperate enough yet.
 """
 col2im!
 
-function col2im!(x::AbstractArray{T,4}, col::AbstractArray{T,2}, cdims::ConvDims) where T
+function col2im!(x::AbstractArray{T,4}, col::AbstractArray{T,2}, cdims::ConvDims, beta::T=T(0)) where T
     if spatial_dims(cdims) != 3
         throw(DimensionMismatch("col2im!() only accepts 3d convoluitional inputs"))
     end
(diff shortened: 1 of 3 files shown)


CarloLucibello:
I wonder how it is possible this path was never tested. Thank you for this quick fix! Can you add a test?

wsmoses:
I have enabled the tests marked broken that should've checked for this. Let's see if they pass now.

wsmoses:
Hm those fail for other reasons it seems, added an explicit distinct test.

src/impl/conv_im2col.jl
```diff
@@ -162,7 +162,7 @@ function ∇conv_data_im2col!(
                     col_ptr = pointer(col_slice)
                     gemm!(Val(false), Val(true), M, N, K, alpha, dy_ptr, w_ptr, T(0), col_ptr)
```
ToucheSir, line 163: While trying to determine how to simplify https://github.com/FluxML/NNlib.jl/pull/539/files#diff-92d81e5ae578721d9c48a17e46ad5e36ca65e58cd5f786e5789820f16d9bc9e7R306-R312, it occurred to me that we probably shouldn't be handling this in `col2im!`. `im2col!` doesn't handle `beta`, and `∇conv_filter_im2col!` does via https://github.com/FluxML/NNlib.jl/blob/v0.9.6/src/impl/conv_im2col.jl#L110. Hence, perhaps this would work instead:
```suggestion
                    gemm!(Val(false), Val(true), M, N, K, alpha, dy_ptr, w_ptr, beta, col_ptr)
```

wsmoses: Unfortunately not, because the col2im sets dx to 0. Also the gemm! here is not the correct matrix to apply beta to.

ToucheSir: That makes sense. But then the way it's done now also doesn't work, because there's no `beta * grad` as mentioned in https://github.com/FluxML/NNlib.jl/issues/538. Of the 3 branches in https://github.com/FluxML/NNlib.jl/pull/539/files#diff-92d81e5ae578721d9c48a17e46ad5e36ca65e58cd5f786e5789820f16d9bc9e7R306-R312, only the first `fill!(...)` does anything?

wsmoses: So the beta == 0 case, is optimized to fill since thats saying x = alpha * newvals. the beta = 1 case, can just be optimized to x += alpha * newvals, so we don't need any additional work since x already contains x * old(x). The alternate case just requires x *= beta, since then the final x will be x = beta * old(x) + newvals.

ToucheSir: So the alternate case should be `x .*= beta` then?

wsmoses: Yes, looks like I missed that =

wsmoses: Fixed (and added test) for that third case

wsmoses: And with that fix it actually does look like a bunch of previously failing tests now pass.

ToucheSir (approved):
This LGTM now, the remaining tests are actually fixed broken tests on the CUDA integration side. Will merge tonight unless you have more changes planned.

wsmoses:
nope, am all done, go for it!

PR #540, opened 2023-09-26
Possible way to implement a LoopVectorization extension for conv2d & meanpool2d

jonas208:
Some tests have shown that standard for-loop implementations for conv using LoopVectorization (LV) can be faster than the standard NNlib's im2col + gemm approach. There has already been some talks about a possible LV extension on Slack (and in one of the ML community meetings). For example, meanpooling can also be implemented using LV, but the small boost in performance is probably negligible in large networks (pooling is not as performance critical as conv I think). For conv I found out that:
- Acceleration is usually greatest when the inputs have a large spatial size and few channels. 
- Using stride > 1, dilation > 1 or groups > 1 can slow down things a bit. (there are multiple impls for specialized cases)
- The current state of LV `∇conv_filter!` isn't really faster than the 
original implementation in some situations, for that reason, I left it out for the moment. (The same applies to the backwardpass of meanpool.)
- Important: For inputs with very many channels (and a relatively small spatial size), the LV-Impl is often slower!
- 
Tests were run on a Ryzen 9 5900X (on Windows 10).

I also wrote some tests (not sure if they fit properly in the NNlib's test strategy).
I hope the code is somewhat usable and helpful.


jonas208:
Was this script run as well? https://github.com/FluxML/FluxMLBenchmarks.jl/blob/main/benchmark/benchmark/nnlib/conv.jl
I‘m not sure, but it seems that only [flux] [mlp] was run (or is conv part of this group as well?).

ToucheSir:
It should've run, I've asked folks more familiar with the benchmarking why it didn't. For now, your local benchmarking results are fine.

jonas208:
Just updated `/test/ext_loopvectorizationruntests.jl` to run some benchmarks (on CI probably with just one thread I guess), to make some of the test results more precise with actual numbers, these are some local results from the modified script:
```
without LoopVectorization
  153.013 ms (171 allocations: 651.41 MiB)
  169.605 ms (173 allocations: 369.11 MiB)
  83.041 ms (170 allocations: 319.79 MiB)
  89.114 ms (172 allocations: 199.92 MiB)
  63.385 ms (431 allocations: 325.76 MiB)
  86.451 ms (431 allocations: 202.99 MiB)
  8.591 ms (13 allocations: 35.61 MiB)
  2.694 ms (13 allocations: 1.81 MiB)
with LoopVectorization
  42.939 ms (41 allocations: 319.05 MiB)
  18.518 ms (2164 allocations: 36.86 MiB)
  37.051 ms (40 allocations: 156.63 MiB)
  14.204 ms (2096 allocations: 36.87 MiB)
  65.591 ms (40 allocations: 196.93 MiB)
  85.284 ms (9990 allocations: 111.64 MiB)
  5.962 ms (4 allocations: 35.61 MiB)
  810.100 μs (4 allocations: 1.80 MiB)
Test Summary:         | Pass  Total     Time
Convolution & Pooling |    3      3  3m48.3s
```

jonas208:
Has anyone an idea why the results on some CI devices are correct but sometimes totally wrong? Some weird LV behavior (the version should be the same on all devices)?
@ToucheSir Maybe @chriselrod has an idea?

chriselrod:
Note that `@tturbo` and `@turbo` may really benefit from knowing the sizes at compile time.
I had thought that this was part of the `ConvDims` type, but that appears not to be the case.
Probably not worth compiling a bunch of extra convolution types, but could be worth a try.

Note also that `SimpleChains.jl` defines a lot of layer types, as well as a `ChainRules.rrule` to make `SimpleChains` objects themselves usable as layers in other networks. Defining the `ChainRules` makes them Zygote-compatible (i.e., usable by Zygote, they do not support using Zygote).
https://github.com/PumasAI/SimpleChains.jl/blob/87a5b400d798bd38dfde8ab93cce959a2b7d3ce3/test/runtests.jl#L116-L121

> Maybe @chriselrod has an idea?

Bugs?

ext/NNlibLoopVectorizationExt/conv.jl
```diff
@@ -0,0 +1,181 @@
...
+        # println("forward: second specialized case for better performance")
+
+        @tturbo for index_batch in 1:batches
+            for out_channel in 1:out_channels, y_out in 1:output_height, x_out in 1:output_width
+                m = y_out + (y_stride - 1) * (y_out - 1)
+                n = x_out + (x_stride - 1) * (x_out - 1)
+                value = zero(T)
+                for in_channel in 1:in_channels, y_w in 1:weight_height, x_w in 1:weight_width
+                    y_in = m + (y_w - 1) * y_dilation
+                    x_in = n + (x_w - 1) * x_dilation
```
chriselrod, line 62: I'm not sure if it makes a difference, but LV's handling of indices written within an index expression might be better.
It has separate code for building indices by parsing index expressions vs parsing the loop body.
I'd have to look at the details to see the degree to which they're different. In theory, they should do the same thing.

LV would definitely benefit from the special case of `x_dilation` being `1`.
Might be worth branching over (also, special case `-1` or some other some other common small factors).

jonas208:
@chriselrod Thank you for the quick reply! Just added a minimal example in the `runtests.jl` script: https://github.com/jonas208/NNlib.jl/blob/lv-ext2/test/ext_loopvectorization/minimal_test.jl
To be really sure that the implementation is not just unequal to NNlib because of some errors in the logic, the minimal example compares two exact implementations with the difference that the first uses `@turbo` and the other doesn't. 

1):
```julia
for index_batch in 1:current_batch_size # normally @threads is used here
        @turbo for out_channel in 1:out_channels, y_out in 1:output_height, x_out in 1:output_width
            for in_channel in 1:in_channels, y_w in 1:weight_height, x_w in 1:weight_width
                input_gradient[x_out + x_w - 1, y_out + y_w - 1, in_channel, index_batch] += weight[x_w, y_w, in_channel, out_channel] * output_gradient[x_out, y_out, out_channel, index_batch]
            end
        end
end
```
2):
```julia
for index_batch in 1:current_batch_size # normally @threads is used here
        for out_channel in 1:out_channels, y_out in 1:output_height, x_out in 1:output_width
            for in_channel in 1:in_channels, y_w in 1:weight_height, x_w in 1:weight_width
                input_gradient[x_out + x_w - 1, y_out + y_w - 1, in_channel, index_batch] += weight[x_w, y_w, in_channel, out_channel] * output_gradient[x_out, y_out, out_channel, index_batch]
            end
        end
end
```
On my local machine, I never got unequal results (using Julia 1.9.3 on Windows 10, running on an AMD Ryzen 9 5900X). It seems that the results are also always correct on CI devices running MacOS (maybe the MacOS device always runs on the same CPU).
On CI Windows or CI Ubuntu, results are sometimes wrong and sometimes not. I also added `cpuinfo()` from `CpuId.jl` to get some information about the used CPUs. After rerunning the tests, maybe some regularities in the issues can be found.

jonas208:
Some logs:
Worked on
```
| Brand              | Intel(R) Xeon(R) CPU E5-2673 v4 @ 2.30GHz               |
| Vendor             | :Intel                                                  |
| Architecture       | :Broadwell                                              |
```
Worked on
```
| Brand              | Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz                  |
| Vendor             | :Intel                                                     |
| Architecture       | :IvyBridge                                                 |
```
Failed on
```
| Brand              | Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz           |
| Vendor             | :Intel                                                  |
| Architecture       | :UnknownIntel                                           |
```
Failed on
```
| Brand              | Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz          |
| Vendor             | :Intel                                                  |
| Architecture       | :Skylake                                                |
```

jonas208:
What are your thoughts about having errors sometimes and sometimes not? It is acceptable to have something like this merged in NNlib? If the potential "error" can't be solved, my best suggestion is to always run a self check before overloading the NNlib's methods and print a warning saying the implementation might not work on the current device when the self check fails.
@mcabbott @ToucheSir @CarloLucibello (or other NNlib maintainers)

chriselrod:
Oh, no, it is expected that LoopVectorization.jl does not support loops like that. 
LoopModels, the rewrite, will. 

But it will take some time before that is ready.

I wrote about it here: https://spmd.org/posts/orthogonalizing_loops/

I applied the necessary rotation manually (and used masks because LV requires rectangular loops) here:
https://github.com/PumasAI/SimpleChains.jl/blob/87a5b400d798bd38dfde8ab93cce959a2b7d3ce3/src/conv.jl#L718C1-L718C1
I suggest following this style of workaround.

jonas208:
Oh right, thank you, I must have apparently no longer thought about it. Thanks for linking the article!

jonas208:
After a lot of benchmarking and lots of smaller attempts to raise the performance a bit more, I mainly found out that:
- for inference with a batch size of 1, LV `conv` clearly outperforms `im2col` (up to 10 times faster in `MobileNetv3` or `EfficientNetv2`)
- for inference with batch size of approx. >16, LV `conv` usually brings a slight performance advantage in bigger models (around 25%)
- LV `conv` allocates (much) less memory (in particular when no padding is needed)

I also implemented LV `conv` without the need to do padding naively (using `calc_padding_regions` and bounds checks), but LV failed to process these loops sometimes (especially then when stride>1 or dilation>1). 

- for LV `∇conv_data!`, only the most specialized case is left over, stride>1 or dilation>1 made the index calculation too complicated
- in some cases, LV `∇conv_data!` is slower, sometimes it's faster, so I'm not quite sure if we should keep this then?
- `im2col` struggles with `EfficientNet(v2)` (maybe because of the high group counts and 1x1 convs)
- setting the Julia threads to 1 can sometimes speed up `im2col` convolution when groups are used (threading over groups can result in an insane amount of allocations, e.g. >1M for `im2col` against a few thousands with LV per inference run in `EfficientNet`)

For benchmarking with whole models, I wrote [this script](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark.jl) which can compare inference/backward times for a customizable set of models and batch sizes.
The output is a `.csv` table with the timings and the acceleration factor. To compare Flux with PyTorch, [this python script](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark_result_pytorch.csv) can be useful.
These were my results (benchmarked on a Ryzen 9 5900X):
- [im2col against LV](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark_result_julia.csv)
- [PyTorch Timings](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark_result_pytorch.csv)

It looks like only the CI on buildkite is run on this PR anymore.

This PR should now be ready for reviewing :)
Btw, does anyone have further ideas to improve performance (e.g. memory blocking)?
@maintainers, e.g. @ToucheSir @mcabbott @CarloLucibello

ToucheSir:
If it's not too much trouble, could you try setting `LinearAlgebra.BLAS.set_num_threads(1)` and seeing if that affects the im2col results? Presently, we way oversubscribe on CPU because both BLAS and NNlib/Julia are managing separate threadpools.

jonas208:
Sure, I got [these results](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark_result_julia_BLAS.set_num_threads(1).csv).

jonas208:
Just had a quick look at the results on buildkite on the AMD GPU server. I saw massive (very unusual) differences between LV (75ms) and im2col (15+seconds) (never occurred otherwise, e.g. not even on the CUDA GPU server). Does anyone know what could be reasons for this? https://buildkite.com/julialang/nnlib-dot-jl/builds/1116#018b13d7-e2fe-469f-8b5a-ac06f25b0bee/286-568
Starting from the 4th row at "with LV", every two rows measurements occur again, which used im2col (because from then on stride>1, means that im2col is used again for backward pass). The last 8 lines are timings for the activations.

mcabbott:
A quick bug report from trying this out on this example: https://github.com/FluxML/Flux.jl/issues/2350 fails:

<details>

```julia
(jl_fgGTBI) pkg> add https://github.com/jonas208/NNlib.jl#lv-ext2  ^C

julia> using NNlib, LoopVectorization

julia> function train_loop(model, optimizer, train_loader, test_loader; epochs=5)
           for epoch ∈ 1:epochs
               iter = tqdm(train_loader)
               total = 0
               corrects = 0
               for (X, Y) ∈ iter
                   grads = Flux.gradient(model) do m
                       predicted = m(X)
                       ignore() do 
                           b_size = size(X)[end]
                           corrects += sum(onecold(predicted, 0:9) .== onecold(Y, 0:9))
                           total += b_size
                       end
                       logitcrossentropy(predicted, Y)
                   end
                   optimizer, model = Flux.Optimise.update!(optimizer, model, grads[1])
                   set_postfix(iter, accuracy=corrects / total)
               end

               val_accuracy = accuracy(model, test_loader)
               @info "Epoch $epoch/5 | Accuracy : $val_accuracy"
           end
       end
train_loop (generic function with 1 method)

julia> train_loop(model, optimizer, TRAIN_LOADER, TEST_LOADER)
0.0%┣                                                                  ┫ 0/469 [00:00<00:00, -0s/it]
ERROR: conversion to pointer not defined for Array{Float32, 4}
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] unsafe_convert(::Type{Ptr{Float32}}, a::Array{Float32, 4})
    @ Base ./pointer.jl:67
  [3] unsafe_convert
    @ ~/.julia/packages/OffsetArrays/0MOrf/src/OffsetArrays.jl:464 [inlined]
  [4] pointer(x::OffsetArrays.OffsetArray{Float32, 4, Array{Float32, 4}})
    @ Base ./abstractarray.jl:1253
  [5] memory_reference
    @ ~/.julia/packages/LayoutPointers/qGkBo/src/stridedpointers.jl:21 [inlined]
  [6] memory_reference
    @ ~/.julia/packages/LayoutPointers/qGkBo/src/stridedpointers.jl:18 [inlined]
  [7] stridedpointer_preserve
    @ ~/.julia/packages/LayoutPointers/qGkBo/src/stridedpointers.jl:100 [inlined]
  [8] ∇conv_data!(input_gradient::Array{…}, output_gradient::Array{…}, weight::Array{…}, cdims::DenseConvDims{…})
    @ NNlibLoopVectorizationExt ~/.julia/packages/NNlib/HV0kw/ext/NNlibLoopVectorizationExt/conv.jl:160
  [9] #∇conv_data#241
    @ ~/.julia/packages/NNlib/HV0kw/src/conv.jl:99 [inlined]
 [10] ∇conv_data
    @ ~/.julia/packages/NNlib/HV0kw/src/conv.jl:95 [inlined]
 [11] #380
    @ ~/.julia/packages/NNlib/HV0kw/src/conv.jl:350 [inlined]
 [12] unthunk
    @ ~/.julia/packages/ChainRulesCore/7MWx2/src/tangent_types/thunks.jl:204 [inlined]
 [13] wrap_chainrules_output
    @ ~/.julia/packages/Zygote/YYT6v/src/compiler/chainrules.jl:110 [inlined]
 [14] map
    @ ./tuple.jl:283 [inlined]
 [15] map
    @ ./tuple.jl:284 [inlined]
 [16] wrap_chainrules_output
    @ ~/.julia/packages/Zygote/YYT6v/src/compiler/chainrules.jl:111 [inlined]
 [17] ZBack
    @ ~/.julia/packages/Zygote/YYT6v/src/compiler/chainrules.jl:211 [inlined]
 [18] Conv
    @ ~/.julia/packages/Flux/u7QSl/src/layers/conv.jl:202 [inlined]
 [19] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Array{Float32, 4})
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface2.jl:0
 [20] macro expansion
    @ ~/.julia/packages/Flux/u7QSl/src/layers/basic.jl:53 [inlined]
 [21] _applychain
    @ ~/.julia/packages/Flux/u7QSl/src/layers/basic.jl:53 [inlined]
 [22] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Matrix{Float32})
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface2.jl:0
 [23] Chain
    @ ~/.julia/packages/Flux/u7QSl/src/layers/basic.jl:51 [inlined]
 [24] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Matrix{Float32})
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface2.jl:0
 [25] #25
    @ ./REPL[53]:8 [inlined]
 [26] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface2.jl:0
 [27] (::Zygote.var"#75#76"{Zygote.Pullback{Tuple{…}, Tuple{…}}})(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface.jl:45
 [28] gradient(f::Function, args::Chain{Tuple{…}})
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/compiler/interface.jl:97
 [29] train_loop(model::Chain{…}, optimizer::@NamedTuple{…}, train_loader::DataLoader{…}, test_loader::DataLoader{…}; epochs::Int64)
    @ Main ./REPL[53]:7
 [30] train_loop(model::Chain{…}, optimizer::@NamedTuple{…}, train_loader::DataLoader{…}, test_loader::DataLoader{…})
    @ Main ./REPL[53]:1
 [31] top-level scope
    @ REPL[54]:1
Some type information was truncated. Use `show(err)` to see complete types.

(jl_fgGTBI) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_fgGTBI/Project.toml`
  [bdcacae8] LoopVectorization v0.12.165
  [872c559c] NNlib v0.9.7 `https://github.com/jonas208/NNlib.jl#lv-ext2`
  
julia> versioninfo()
Julia Version 1.11.0-DEV.773
Commit 855cd5662d* (2023-10-29 15:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.6.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
  Threads: 5 on 4 virtual cores
Environment:
  JULIA_NUM_THREADS = 4
```

</details>

chriselrod:
That's because you're on Julia master.

mcabbott:
Ah thanks, I didn't know. Piratically defining `Base.unsafe_convert(::Type{Ptr{T}}, x::Array{T}) where T = pointer(x)`, it now runs, and `@time train_loop(model, optimizer, TRAIN_LOADER, TEST_LOADER; epochs=1)` reports 48s with this PR, 34s without. At least on this machine, M1 with apple blas, haven't investigated further.

One minor comment is that `Flux.nil` numbers as used by `@autosize` seem to trigger warnings (on the first run):
```julia
julia> Flux.@autosize (28,28,1,1) Chain(MeanPool((2,2)), Flux.flatten, Dense(_ => 10))
┌ Warning: #= /Users/me/.julia/packages/NNlib/HV0kw/ext/NNlibLoopVectorizationExt/pooling.jl:57 =#:
│ `LoopVectorization.check_args` on your inputs failed; running fallback `@inbounds @fastmath` loop instead.
│ Use `warn_check_args=false`, e.g. `@turbo warn_check_args=false ...`, to disable this warning.
└ @ NNlibLoopVectorizationExt ~/.julia/packages/LoopVectorization/xHfLl/src/condense_loopset.jl:1148
Chain(
  MeanPool((2, 2)),
  Flux.flatten,
  Dense(196 => 10),                     # 1_970 parameters
) 
```

chriselrod:
That message is `@warn maxlog=1`, so it should generally only display once.
It means (as it says) `@turbo` is not actually being used.

mcabbott:
Yes I understand about the logging. But I think it's a sign that e.g. `NNlib.meanpool!(output::Array{T,4}, ...) where {T<:Real}` [here](https://github.com/FluxML/NNlib.jl/pull/540/files#diff-1576b0d493c7d2fbeaf676f2a63b8c5cbdeadfe9b333f6fc6e7fd8ee1abe58d3R12) is too wide a signature. Shouldn't it restrict `T` to the list of types which will work? Leave the existing `::AbstractArray{<:Any, 4}` method as the fallback.

jonas208:
@mcabbott (and maybe @chriselrod if you're interested) Sorry for the late reply! I just saw it again now.
> Shouldn't it restrict T to the list of types which will work? Leave the existing ::AbstractArray{<:Any, 4} method as the fallback.

Yes, thanks, this would be better. I will probably change this in the next days.
> reports 48s with this PR, 34s without

Due to the ratio between the large number of channels compared to the low spatial size of 28x28, it's somewhat expected that this example isn't optimal for the current LV implementation. On the other hand, no stride/dilation/groups are used, so the most optimized path is chosen. So in general, it shouldn't be significantly slower than im2col in such cases (probably at least on x86 AVX2). 

On my Ryzen processor, the LV impl took ~33s and the default impl took ~38s with 24 Julia threads and 12 BLAS-threads or ~30s with 24 Julia threads and 1 BLAS-thread. I made the experience that turning off BLAS-threads can often speed up the default im2col approach, but sometimes it can also make it worse (especially the backward passes). 
At the moment, however, I can't think of any other ways to further optimize the LV method. Are there any? If possible, it may be worth trying to include [oneDNN](https://github.com/oneapi-src/oneDNN) as an accelerator package (similar to NNPACK, but better and more stable).

Apart from that, if you have time, it would be interesting to see some more benchmark results on the M1, for example the tests in the [benchmark script](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark.jl) which produces [this csv. table](https://github.com/jonas208/NNlib.jl/blob/lv-ext2/benchmark_result_julia.csv) to compare LV against im2col.

maxfreu:
Just a side note: The pytorch timings are probably sub-optimal. One would use `torch.compile()` nowadays and `with torch.no_grad()` for inference, to actually suppress the computation of gradients. None that `torch.compile()` then requires to benchmark the second run, like in julia. So ultimately, the bar is even higher...

PR #542, opened 2023-09-30
Don't run CPU tests on AMDGPU CI machines

pxl-th:
If it is alright, I'd like to disable CPU tests on AMD GPU CI machines.
AFAIK, there's only one CI machine right now available for AMD GPUs.
Having it run tests that take ~1 hour increases CI queue significantly.

I've enabled CPU tests on CUDA CI machines, which we have a lot more.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,8 +14,10 @@ steps:
       queue: "juliagpu"
       cuda: "*"
     env:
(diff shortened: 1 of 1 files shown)


pxl-th:
Also, @wsmoses, there are a lot of failures with Enzyme tests and AMDGPU.

CarloLucibello (approved):

PR #543, opened 2023-10-02
Allow non-Int64 indices in scatter

pxl-th:
- Allow non-`Int64` indices in scatter by promoting them to `Int64` inside the kernel.
This is especially useful when performing reduction over big arrays into a few bins.
Indices can be of `Int8` type, saving a lot of space.
- Fix typo in `Test_Enzyme`. I guess the intention was to disallow any Julia 1.10 version.
```julia
julia> v"1.10.0-beta2" <= v"1.10"
true

julia> v"1.10.0-beta2" <= v"1.10-"
false
```

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/scatter.jl
+++ b/src/scatter.jl
@@ -108,7 +108,7 @@ end
 
 @kernel function _scatter!(op::OP, dst, src, idxs) where OP
     i = @index(Global)
-    @inbounds idx = Tuple(idxs[i])
+    @inbounds idx = Tuple(_atomix_convert_i64(idxs[i]))
     @inbounds Atomix.modify!(Atomix.IndexableRef(dst, idx), op, src[i])
     # FIXME `@atomic` macro silently fails to perform atomic op below
     # @atomic dst[idx...] = op(dst[idx...], src[i])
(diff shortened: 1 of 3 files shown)


src/scatter.jl
```diff
@@ -119,14 +119,20 @@ end
...
     # dim_i = Tuple(dim_ids[k + 1])
     # idx = idxs[j + 1]
     # @atomic dst[dim_i..., idx...] = op(dst[dim_i..., idx...], src[i])
 end
 
+# Allow non-Int64 indices by converting them to Int64 when index eltype <: Integer.
+# All other index types (tuples, cartesian indices) must be in Int64 already.
+@inline _atomix_convert_i64(x::Int64) = x
+@inline _atomix_convert_i64(x::Integer) = Int(x)
+@inline _atomix_convert_i64(x) = x
```
ToucheSir, line 134: What does the `_atomix_` here mean?

pxl-th: Also, I forgot to mention that casting to i64 is only needed for atomics, where we take `Atomix.pointer(::Atomix.IndexableRef)` which for only with `Int` type.

ToucheSir (approved):
LGTM, merge when ready

PR #550, opened 2023-11-08
Set compatibilities for standard packages

bicycle1885:
This is required by the [automatic merging guidelines](https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/).


Project.toml
```diff
@@ -35,6 +35,9 @@ ChainRulesCore = "1.13"
 EnzymeCore = "0.5, 0.6"
 GPUArraysCore = "0.1"
 KernelAbstractions = "0.9.2"
+LinearAlgebra = "<0.0.1, 1"
```
pxl-th, line 38: Why `<0.0.1`? Shouldn't it be just `1` for all base libraries?

pxl-th: Or just `1.9`, since this is the minimal version of Julia NNlib supports.

ToucheSir: I thought the `<0.0.1` was required to make all Julia versions before 1.10 work? For the upper bound I'm not sure, the guidance on this was never very clear.

pxl-th: Ah, I see

Project.toml
```diff
@@ -35,6 +35,9 @@ ChainRulesCore = "1.13"
...
+Pkg = "<0.0.1, 1"
+# Random = "<0.0.1, 1"
```
pxl-th, line 40:
```suggestion
Random = "<0.0.1, 1"
```

Project.toml
bicycle1885: `<0.0.1` is from my attempts to make the CI happy, which failed because of some incompatibility issue of Random.jl (at least the error message suggested that). I first thought my change in this PR caused that error, but I discovered it was there before the change! So, it might be okay to remove it if the package seems to work well in your environment.

pxl-th (approved):

pxl-th:
@ToucheSir this looks good to go. Maybe let's merge it and give another try to version registration?

Issue #552, opened 2023-12-14
Differentiable Image Rotatation with CUDA


roflmaostc:
### Motivation and description

Hi!
I was working on [DiffImageRotation.jl](https://github.com/roflmaostc/DiffImageRotation.jl) to rotate images differentiably with KernelAbstractions.jl.

So far only bilinear interpolation is supported but nearest neighbors would be straightforward too.

Should we try efforts to integrate this into this NNlib.jl?

Best,

Felix

CC: @maxfreu (he suggested it on discourse)

maxfreu:
Disclaimer: I haven't worked on the NNlib codebase for quite a while, so the following assumes that there is no rotation code present as of now. Searching for "rotat" in the repo didn't yield any results, so...

Your implementation is nice and small, just one file. So the source could go directly into src, KA is already a dependency of NNlib. It only needs some tweaks to include a channel dimension, which should be easy to add. In NNlib, the convention is to have reverse pytorch dim ordering, so WHCN. It makes sense to treat the channel dimension differently, depending which device the kernel is run on. On the CPU one would treat the batch and channel dimension as one and parallelize over that. On the GPU, it's faster to parallelize spatially, so across width and height, looping over the channel dimension in the kernel. You can look at the code for bilinear upsampling, @pxl-th has done magnificent work on that [here](https://github.com/FluxML/NNlib.jl/blob/7f7007904bb1a60da7a470f06710c0bf2cdeb0c9/src/upsample.jl#L441-L512). Oh and we should think forward a bit to possibly include nearest neighbor rotation in the future. So either name the function imrotate_bilinear or include a switch `method=:bilinear` or so.

Regarding the tests: That should also be pretty straight-forward to include here, as you already have them, which is great! NNlib offers some tools to test your reverse adjoints against the forward diff results. The [upsampling tests](https://github.com/FluxML/NNlib.jl/blob/master/test/upsample.jl) are an example how you can write a test that either runs on CPU or GPU and uses automatic tests against forwarddiff via `gradtest_fn` (which is just gradtest or gputest from test_utils.jl, depending on the device it's being run on).

Just fork NNlib, copy your files over, open a draft PR and we guide you through! :)

roflmaostc:
So restricting to 4D array should be good?
I also implemented `method=:nearest` right now.

maxfreu:
[...]

PR #553, opened 2023-12-17
Add image rotation

roflmaostc:
Hi,

based on  [DiffImageRotation.jl](https://github.com/roflmaostc/DiffImageRotation.jl) and as discussed in #552 I tried to add image rotation into NNlib.jl.

Some comments:
I always assumed a 4D array. In principle, the results match the ones from  [ImageTransformations.jl](https://github.com/JuliaImages/ImageTransformations.jl) closely. But, the boundary handling is different and sometimes rounding artifacts occur.
I compared this a little but I think there is space for interpretation how to do it. There is some code left, to change the boundary handling at the cost of ~20% performance.

### PR Checklist

- [x] Tests are added
- [x] Documentation -> comes later

## Benchmarks
@maxfreu mentioned something about different parallelization schemes for CPU vs GPU.

Benchmarking between my implementation and the parallelized elementwise application shows that the timings are roughly equal. So not sure if we should make the code much more verbose and complicated? But please share your experience with me! Especially, if we can squeeze out more performance out of CUDA, that would be great! But on CPU we already beat existing code:

Some benchmarking:
```julia
julia> CUDA.versioninfo()
CUDA runtime 11.8, artifact installation
CUDA driver 11.4
NVIDIA driver 470.223.2

CUDA libraries: 
- CUBLAS: 11.11.3
- CURAND: 10.3.0
- CUFFT: 10.9.0
- CUSOLVER: 11.4.1
- CUSPARSE: 11.7.5
- CUPTI: 18.0.0
- NVML: 11.0.0+470.223.2

Julia packages: 
- CUDA: 5.1.1
- CUDA_Driver_jll: 0.7.0+0
- CUDA_Runtime_jll: 0.10.1+0

Toolchain:
- Julia: 1.9.4
- LLVM: 14.0.6

1 device:
  0: NVIDIA GeForce RTX 3060 (sm_86, 11.369 GiB / 11.749 GiB available)


julia> versioninfo()
Julia Version 1.9.4
Commit 8e5136fa297 (2023-11-14 08:46 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × AMD Ryzen 5 5600X 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver3)
  Threads: 12 on 12 virtual cores
Environment:
  JULIA_NUM_THREADS = 12
```

### Comparison to ImageTransformations:

```julia
# multithreaded
julia> arr_2D = randn((2000, 3000));

julia> @btime NNlib.imrotate(reshape($arr_2D, 2000, 3000, 1, 1), deg2rad(13));
  24.449 ms (153 allocations: 45.79 MiB)

julia> @btime ImageTransformations.imrotate($arr_2D, deg2rad(13));
  81.811 ms (3 allocations: 67.55 MiB)

# julia -t 1
julia> @btime NNlib.imrotate(reshape($arr_2D, 2000, 3000, 1, 1), deg2rad(13));
  45.397 ms (29 allocations: 45.78 MiB)

julia> @btime ImageTransformations.imrotate($arr_2D, deg2rad(13));
  95.694 ms (3 allocations: 67.55 MiB)
```

```julia
julia> using CUDA, NNLib, BenchmarkTools, Tullio

julia> arr = randn(Float32, (256, 256, 3, 32));

julia> arrc = CuArray(arr);

julia> @btime NNlib.imrotate($arr, deg2rad(37));
  8.020 ms (147 allocations: 24.01 MiB)

julia> @btime @tullio out := sin.($arr[i, j, c, b]);
  8.358 ms (196 allocations: 10.77 KiB)

julia> @btime CUDA.@sync NNlib.imrotate($arrc, deg2rad(37));
  510.106 μs (99 allocations: 5.39 KiB)

julia> @btime CUDA.@sync sin.($arrc);
  291.288 μs (37 allocations: 2.36 KiB)


```

Would be great if someone reviews this and we can merge it! Not urgent for me, since DiffImageRotation.jl is registered and I anyway use this currently.

### DIFF
--- a/src/rotation.jl
+++ b/src/rotation.jl
@@ -1,7 +1,7 @@
 """
     _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
 
-this rotates the coordinates and either applies round(nearest neighbour)
+This rotates the coordinates and either applies round(nearest neighbour)
 or floor for :bilinear interpolation)
 """
 @inline function _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
--- a/src/rotation.jl
+++ b/src/rotation.jl
@@ -178,7 +178,7 @@ function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, midpoint=size(
         throw(ArgumentError("No interpolation method such as $method"))
     end
     kernel!(out, arr, sinθ, cosθ, midpoint, size(arr, 1), size(arr, 2),
-            ndrange=(size(arr, 1), size(arr, 2), size(arr, 3), size(arr, 4)))
+            ndrange=size(arr))
 	return out
 end
 
@@ -214,7 +214,7 @@ function ∇imrotate(dy, arr::AbstractArray{T, 4}, θ; method=:bilinear,
     end
     # don't pass arr but dy! 
     kernel!(out, dy, sinθ, cosθ, midpoint, size(arr, 1), size(arr, 2),
-            ndrange=(size(arr, 1), size(arr, 2), size(arr, 3), size(arr, 4)))
+            ndrange=size(arr))
     return out
 end
 
--- a/src/rotation.jl
+++ b/src/rotation.jl
@@ -159,10 +159,10 @@ julia> NNlib.imrotate(arr, deg2rad(45), method=:nearest)
  0.0  0.0  0.0
 ```
 """
-function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1) where T
-    @assert (T <: Integer && method==:nearest || !(T <: Integer)) "If the array has an Int eltype, only method=:nearest is supported"
-    @assert typeof(rotation_center) <: Tuple "rotation_center keyword has to be a tuple"
-    
+function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center::Tuple=size(arr) .÷ 2 .+ 1) where T
+    if (T <: Integer && method==:nearest || !(T <: Integer)) == false
+        throw(ArgumentError("If the array has an Int eltype, only method=:nearest is supported"))
+    end
     # prepare out, the sin and cos and type of rotation_center
     sinθ, cosθ, rotation_center, out = _prepare_imrotate(arr, θ, rotation_center) 
     # such as 0°, 90°, 180°, 270° and only if the rotation_center is suitable
@@ -197,7 +197,7 @@ Adjoint for `imrotate`. Gradient only with respect to `arr` and not `θ`.
 * `rotation_center=size(arr) .÷ 2 .+ 1` rotates around a real center pixel for even and odd sized arrays
 """
 function ∇imrotate(dy, arr::AbstractArray{T, 4}, θ; method=:bilinear, 
-                                               rotation_center=size(arr) .÷ 2 .+ 1) where T
+                                               rotation_center::Tuple=size(arr) .÷ 2 .+ 1) where T
     
     sinθ, cosθ, rotation_center, out = _prepare_imrotate(arr, θ, rotation_center) 
     # for the adjoint, the trivial rotations go in the other direction!
--- a/src/rotation.jl
+++ b/src/rotation.jl
@@ -1,10 +1,10 @@
 """
-    rotate_coordinates(sinθ, cosθ, i, j, rotation_center, round_or_floor)
+    _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
 
 this rotates the coordinates and either applies round(nearest neighbour)
 or floor for :bilinear interpolation)
 """
-@inline function rotate_coordinates(sinθ, cosθ, i, j, rotation_center, round_or_floor)
+@inline function _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
     y = i - rotation_center[1]
     x = j - rotation_center[2]
     yrot = cosθ * y - sinθ * x + rotation_center[1]
@@ -18,11 +18,11 @@ end
 
 
 """
-   bilinear_helper(yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int) 
+   _bilinear_helper(yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int) 
 
 Some helper variables
 """
-@inline function bilinear_helper(yrot, xrot, yrot_f, xrot_f)
+@inline function _bilinear_helper(yrot, xrot, yrot_f, xrot_f)
     xdiff = (xrot - xrot_f)
     xdiff_1minus = 1 - xdiff
     ydiff = (yrot - yrot_f)
@@ -96,8 +96,10 @@ end
 """
     imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1)
 
-Rotates a matrix around the center pixel `rotation_center`. `rotation_center` is defined such that there
-is a real center pixel for even and odd values which is rotated around.
+Rotates an array in the first two dimensions around the center pixel `rotation_center`. 
+`rotation_center` is defined such that there is a real center pixel for even and odd values which is rotated around.
+For an even sized array of size `(4,4)` this would be `(3,3)`, for an odd array of size `(3,3)` this would be `(2,2)`
+
 The angle `θ` is interpreted in radians.
 
 The adjoint is defined with ChainRulesCore.jl. This method also runs with CUDA (and in principle all KernelAbstractions.jl supported backends).
@@ -223,7 +225,7 @@ end
     i, j, c, b = @index(Global, NTuple)
 
     r(x...) = round(x..., RoundNearestTiesAway)
-    _, _, _, _, yrot_int, xrot_int = rotate_coordinates(sinθ, cosθ, i, j, rotation_center, r) 
+    _, _, _, _, yrot_int, xrot_int = _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, r) 
     if 1 ≤ yrot_int ≤ imax && 1 ≤ xrot_int ≤ jmax
         @inbounds out[i, j, c, b] = arr[yrot_int, xrot_int, c, b]
     end
@@ -233,11 +235,11 @@ end
 @kernel function imrotate_kernel_bilinear!(out, arr, sinθ, cosθ, rotation_center, imax, jmax)
     i, j, c, b = @index(Global, NTuple)
     
-    yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int = rotate_coordinates(sinθ, cosθ, i, j, rotation_center, floor) 
+    yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int = _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, floor) 
     if 1 ≤ yrot_int ≤ imax - 1 && 1 ≤ xrot_int ≤ jmax - 1 
 
         ydiff, ydiff_1minus, xdiff, xdiff_1minus = 
-            bilinear_helper(yrot, xrot, yrot_f, xrot_f)
+            _bilinear_helper(yrot, xrot, yrot_f, xrot_f)
         @inbounds out[i, j, c, b] = 
             (   xdiff_1minus    * ydiff_1minus  * arr[yrot_int      , xrot_int      , c, b]
              +  xdiff_1minus    * ydiff         * arr[yrot_int + 1  , xrot_int      , c, b]
@@ -251,7 +253,7 @@ end
     i, j, c, b = @index(Global, NTuple)
 
     r(x...) = round(x..., RoundNearestTiesAway)
-    _, _, _, _, yrot_int, xrot_int = rotate_coordinates(sinθ, cosθ, i, j, rotation_center, r) 
+    _, _, _, _, yrot_int, xrot_int = _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, r) 
     if 1 ≤ yrot_int ≤ imax && 1 ≤ xrot_int ≤ jmax 
         Atomix.@atomic out[yrot_int, xrot_int, c, b] += arr[i, j, c, b]
     end
@@ -261,11 +263,11 @@ end
 @kernel function ∇imrotate_kernel_bilinear!(out, arr, sinθ, cosθ, rotation_center, imax, jmax)
     i, j, c, b = @index(Global, NTuple)
 
-    yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int = rotate_coordinates(sinθ, cosθ, i, j, rotation_center, floor) 
+    yrot, xrot, yrot_f, xrot_f, yrot_int, xrot_int = _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, floor) 
     if 1 ≤ yrot_int ≤ imax - 1 && 1 ≤ xrot_int ≤ jmax - 1
         o = arr[i, j, c, b]
         ydiff, ydiff_1minus, xdiff, xdiff_1minus = 
-            bilinear_helper(yrot, xrot, yrot_f, xrot_f)
+            _bilinear_helper(yrot, xrot, yrot_f, xrot_f)
         Atomix.@atomic out[yrot_int     ,   xrot_int    , c, b]  += xdiff_1minus    * ydiff_1minus * o
         Atomix.@atomic out[yrot_int + 1 ,   xrot_int    , c, b]  += xdiff_1minus    * ydiff      * o
         Atomix.@atomic out[yrot_int     ,   xrot_int + 1, c, b]  += xdiff           * ydiff_1minus * o
--- a/src/rotation.jl
+++ b/src/rotation.jl
@@ -1,14 +1,14 @@
 """
-    rotate_coordinates(sinθ, cosθ, i, j, midpoint, round_or_floor)
+    rotate_coordinates(sinθ, cosθ, i, j, rotation_center, round_or_floor)
 
 this rotates the coordinates and either applies round(nearest neighbour)
 or floor for :bilinear interpolation)
 """
-@inline function rotate_coordinates(sinθ, cosθ, i, j, midpoint, round_or_floor)
-    y = i - midpoint[1]
-    x = j - midpoint[2]
-    yrot = cosθ * y - sinθ * x + midpoint[1]
-    xrot = sinθ * y + cosθ * x + midpoint[2]
+@inline function rotate_coordinates(sinθ, cosθ, i, j, rotation_center, round_or_floor)
+    y = i - rotation_center[1]
+    x = j - rotation_center[2]
+    yrot = cosθ * y - sinθ * x + rotation_center[1]
+    xrot = sinθ * y + cosθ * x + rotation_center[2]
     yrot_f = round_or_floor(yrot)
     xrot_f = round_or_floor(xrot)
     yrot_int = round_or_floor(Int, yrot)
@@ -33,42 +33,42 @@ end
 
 
 """
-    _prepare_imrotate(arr, θ, midpoint)
+    _prepare_imrotate(arr, θ, rotation_center)
 
 Prepate `sin` and `cos`, creates the output array and converts type
-of `midpoint` if required.
+of `rotation_center` if required.
 """
-function _prepare_imrotate(arr::AbstractArray{T}, θ, midpoint) where T
+function _prepare_imrotate(arr::AbstractArray{T}, θ, rotation_center) where T
     # needed for rotation matrix
     θ = mod(real(T)(θ), real(T)(2π))
-    midpoint = real(T).(midpoint)
+    rotation_center = real(T).(rotation_center)
     sinθ, cosθ = sincos(real(T)(θ)) 
     out = similar(arr)
     fill!(out, 0)
-    return sinθ, cosθ, midpoint, out
+    return sinθ, cosθ, rotation_center, out
 end
 
 
 """
-    _check_trivial_rotations!(out, arr, θ, midpoint) 
+    _check_trivial_rotations!(out, arr, θ, rotation_center) 
 
-When `θ = 0 || π /2 || π || 3/2 || π` and if `midpoint` 
+When `θ = 0 || π /2 || π || 3/2 || π` and if `rotation_center` 
 is in the middle of the array.
-For an even array of size 4, the midpoint would need to be 2.5.
-For an odd array of size 5, the midpoint would need to be 3.
+For an even array of size 4, the rotation_center would need to be 2.5.
+For an odd array of size 5, the rotation_center would need to be 3.
 
 In those cases, rotations are trivial just by reversing or swapping some axes.
 """
-function _check_trivial_rotations!(out, arr, θ, midpoint; adjoint=false)
+function _check_trivial_rotations!(out, arr, θ, rotation_center; adjoint=false)
     if iszero(θ)
         out .= arr
         return true 
     end
     # check for special cases where rotations are trivial
     if (iseven(size(arr, 1)) && iseven(size(arr, 2)) && 
-        midpoint[1] ≈ size(arr, 1) ÷ 2 + 0.5 && midpoint[2] ≈ size(arr, 2) ÷ 2 + 0.5) ||
+        rotation_center[1] ≈ size(arr, 1) ÷ 2 + 0.5 && rotation_center[2] ≈ size(arr, 2) ÷ 2 + 0.5) ||
         (isodd(size(arr, 1)) && isodd(size(arr, 2)) && 
-        (midpoint[1] == size(arr, 1) ÷ 2 + 1 && midpoint[1] == size(arr, 2) ÷ 2 + 1))
+        (rotation_center[1] == size(arr, 1) ÷ 2 + 1 && rotation_center[1] == size(arr, 2) ÷ 2 + 1))
         if θ ≈ π / 2 
             if adjoint == false
                 out .= reverse(PermutedDimsArray(arr, (2, 1, 3, 4)), dims=(2,))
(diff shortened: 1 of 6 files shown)


ToucheSir:
Thanks! I'll leave it to @maxfreu to comment on the algorithmic piece, but the biggest outstanding addition here would be making the test suite device-agnostic so it can be tested on GPU CI too. Have a look at the upsample tests for how this is done if you haven't already.

roflmaostc:
ImageTransformations.jl doesn't work on GPUs but we can probably wrap it in `Arrays`.

But yeah, I see and can fix it!

test/rotation.jl
```diff
@@ -0,0 +1,97 @@
+
```
maxfreu, line 1: As Brian said, it would be good to make the tests device-aware. I suggest the following:
- Wrap your testsets in a function rotation_testsuite, taking the device (CPU, GPU) as argument
- All your current forward-pass tests against other libraries can be kept.
- If not too much work, modify the gradient tests to use `gradtest`, which should make them nice and short.
- If the device is CPU, run the tests as they are
- If the device is GPU, check the GPU results + gradients against the CPU results, which can be done automatically via gputest.

So essentially: 1) Check CPU against other libs 2) check CPU grads against ForwardDiff 3) check GPU forward and backward against CPU, if GPU is present. You don't have to copy the upsample tests 1:1, but you can draw inspiration from there.

Having good tests is essential to avoid headache later on, so stay strong! :D

roflmaostc:
Thanks for the comments.
I plan to incorporate it in a couple of weeks. 
But for Julia 1.10 we need to wait for this [PR to update ImageTransformations.jl](https://github.com/JuliaImages/ImageTransformations.jl/pull/168) because otherwise some dependencies were held back at non 1.10 compatible versions.

src/rotation.jl
```diff
@@ -0,0 +1,174 @@
...
+    # needed for rotation matrix
+    θ = mod(real(T)(θ), real(T)(2π))
+
+    if iszero(θ)
+        out .= arr
+        return out
+    end
+
+    # check for special cases where rotations are trivial
+    if midpoint[1] ≈ size(arr, 1) ÷ 2 + 0.5 && midpoint[2] ≈ size(arr, 2) ÷ 2 + 0.5
```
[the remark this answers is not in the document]
roflmaostc: Yeah, the trivial rotations can be only avoided if the array is even and the midpoint is in the middle of two pixels.

Or in the odd case, when it is in the middle. 
So yeah, I forgot the odd case for comparison here :)

src/rotation.jl
```diff
@@ -0,0 +1,174 @@
...
+        ydiff_diff = 1 - ydiff
+        # in case we hit the boundary stripe, then we need to avoid out of bounds access
+        Δi = 1#yrot_int != imax
+        Δj = 1#xrot_int != jmax
+      
+        # we need to avoid that we access arr[0]
+        # in rare cases the rounding clips off values on the left and top border
+        # we still try to access them with this extra comparison
+        Δi_min = 0#yrot_int == 0
+        Δj_min = 0#xrot_int == 0
```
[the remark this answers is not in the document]
roflmaostc: Now i match with ImageTransformations.

roflmaostc:
There is an issue with FillArrays.jl which I don't know how to solve. It seems to fail catch the backend with FillArrays.
FillArrays is not an explicit dependency otherwise I would do a dispatch in case of that.

```
Test gradients: Error During Test at /home/fxw/.julia/dev/NNlib.jl/test/rotation.jl:78
  StackOverflowError:
  Stacktrace:
   [1] get_backend(A::FillArrays.Fill{Float32, 4, NTuple{4, Base.OneTo{Int64}}}) (repeats 79984 times)
     @ KernelAbstractions ~/.julia/packages/KernelAbstractions/GCOhX/src/KernelAbstractions.jl:466
```

src/rotation.jl
```diff
@@ -0,0 +1,250 @@
...
+end
+
+
+# is this rrule good? 
+# no @thunk and @unthunk
+function ChainRulesCore.rrule(::typeof(imrotate), array::AbstractArray{T}, θ; 
+                              method=:bilinear, midpoint=size(array) .÷ 2 .+ 1) where T
+    res = imrotate(array, θ; method, midpoint)
+    function pb_rotate(dy)
+        ad = ∇imrotate(unthunk(collect(dy)), θ; method, midpoint)
```
roflmaostc, line 245: @maxfreu this `collect` is just to check the tests.
But this does not work for CUDA. I Introduced it to avoid this StackOverflow bug mit FillARrays.

ToucheSir:
The StackOverflow is https://github.com/JuliaGPU/KernelAbstractions.jl/issues/366 manifesting again. https://github.com/JuliaGPU/KernelAbstractions.jl/blob/96b89f994e08b48b0a3142d7db4b1c97be054b41/src/KernelAbstractions.jl#L466 should probably check if `parent(A) === A` before recursing. I would recommending bumping that issue or opening a new one.

roflmaostc:
But would KA know that the desired array is a CuArray and not an Array? FillArrays don't have hardware backends, right? So I guess we would need to fill a real array with the FillArray?

src/rotation.jl
```diff
@@ -0,0 +1,250 @@
...
```
ToucheSir, line 245: I don't see why the desired array would be a CuArray or a normal Array. It should be possible to pass a `FillArray` to a KA kernel as long as its eltype `isbits` and the kernel doesn't try to mutate it. The problem here seems to be that `pb_rotate` assumes `dy` is not a fancy array type. A fix would be to pass the input `array` to `∇imrotate` as well. Then `_prepare_imrotate` can call similar on `array` to get an output of the correct type instead of relying on `similar(dy)` being the correct type.
```suggestion
    function pb_rotate(dy)
        ad = ∇imrotate(unthunk(dy), arr, θ; method, midpoint)
```

roflmaostc:
Thanks, I fixed it and added documentation!

roflmaostc:
Some GPU test fail but also in parts I didn't touch. Maybe I should relax the `.\approx` tests with a relative error?

roflmaostc:
Another question regardin KA.jl: should we include a `KernelAbstractions.synchronize(backend)`?

src/rotation.jl
```diff
@@ -0,0 +1,286 @@
...
+    if method == :bilinear
+        kernel! = ∇imrotate_kernel_bilinear!(backend)
+    elseif method == :nearest
+        kernel! = ∇imrotate_kernel_nearest!(backend)
+    else 
+        throw(ArgumentError("No interpolation method such as $method"))
+    end
+    # don't pass arr but dy! 
+    kernel!(out, dy, sinθ, cosθ, midpoint, size(arr, 1), size(arr, 2),
+            ndrange=(size(arr, 1), size(arr, 2), size(arr, 3), size(arr, 4)))
```
maxfreu, line 217: `ndrange=size(arr)`? arr is 4D anyway.

test/rotation.jl
```diff
@@ -0,0 +1,90 @@
+function upsample_testsuite(Backend)
```
CarloLucibello, line 1:
```suggestion
function rotation_testsuite(Backend)
```
?

roflmaostc:
I included your comments and relaxed some tests since they failed on CUDA.

test/rotation.jl
```diff
@@ -0,0 +1,91 @@
...
+        for method in [:nearest, :bilinear]
+            for angle in deg2rad.([0, 35, 45, 90, 135, 170, 180, 270, 360])
+                res1 = cpu(NNlib.imrotate(arr, angle; method))
+                res3 = cpu(NNlib.imrotate(arr2, angle; method))
+                if method == :nearest
+                    res2 = ImageTransformations.imrotate(cpu(arr)[:, :, 1, 1], angle, axes(arr)[1:2], method=Constant(), fillvalue=0)
+                elseif method == :bilinear
+                    res2 = ImageTransformations.imrotate(cpu(arr)[:, :, 1, 1], angle, axes(arr)[1:2], fillvalue=0)
+                end
+                @test all(.≈(1 .+ res1[:, :, :, :], 1 .+ res2[:, :], rtol=rtol))
```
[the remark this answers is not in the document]
roflmaostc: `arr` is on the `device` right?
And looking here, seems like CUDA fails?
https://buildkite.com/julialang/nnlib-dot-jl/builds/1190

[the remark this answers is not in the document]
maxfreu: It could be that it simply fails due to automatic conversion to float32. Does it work with a bit relaxed tolerance?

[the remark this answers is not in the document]
roflmaostc: the tolerance is already at 1f-3 in relative which is actually quite low for Float32. And I test everything in Float32 already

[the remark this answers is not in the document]
maxfreu: Oh yeah, sorry, just saw your change. Hmmm. Have you checked the output for the failing cases visually?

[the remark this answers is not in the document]
roflmaostc: How can I do that? It's only on the buildkite? Locally it's fine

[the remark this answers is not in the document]
maxfreu: Yes, I meant locally. I also just tested locally and it works. Maybe it's time for a good old print statement to see, what is happening in buildkite...

test/rotation.jl
```diff
@@ -0,0 +1,91 @@
...
+        @testset "Simple test" begin
+            arr = device(zeros((6, 6, 1, 1))); 
+            arr[3:4, 4, 1, 1] .= 1;
+            @test all(cpu(NNlib.imrotate(arr, deg2rad(45))) .≈ [0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.29289321881345254 0.585786437626905 0.0; 0.0 0.0 0.08578643762690495 1.0 0.2928932188134524 0.0; 0.0 0.0 0.0 0.08578643762690495 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0])
+        end
+    end
+
+
+    @testset "Compare with ImageTransformations" begin
+        arr = device(zeros(T, (51, 51, 1, 1)))
```
maxfreu, line 18: I suggest to rename arr to x1, arr2 to x2, res1 to y1, res3 to y2 and res2 to y_true to make input, output and true output clearer.

test/rotation.jl
```diff
@@ -0,0 +1,91 @@
...
+        end
+        
+        arr = device(zeros(T, (52, 52, 1, 1)))
+        arr[15:40, 15:40, :, :] .= device(1 .+ randn((26, 26)))
+        
+        arr2 = device(zeros(T, (52, 52, 5, 1)))
+        arr2[15:40, 15:40, :, 1] .= device(arr[15:40, 15:40, :, 1])
+
+        for method in [:nearest, :bilinear]
+            for angle in deg2rad.([0, 35,  90, 170, 180, 270, 360])
```
maxfreu, line 47: 135 is missing compared to above tests

test/rotation.jl
```diff
@@ -0,0 +1,91 @@
...
+                end
+                @test all(1 .+ res1[:, :, :, :] .≈ 1 .+ res2[:, :])
+                @test all(1 .+ res1[:, :, :, :] .≈ 1 .+ res3[:, :, :, :])
+                @test all(1 .+ res1[:, :, :, :] .≈ 1 .+ res3[:, :, :, :])
+            end
+        end
+
+    end
+
+    @testset "Compare for plausibilty" begin
```
maxfreu, line 63: Another good plausibility check would be to generate a random array and to rotate it forth and back with the same angle. But that just as a side note.

roflmaostc: yeah this is only approximately true, bilinear interpolation washes out the features a little.

test/rotation.jl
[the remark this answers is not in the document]
roflmaostc: did so, check tomorrow the output then

test/rotation.jl
```diff
@@ -0,0 +1,80 @@
...
+                                res1 = cpu(NNlib.imrotate(arr1, angle; method))
+                                res2 = cpu(NNlib.imrotate(arr2, angle; method))
+                            end
+                            if method == :nearest
+                                res_IT = ImageTransformations.imrotate(cpu(arr1)[:, :, 1, 1], angle, axes(arr1)[1:2], method=Constant(), fillvalue=0)
+                            elseif method == :bilinear
+                                res_IT = ImageTransformations.imrotate(cpu(arr1)[:, :, 1, 1], angle, axes(arr1)[1:2], fillvalue=0)
+                            end
+                            @show res1, res_IT 
+                            @test all(.≈(1 .+ res1[:, :, :, :], 1 .+ res_IT[:, :], rtol=rtol))
```
maxfreu, line 41: Why do you add 1 here?

test/rotation.jl
```diff
@@ -0,0 +1,80 @@
+function rotation_testsuite(Backend)
+    device(x) = adapt(Backend(), x)
+    gradtest_fn = Backend == CPU ? gradtest : gputest
+    T = Float32
+    atol = T == Float32 ? 1e-3 : 1e-6
+    rtol = T == Float32 ? 1f-2 : 1f-6
+    angles = deg2rad.([0, 0.0, 0.0001, 35, 90, -90, -90.0123, 170, 180, 270, 360, 450, 1234.1234])
```
maxfreu, line 7: `deg2rad` always returns Float64, so you can omit 0.0

test/rotation.jl
roflmaostc: because of the zeros. Sometimes zero comparisons fail (didn't check in this case)

roflmaostc:
Now buildkite does not fail on the rotations anymore...

test/rotation.jl
```diff
@@ -0,0 +1,80 @@
+function rotation_testsuite(Backend)
+    device(x) = adapt(Backend(), x)
+    gradtest_fn = Backend == CPU ? gradtest : gputest
+    T = Float32
+    atol = T == Float32 ? 1e-3 : 1e-6
+    rtol = T == Float32 ? 1f-2 : 1f-6
```
maxfreu, line 6: When I wrote the comment on relaxing the tolerances I hadn't seen that they were already high. So I suggest to restore it to 1f-3 or lower. Optimal would be to test where it fails and then set it a little higher than that.

test/rotation.jl
maxfreu: ok, just keep it then :)

src/rotation.jl
```diff
@@ -0,0 +1,286 @@
...
+
+julia> NNlib.imrotate(arr, deg2rad(45), method=:nearest)
+3×3×1×1 Array{Float64, 4}:
+[:, :, 1, 1] =
+ 0.0  0.0  1.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+```
+"""
+function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, midpoint=size(arr) .÷ 2 .+ 1) where T
```
maxfreu, line 162: Does the midpoint always have to be the center pixel? Or does the algorithm support rotations around arbitrary points within the image? If arbitrary points are supported, I'd switch from midpoint to rotation_center. Also, is the default of sz/2+1 valid for all image sizes? If not I think it would be more secure to set the default to (0,0) and to check the sizes individually within the function body and set the default there to a value that is the center, no matter which size the image has.

test/rotation.jl
```diff
@@ -0,0 +1,79 @@
...
+    @testset "imrotate" begin
+        @testset "Simple test" begin
+            arr = device(zeros((6, 6, 1, 1))); 
+            arr[3:4, 4, 1, 1] .= 1;
+            @test all(cpu(NNlib.imrotate(arr, deg2rad(45))) .≈ [0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.29289321881345254 0.585786437626905 0.0; 0.0 0.0 0.08578643762690495 1.0 0.2928932188134524 0.0; 0.0 0.0 0.0 0.08578643762690495 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.0])
+        end
+
+
+        @testset "Compare with ImageTransformations" begin
+            for sz in [(51,51,1,1), (52,52,1,1)]
```
maxfreu, line 18: Oh and what about images of odd and even size like (51,52)? Testing thoroughly is really painful :D

src/rotation.jl
roflmaostc: midpoint is meant to be the rotation_center, I change the name!

ImageTransformations does not allow for it.

There is also a difference to it. This routine always rotates around one pixel.
ImageTransformations rotates around the numerical center, which is in the case of even arrays at X.5

test/rotation.jl
roflmaostc: There's indeed some tests failing in the case of nearest neighbour interpolation.

It's just at some angles and seems more like some rounding issues to me. So not entirely sure how to treat it.

src/rotation.jl
roflmaostc: In one case (90°) ImageTransformations rather looked a little sketchy

src/rotation.jl
```diff
@@ -0,0 +1,288 @@
...
+
+julia> NNlib.imrotate(arr, deg2rad(45), method=:nearest)
+3×3×1×1 Array{Float64, 4}:
+[:, :, 1, 1] =
+ 0.0  0.0  1.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+```
+"""
+function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1) where T
```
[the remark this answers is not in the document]
roflmaostc: Since this `imrotate` has the different one than `ImageTransformations.imrotate` I have to change it to match the one of the latter.

The default `rotation_center=size(arr) .\div 2 .+1` means that both for even and odd arrays, there is a real center pixel (the same which is chosen by FFTs). For array of size 10, it is 6. For size of 11, it is also 6.

test/rotation.jl
roflmaostc: Seems like the tolerance works with 1f-3 now.

src/rotation.jl
```diff
@@ -0,0 +1,288 @@
...
+end
+
+
+# is this rrule good? 
+# no @thunk and @unthunk
+function ChainRulesCore.rrule(::typeof(imrotate), arr::AbstractArray{T}, θ; 
+                              method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1) where T
+    res = imrotate(arr, θ; method, rotation_center)
+    function pb_rotate(dy)
+        ad = ∇imrotate(unthunk(dy), arr, θ; method, rotation_center)
```
maxfreu, line 283: @mcabbott Is this rrule ok? I think thunking the output makes no sense here, right?

src/rotation.jl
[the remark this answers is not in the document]
maxfreu: Ok, thanks for the explanation.

src/rotation.jl
roflmaostc: Yeah I have no clue actually about it :smile:

mcabbott: `unthunk(dy)` seems fine.

Why define `∇imrotate` at all though, and not just call `imrotate(arr, -θ)`?

src/rotation.jl
```diff
@@ -0,0 +1,288 @@
...
+3×3×1×1 Array{Float64, 4}:
+[:, :, 1, 1] =
+ 0.0  0.0  1.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+```
+"""
+function imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1) where T
+    @assert (T <: Integer && method==:nearest || !(T <: Integer)) "If the array has an Int eltype, only method=:nearest is supported"
+    @assert typeof(rotation_center) <: Tuple "rotation_center keyword has to be a tuple"
```
mcabbott, line 164: This is minor but I think the first should test and `throw`, and the second should just be `rotation_center::Tuple` in the signature.

src/rotation.jl
roflmaostc: that is not the same under bilinear interpolation assumption, isn't it?
For nearest neighbour it is probably the same (excluding some rounding errors).

roflmaostc:
```
julia> NNlib.∇imrotate(arr, arr, 0.1)
5×5×1×1 Array{Float64, 4}:
[:, :, 1, 1] =
 0.0         0.0        0.00994175  0.0848958  0.0
 0.0848958   0.90961    0.991056    0.90961    0.0
 0.00994175  0.991056   1.01799     0.991056   0.00994175
 0.0         0.90961    0.991056    0.90961    0.0848958
 0.0         0.0848958  0.00994175  0.0        0.0

julia> NNlib.imrotate(arr, -0.1)
5×5×1×1 Array{Float64, 4}:
[:, :, 1, 1] =
 0.0         0.0        0.00999167  0.0884453  0.0
 0.0884453   0.905162   1.0         0.905162   0.0
 0.00999167  1.0        1.0         1.0        0.00999167
 0.0         0.905162   1.0         0.905162   0.0884453
 0.0         0.0884453  0.00999167  0.0        0.0
``

Comment (maxfreu) [reply src/rotation.jl:285]: How does arr look like? And I think dy should be all ones to see the right gradient, not arr, correct?

Comment (roflmaostc) [reply src/rotation.jl:285]:
```
arr = ones((5,5,1,1)
arr[2:4, 2:4, 1,1] .= 1
```

Comment (roflmaostc) [reply src/rotation.jl:285]: I think that makes sense because some pixels distribute their values over two pixels, so their rrule has 2s at exactly those pixels.

```
julia> NNlib.imrotate(arr, -0.8, method=:nearest)
5×5×1×1 Array{Float64, 4}:
[:, :, 1, 1] =
 0.0  0.0  1.0  0.0  0.0
 0.0  1.0  1.0  1.0  0.0
 1.0  1.0  1.0  1.0  1.0
 0.0  1.0  1.0  1.0  0.0
 0.0  0.0  1.0  0.0  0.0

julia> NNlib.∇imrotate(one.(arr), arr, -0.8, method=:nearest)
5×5×1×1 Array{Float64, 4}:
[:, :, 1, 1] =
 0.0  1.0  0.0  1.0  0.0
 1.0  2.0  1.0  2.0  1.0
 0.0  1.0  1.0  1.0  0.0
 1.0  2.0  1.0  2.0  1.0
 0.0  1.0  0.0  1.0  0.0
```

Comment (maxfreu) [reply src/rotation.jl:285]: I think it's ok to leave things as they are then. As long as it's correct and fast everything is ok :)

Comment (roflmaostc) [reply src/rotation.jl:285]: Yeah I think it's fine and the `rrule` tests seem to work.

**Comment by roflmaostc** [comment]
I think the only open point is the slight disagreement for nearest neighbor for some specific angles to `ImageTransformations`. 
But I believe those are rounding issues and don't change the results significantly. At least, I don't see what would be wrong here.

In fact, [ImageTransformations](https://github.com/JuliaImages/ImageTransformations.jl/blob/3be40eab0a635df38c00fa13c54b2435974a852d/test/warp.jl#L428) is also not tested for that many angles to some reference implementation.

**Comment by maxfreu** [comment]
Ok, now the tests are working - except for one gradient test. That should be investigated and if it's just about tolerance, that can be relaxed a bit. But now somehow 300+ other tests fail. Is this related to your changes or the latest merge by @ToucheSir ?

**Comment by roflmaostc** [comment]
I can check later or tomorrow for the failing gradient test.

**Comment by ToucheSir** [comment]
It may also be a regression caused by changes in a different package like Zygote or ChainRules(Core). The error looks like some method ambiguity issues have shown up again...

**Comment by roflmaostc** [comment]
I lowered the accuracy for the `gradtest`. But locally it was running fine with 1e-6 too.

**Comment by maxfreu** [comment]
I just checked the test logs more carefully and indeed, the test failure is unrelated to your changes and a victim of the same error that broke most other tests. 1e-6 should be ok, sorry for the noise.

**Comment by ToucheSir** [comment]
I see, https://github.com/JuliaDiff/ChainRules.jl/pull/762 is causing Zygote to be significantly downgraded and breaking. Normally I would be comfortable ignoring that if none of the functionality this PR adds is affected, but there do seem to be tests that downgrade causes errors in that obscure the actual error. For now, we may have to park this until the ChainRules PR is merged.

File: src/rotation.jl [hunk 20]
```diff
@@ -0,0 +1,290 @@
+"""
+    _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
+
+This rotates the coordinates and either applies round(nearest neighbour)
+or floor for :bilinear interpolation)
+"""
+@inline function _rotate_coordinatess(sinθ, cosθ, i, j, rotation_center, round_or_floor)
```
maxfreu, line 7: typo in the function name? (ss)

roflmaostc:
I think ChainRules got merged

src/rotation.jl
```diff
@@ -0,0 +1,290 @@
...
+end
+
+
+"""
+    imrotate(arr::AbstractArray{T, 4}, θ; method=:bilinear, rotation_center=size(arr) .÷ 2 .+ 1)
+
+Rotates an array in the first two dimensions around the center pixel `rotation_center`. 
+`rotation_center` is defined such that there is a real center pixel for even and odd values which is rotated around.
+For an even sized array of size `(4,4)` this would be `(3,3)`, for an odd array of size `(3,3)` this would be `(2,2)`
+
```
maxfreu, line 102: I think the documentation could be a bit clearer regarding the rotation center. You say `rotation_center` is defined such that there is a real pixel. Does that mean that the user *has to* supply the rotation center in this way? Or is it possible to rotate around any float 2D coords?

roflmaostc: It rotates around any given position. But the default is a integer center pixel.
[1,2,3,4] would be around the 3. and [1,2,3] around the 2.

maxfreu: Now it is clearer, thank you!

maxfreu:
I think this PR is finally ready :pizza:

ToucheSir (approved):
:shipit:

ToucheSir:
CI failures are known and unrelated. Thanks @roflmaostc for the great work!

roflmaostc:
Thanks for your work!

Does NNlib.jl actually have more methods with preallocated buffers?
In [DiffImageRotation.jl](https://github.com/roflmaostc/DiffImageRotation.jl/blob/823d4d2be6627847793a11e0747fdce966d49b41/src/DiffImageRotation.jl#L103) I had this special function, to save some memory allocation.

This helps if you rotate a lot of images and want to add them into the same buffer each.

ToucheSir:
Yes, see functions like `conv!` or `scatter!`. Traditionally most of those have not been differentiable, but now many have Enzyme AD rules and I think that's the best path going forwards for differentiable operations in NNlib.

roflmaostc:
Ok, looks good. I might tackle this in future then!

maxfreu:
Thanks for your work also from my side! :)

PR #560, opened 2024-01-11
Use julia-actions/cache

ToucheSir:
This does everything automatically for us and should speed up CI times. The julia-buildpkg option even warns about this, see e.g. https://github.com/FluxML/NNlib.jl/actions/runs/7492256201. 

### PR Checklist

- [ ] ~~Tests are added~~
- [ ] ~~Documentation, if applicable~~

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -5,6 +5,11 @@ on:
     tags: [v*]
   pull_request:
 
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
 jobs:
   test:
     name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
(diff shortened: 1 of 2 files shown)


ToucheSir:
Enabled this for all relevant workflows and confirmed (because I didn't do it right on the first commit) that caching is working now.

.github/workflows/Downstream.yml
```diff
@@ -37,6 +42,7 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
           arch: x64
+      - uses: julia-actions/cache@v1
```
darsnack, line 45: Why v1 instead of v3?

ToucheSir: Were you thinking of https://github.com/actions/cache? The latest version of https://github.com/julia-actions/cache is 1.5.3, so still major version v1. I know because I initially got confused by this too.

darsnack: Ah you're right. I thought this PR was just removing parameters to the action and didn't see the action itself changed.

darsnack (approved):

PR #570, opened 2024-03-26
Start replacing tests with DifferentiationInterface

gdalle:
Following today's conversation between the maintainers of Flux and DifferentiationInterface, here's a first attempt at replacing small parts of the test suite with a more generic version.

To save time I have deactivated all workflows except testing, and commented out all tests except the ones I'm working on

### PR Checklist

- [ ] Tests are added


test/softmax.jl
```diff
@@ -110,21 +119,33 @@ end
...
+
+    DIT.test_differentiation(
+        ADTypes.AutoZygote(),
+        [DI.gradient],
+        scenarios;
+        correctness=ADTypes.AutoFiniteDifferences(
+            FiniteDifferences.central_fdm(5, 1)
+        ),
+        detailed=true,
+    )
```
gdalle, line 148: This part replaces gradtest.
It compares the behavior of Zygote with that of FiniteDifferences over a set of scenarios.

gdalle:
@adrhill first roadblock with our approach: we'd like to validate the `pullback` of Zygote using FiniteDifferences, but as a "forward mode" backend, the latter doesn't have a natural pullback definitin

gdalle:
Ok the solution is to use https://juliadiff.org/FiniteDifferences.jl/latest/pages/api/#FiniteDifferences.j%E2%80%B2vp

PR #574, opened 2024-04-12
Add export for imrotate gradient

roflmaostc:
I think we should have added the export for \nabla imrotate in  #553?


mcabbott (approved):
Nnlib exports this for softmax:

https://github.com/FluxML/NNlib.jl/blob/07833637dec96d12d0614308d3145b432fdb320a/src/NNlib.jl#L55

So I don't see why not here too. It's very unlikely to collide with anything from elsewhere.

Issue #565, opened 2024-02-12
Missing EnzymeRule for `∇conv_data!`


wsmoses:
### Motivation and description

Arises when differentiating this model.

```
using Enzyme, Flux, Random, Test

Enzyme.API.runtimeActivity!(true)

loss(model, x) = sum(model(x))

function test_model(model, x, mi)
    println(model)
    l = loss(model, x)

    Flux.reset!(model)
    grads_flux = Flux.gradient(m -> loss(m, x), model)[1]

    grads_enzyme = Flux.fmap(model) do x
        x isa Array ? zero(x) : x
    end
    Flux.reset!(model)
    Enzyme.autodiff(Reverse, loss, Active, Duplicated(model, grads_enzyme), Const(x))

    @testset "Model $mi" begin
        Flux.reset!(model)
        @test loss(model, x) == l # Check loss doesn't change with multiple runs

        for i in eachindex(grads_flux.layers)
            layer_flux = grads_flux.layers[i]
            layer_enzyme = grads_enzyme.layers[i]
            for field in (:weight, :bias, :scale)
                if hasfield(typeof(layer_flux), field)
                    @test isapprox(getfield(layer_flux, field), getfield(layer_enzyme, field))
                end
            end
            if hasfield(typeof(layer_flux), :cell)
                for field in (:Wh, :Wi, :b)
                    @test isapprox(getfield(layer_flux.cell, field), getfield(layer_enzyme.cell, field))
                end
            end
        end
    end
end

models_xs = [
    [
        Chain(ConvTranspose((3, 3), 3 => 7, stride=2)),
        rand(Float32, 10, 10, 3, 50),
    ],
]
```

cc @jgreener64 @ToucheSir

PR #582, opened 2024-05-11
Enzyme: add derivatives for ∇conv_filter and ∇conv_data

wsmoses:
Add enzyme rule for missing conv utilities

Fixes https://github.com/FluxML/NNlib.jl/issues/565


wsmoses:
@CarloLucibello could I get a review here as well?

CarloLucibello (approved):

PR #583, opened 2024-05-11
Enzyme: broaden tests

wsmoses:


wsmoses:
@CarloLucibello @ToucheSir given that the AMD/CUDA tests here both pass, I am presently unaware of any nnlib + GPU 
 + Enzyme issue, if you have one.

PR #585, opened 2024-05-18
Add support for AMDGPU 0.9

pxl-th:
- Add support for AMDGPU 0.9.
- Remove MIOpen softmax wrappers (rely on pure Julia implementations) as they have issues: https://github.com/JuliaGPU/AMDGPU.jl/pull/630

### PR Checklist

- [ ] Tests are added


darsnack (approved):
Should we switch back to MIOpen implementations once the upstream issue is fixed? If so, can you create an issue here to track it?

pxl-th:
Thanks for merging! Yes, if they fix it we can add it back. Although I'd prefer to have more Julia-native kernels :)
Can we also tag a release for NNlib? I'd like to update Flux as well next.

PR #591, opened 2024-06-15
Use GPU-friendly rounding

### LINKED ISSUE #573: Warning on using Upsample on AMDGPU
i have an amd rx 7900xtx with ubuntu.
```
noise = 0.5f0 + 0.001f0 .+ AMDGPU.randn(Float32, 32, 32, 1, 1)
layer1 = Upsample(:bilinear, size=(64, 64)) |> gpu
layer1(noise)
```

when i run this i get 

```
┌ Warning: Global hostcalls detected: [:malloc_hostcall].
│ Use `AMDGPU.synchronize(; stop_hostcalls=false)` to synchronize and stop them.
│ Otherwise, performance might degrade if they keep running in the background
```

pxl-th:
- Use GPU-friendly rounding.
Otherwise we create boxed value that is used for reporting exceptions which triggers hostcalls on AMDGPU.
- Use `Int` for indexing (as it is default anyway) instead of converting to `UInt32`.

Fixes #573.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -121,7 +121,7 @@ end
     ix = compute_source_index(x, iW, padding_mode)
     iy = compute_source_index(y, iH, padding_mode)
     # Get corner pixel values from (ix, iy) in north-east-south-west directions.
-    ix_nw, iy_nw = floor(Int, ix), floor(Int, iy)
+    ix_nw, iy_nw = unsafe_trunc(Int, floor(ix)), unsafe_trunc(Int, floor(iy))
     ix_ne, iy_ne = ix_nw + 1, iy_nw
     ix_sw, iy_sw = ix_nw, iy_nw + 1
     ix_se, iy_se = ix_ne, iy_sw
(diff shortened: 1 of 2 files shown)


mcabbott (commented):
Is there any way this could have a test?

And are you reasonably sure that changing UInt32 to Int won't hurt CUDA cases? I don't know why that was written that way, it looks odd to me, but haven't investigated original PR etc.

pxl-th:
> Is there any way this could have a test?

Test that would check we are **not** launching hostcalls on AMDGPU?

> And are you reasonably sure that changing UInt32 to Int won't hurt CUDA cases? I don't know why that was written that way, it looks odd to me, but haven't investigated original PR etc.

I think that was me being obsessed with keeping `Int32` indexing back then. It might save us a few registers, but beyond that there should be no visible performance impact.
We are still producing `Int64` with that code, because `@index` returns them, so does `size(..)` and so on.

mcabbott (approved):

Issue #405, opened 2022-04-07
Convolutions error using Measurements.jl


adrhill:
Passing an Array of Measurements to a `Conv` layer throws the error `ERROR: UndefRefError: access to undefined reference`:

```julia
julia> using Flux, Measurements

(@v1.7) pkg> status Flux
      Status `~/.julia/environments/v1.7/Project.toml`
  [587475ba] Flux v0.12.10

(@v1.7) pkg> status Measurements
      Status `~/.julia/environments/v1.7/Project.toml`
  [eff96d63] Measurements v2.7.1

julia> x = rand(Float32, 5, 5, 3, 1);

julia> m = measurement.(x, 0.02f0)
5×5×3×1 Array{Measurement{Float32}, 4}:
[:, :, 1, 1] =
 0.843±0.02   0.78±0.02  0.164±0.02  0.313±0.02   0.578±0.02
 0.188±0.02  0.119±0.02  0.573±0.02  0.342±0.02  0.0012±0.02
 0.434±0.02   0.99±0.02  0.901±0.02  0.881±0.02   0.569±0.02
 0.149±0.02  0.184±0.02  0.553±0.02  0.827±0.02   0.328±0.02
 0.544±0.02  0.659±0.02  0.159±0.02  0.426±0.02   0.789±0.02

[:, :, 2, 1] =
 0.903±0.02  0.801±0.02  0.904±0.02  0.101±0.02   0.11±0.02
 0.879±0.02  0.764±0.02   0.11±0.02  0.426±0.02  0.461±0.02
  0.03±0.02  0.911±0.02  0.819±0.02   0.15±0.02  0.922±0.02
  0.91±0.02  0.392±0.02  0.838±0.02  0.356±0.02  0.277±0.02
 0.632±0.02  0.391±0.02   0.94±0.02  0.969±0.02  0.154±0.02

[:, :, 3, 1] =
 0.084±0.02  0.947±0.02  0.891±0.02  0.997±0.02  0.172±0.02
 0.815±0.02  0.285±0.02  0.243±0.02  0.427±0.02  0.639±0.02
 0.663±0.02  0.055±0.02  0.807±0.02  0.782±0.02  0.439±0.02
 0.882±0.02  0.958±0.02  0.869±0.02  0.471±0.02  0.204±0.02
 0.235±0.02  0.544±0.02  0.946±0.02  0.735±0.02  0.302±0.02

julia> c = Conv((3, 3), 3 => 1)
[...]

PR #592, opened 2024-06-17
Support `conv_direct!` on custom datatypes

### LINKED ISSUE #490: Error when using Float64: ERROR: UndefRefError: access to undefined reference
Distilled from this [discussion](https://discourse.julialang.org/t/flux-jl-convolutional-vae-throws-error-after-upgrading-to-1-4-1-10-4/39746). I couldn't remove anything further and still reproduce the crash.

Julia 1.4.1, Flux 0.10.4

```julia
using Flux                                                                       
                                                                                 
function mwe(T)                                                                  
    int1 = Dense(4, 280)                                                         
    resd(X) = reshape(int1(X), 10, 7, 4, :)                                      
    tc1 = ConvTranspose((4, 3), 4 => 4, relu, stride = (2, 2), pad = 1)          
    mdl = Chain(resd, tc1)                                                       
    z = [1,2,3,4]                                                                
    X̂ = mdl(z)                                                                   
    X = randn(T, size(X̂)...)                                                     
    loss(y) = -sum(Flux.binarycrossentropy.(mdl(z), y))                          
    ps = Flux.params(mdl)                                                           
    gs = gradient(ps) do                                                            
        loss(X)                                                                  
    end                                                                          
end                                                                                 

julia> mwe(Float32); # success

julia> mwe(Float64)

ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getindex at ./array.jl:789 [inlined]
 [2] conv_direct!(::Array{AbstractFloat,5}, ::Array{AbstractFloat,5}, ::Array{Float32,5}, ::NNlib.DenseConvDims{3,(4, 3, 1),4,4,(2, 2, 1),(1, 1, 1, 1, 0, 0),(1, 1, 1),false}; alpha::Float64, beta::Bool) at /home/russel/.julia/packages/NNlib/FAI3o/src/impl/conv_direct.jl:98
 [3] conv_direct! at /home/russel/.julia/packages/NNlib/FAI3o/src/impl/conv_direct.jl:51 [inlined]
 [4] conv!(::Array{AbstractFloat,5}, ::Array{AbstractFloat,5}, ::Array{Float32,5}, ::NNlib.DenseConvDims{3,(4, 3, 1),4,4,(2, 2, 1),(1, 1, 1, 1, 0, 0),(1, 1, 1),false}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:99
 [5] conv!(::Array{AbstractFloat,5}, ::Array{AbstractFloat,5}, ::Array{Float32,5}, ::NNlib.DenseConvDims{3,(4, 3, 1),4,4,(2, 2, 1),(1, 1, 1, 1, 0, 0),(1, 1, 1),false}) at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:97
 [6] conv!(::Array{AbstractFloat,4}, ::Array{AbstractFloat,4}, ::Array{Float32,4}, ::NNlib.DenseConvDims{2,(4, 3),4,4,(2, 2),(1, 1, 1, 1),(1, 1),false}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:70
 [7] conv! at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:70 [inlined]
 [8] conv(::Array{AbstractFloat,4}, ::Array{Float32,4}, ::NNlib.DenseConvDims{2,(4, 3),4,4,(2, 2),(1, 1, 1, 1),(1, 1),false}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:116
 [9] conv at /home/russel/.julia/packages/NNlib/FAI3o/src/conv.jl:114 [inlined]
 [10] FluxML/Flux.jl#1837 at /home/russel/.julia/packages/Zygote/YeCEW/src/lib/nnlib.jl:41 [inlined]
 [11] (::Zygote.var"#4556#back#1839"{Zygote.var"#1837#1838"{Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},Array{Float32,4},Array{Float32,4},NNlib.DenseConvDims{2,(4, 3),4,4,(2, 2),(1, 1, 1, 1),(1, 1),false}}})(::Array{AbstractFloat,4}) at /home/russel/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [12] ConvTranspose at /home/russel/.julia/packages/Flux/Fj3bt/src/layers/conv.jl:148 [inlined]
 [13] (::typeof(∂(λ)))(::Array{Float64,4}) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [14] applychain at /home/russel/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:36 [inlined]
 [15] (::typeof(∂(applychain)))(::Array{Float64,4}) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [16] applychain at /home/russel/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:36 [inlined]
 [17] (::typeof(∂(applychain)))(::Array{Float64,4}) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [18] Chain at /home/russel/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:38 [inlined]
 [19] (::typeof(∂(λ)))(::Array{Float64,4}) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [20] loss at /home/russel/Desktop/MWE.jl:13 [inlined]
 [21] (::typeof(∂(λ)))(::Float64) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [22] FluxML/Flux.jl#9 at /home/russel/Desktop/MWE.jl:16 [inlined]
 [23] (::typeof(∂(λ)))(::Float64) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [24] (::Zygote.var"#49#50"{Zygote.Params,Zygote.Context,typeof(∂(λ))})(::Float64) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface.jl:179
 [25] gradient(::Function, ::Zygote.Params) at /home/russel/.julia/packages/Zygote/YeCEW/src/compiler/interface.jl:55
 [26] mwe(::Type{T} where T) at /home/russel/Desktop/MWE.jl:15
 [27] top-level scope at REPL[5]:1
 [28] eval(::Module, ::Any) at ./boot.jl:331
 [29] eval_user_input(::Any, ::REPL.REPLBackend) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [30] run_backend(::REPL.REPLBackend) at /home/russel/.julia/packages/Revise/MgvIv/src/Revise.jl:1023
 [31] top-level scope at none:0
```

adrhill:
Closes #490, #405.

As described in #490, undefined array elements can't be indexed on all datatypes:

```Julia-repl
julia> v1 = Array{Float32}(undef, 3)
3-element Vector{Float32}:
 2.7761215f-21
 1.0f-45
 2.7765319f-21

julia> v2 = Array{Set}(undef, 3)
3-element Vector{Set}:
 #undef
 #undef
 #undef

julia> v1[1] # works
2.7761215f-21

julia> v2[1] # doesn't work
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getindex(A::Vector{Set}, i1::Int64)
   @ Base ./essentials.jl:13
 [2] top-level scope
   @ REPL[34]:1
```

Such indexing is used in the right-hand side of `conv_direct!` on main
https://github.com/FluxML/NNlib.jl/blob/627374ca51090c4e5b9cfbed5c15bdaa992891f8/src/impl/conv_direct.jl#L111

which means that `conv_direct!` doesn't support all input datatypes.

However, assignments work, which is way the problem can be solved by first setting `y` to zero, as suggested [here](https://github.com/FluxML/NNlib.jl/issues/490#issuecomment-1522434576) by @ToucheSir.

### DIFF
--- a/src/impl/conv_direct.jl
+++ b/src/impl/conv_direct.jl
@@ -81,8 +81,10 @@ function conv_direct!(
     # Use `calc_padding_regions` to determine where we do or don't need to worry about padding
     padded_regions, central_region = calc_padding_regions(cdims)
 
-    # Set outputs to zero (https://github.com/FluxML/NNlib.jl/issues/490)
-    y = fill!(y, zero(yT))
+    # Set outputs to zero to support custom datatypes (https://github.com/FluxML/NNlib.jl/issues/490)
+    if iszero(beta)
+        y = fill!(y, zero(yT))
+    end
 
     # Start with the central region
     w_region, h_region, d_region = central_region
--- a/src/impl/conv_direct.jl
+++ b/src/impl/conv_direct.jl
@@ -81,6 +81,9 @@ function conv_direct!(
     # Use `calc_padding_regions` to determine where we do or don't need to worry about padding
     padded_regions, central_region = calc_padding_regions(cdims)
 
+    # Set outputs to zero (https://github.com/FluxML/NNlib.jl/issues/490)
+    y = fill!(y, zero(yT))
+
     # Start with the central region
     w_region, h_region, d_region = central_region
     @inbounds for batch in 1:size(x, 5),
--- a/test/conv.jl
+++ b/test/conv.jl
@@ -2,6 +2,7 @@ using NNlib, Test
 using NNlib: input_size, kernel_size, channels_in, channels_out, channel_multiplier,
              stride, padding, dilation, flipkernel, output_size,
              groupcount
+using Random: AbstractRNG, SamplerType
 
 @testset "ConvDims" begin
     for T in (DenseConvDims, DepthwiseConvDims)
@@ -863,6 +864,42 @@ end
     @test size(NNlib.conv_direct!(y, x, w, cdims)) == y_size
     @test size(NNlib.∇conv_data_direct!(x, y, w, cdims)) == x_size
     @test size(NNlib.∇conv_filter_direct!(w, x, y, cdims)) == w_size
+end
+
+# https://github.com/FluxML/NNlib.jl/issues/490
+# https://github.com/FluxML/NNlib.jl/issues/405
+@testset "conv_direct! - Unusual input types" begin
+    # Create test type that can't be indexed when undefined.
+    # This simulates the worst-case scenario for custom types.
+    struct MyFloat <: Real
+        set::Set{Float32}
+    end
+    
+    # Test that direct indexing fails when undefined.
+    v = Array{MyFloat}(undef, 3)
+    @test_throws UndefRefError v[1]
+    
+    # Define minimal set of functions required for conv_direct!
+    MyFloat(x::MyFloat) = x
+    MyFloat(x::Real) = MyFloat(Set(Float32(x)))
+    
+    Base.:+(x::MyFloat, y::MyFloat) = MyFloat(only(x.set) + only(y.set))
+    Base.:*(x::MyFloat, y::MyFloat) = MyFloat(only(x.set) * only(y.set))
+    Base.promote_rule(::Type{MyFloat}, ::Type{Float32})   = MyFloat
+    Base.rand(::AbstractRNG, ::SamplerType{MyFloat}) = MyFloat(rand(Float32))
+    Base.zero(::MyFloat) = MyFloat(zero(Float32))
+    Base.zero(::Type{MyFloat}) = MyFloat(zero(Float32))
+    
+    # Test conv_direct!
+    x_size = (6, 7, 8, 5, 3)
+    y_size = (5, 6, 7, 4, 3)
+    w_size = (2, 2, 2, 5, 4)
+    x = rand(MyFloat, x_size);
+    w = randn(Float32, w_size);
+    y = Array{MyFloat}(undef, y_size...);
+    cdims = DenseConvDims(x_size, w_size)
+    y_out = NNlib.conv_direct!(y, x, w, cdims)
+    @test size(y_out) == y_size
 end
 
 @testset "AutoDiff: spatial_rank=$spatial_rank" for spatial_rank in (1, 2, 3)
--- a/test/conv.jl
+++ b/test/conv.jl
@@ -899,6 +899,8 @@ end
     y = Array{MyFloat}(undef, y_size...);
     cdims = DenseConvDims(x_size, w_size)
     y_out = NNlib.conv_direct!(y, x, w, cdims)
+
+    @test eltype(y_out) == MyFloat
     @test size(y_out) == y_size
 end
 


src/impl/conv_direct.jl
```diff
@@ -81,6 +81,11 @@ function conv_direct!(
     # Use `calc_padding_regions` to determine where we do or don't need to worry about padding
     padded_regions, central_region = calc_padding_regions(cdims)
 
+    # Set outputs to zero to support custom datatypes (https://github.com/FluxML/NNlib.jl/issues/490)
+    if iszero(beta)
```
gdalle, line 85: If this zeroing impacts performance, perhaps we could check here whether `eltype(y) <: AbstractFloat` and avoid it then?

adrhill: I could also have subtyped `MyFloat` from the test set as an `AbstractFloat`, so that's not enough. What you would actually want to test for is whether the elements are `#undef` or not.

gdalle: Not enough but better than nothing. The right test is probably `eltype(y) <: Union{...}` and we need to figure out all the types for which we can multiply an `undef` element by zero and get away with it.

adrhill: > Not enough but better than nothing. 

To my understanding, `conv_direct!` is supposed to be a fallback implementation for odd datatypes that is implemented naively using for-loops:

> ```Julia
> """
>     conv_direct!(y, x, w, cdims; alpha=1, beta=0)
>
> Direct convolution implementation; used for debugging, tests, and mixing/matching of 
> strange datatypes within a single convolution. Uses naive nested for loop implementation

so I would argue that a large gain in compatibility at the cost of a most likely negligible decrease in performance is perfectly acceptable.

adrhill:
Looks like failures on nightly are unrelated.

mcabbott (approved):

adrhill:
@mcabbott could a patch release be tagged for this?

pxl-th:
Tagged `0.9.18` release: https://github.com/FluxML/NNlib.jl/commit/62f6074a3944ba7db8f09f928db2121f0ea7ef07

adrhill:
Thanks!

src/impl/conv_direct.jl
```diff
@@ -81,6 +81,11 @@ function conv_direct!(
...
+        y = fill!(y, zero(yT))
```
Seelengrab, line 86: As [I mentioned in the original issue](https://github.com/FluxML/NNlib.jl/issues/490#issuecomment-1522434567), naive `fill!`/`fill` doesn't necessarily work with e.g. `BigInt`:

```
julia> a = Vector{BigInt}(undef, 2)
2-element Vector{BigInt}:
 #undef
 #undef

julia> fill!(a, zero(BigInt))
2-element Vector{BigInt}:
 0
 0

julia> a[1] ===a[2]
true
```

I don't know how important these kinds of usecases are for NNlib.jl though.

adrhill: Is that problematic here?
```Julia
julia> a = Vector{BigInt}(undef, 2)
2-element Vector{BigInt}:
 #undef
 #undef

julia> fill!(a, zero(BigInt))
2-element Vector{BigInt}:
 0
 0

julia> for (i, ai) in enumerate(a)
         a[i] = i + 0 * a[i]
       end

julia> a
2-element Vector{BigInt}:
 1
 2
```

PR #598, opened 2024-07-16
ci: update downstream testing

avik-pal:
- removes 3 packages. they were last updated more than 2 years back and don't even support latest NNlib, so no point in spawning those jobs.
- Lux uses ReTestItems.jl so I added some of the env vars that allow parallel testing.


CarloLucibello (approved):

avik-pal:
@CarloLucibello you'll have to merge, I don't have write access

CarloLucibello:
now you have write access! people with write access merge their own PRs conditioned on having received an approval

PR #599, opened 2024-07-16
feat: catch scalar indexing failures early

avik-pal:
fixes #523 (maybe also https://github.com/FluxML/Flux.jl/issues/2440)

```julia
using CUDA, NNlib

x = cu(rand(Float32, 4, 4, 3, 2))
w = cu(rand(Float32, 3, 3, 3, 4))
cdims = DenseConvDims(x, w)

conv(x, w, cdims)
```

<details>
<summary>New Error Message</summary>

```julia
ERROR: AssertionError: `conv!` requires all arguments to support fast scalar indexing. You might be missing an `using cuDNN` or `import cuDNN` statement.
Stacktrace:
  [1] special_scalar_indexing_error(::Val{:conv!}, ::CuArray{Float32, 5, CUDA.DeviceMemory})
    @ NNlibCUDAExt /mnt/research/lux/NNlib.jl/ext/NNlibCUDAExt/utils.jl:41
  [2] (::Base.Fix1{typeof(NNlib.special_scalar_indexing_error), Val{:conv!}})(y::CuArray{Float32, 5, CUDA.DeviceMemory})
    @ Base ./operators.jl:1127
  [3] (::Base.var"#70#71"{Base.Fix1{typeof(NNlib.special_scalar_indexing_error), Val{:conv!}}})(::Nothing, x::CuArray{Float32, 5, CUDA.DeviceMemory})
    @ Base ./tuple.jl:692
  [4] BottomRF
    @ ./reduce.jl:86 [inlined]
  [5] afoldl(::Base.BottomRF{Base.var"#70#71"{Base.Fix1{typeof(NNlib.special_scalar_indexing_error), Val{:conv!}}}}, ::Nothing, ::CuArray{Float32, 5, CUDA.DeviceMemory}, ::CuArray{Float32, 5, CUDA.DeviceMemory}, ::CuArray{Float32, 5, CUDA.DeviceMemory})
    @ Base ./operators.jl:553
  [6] _foldl_impl(op::Base.BottomRF{Base.var"#70#71"{Base.Fix1{typeof(NNlib.special_scalar_indexing_error), Val{:conv!}}}}, init::Nothing, itr::Tuple{CuArray{Float32, 5, CUDA.DeviceMemory}, CuArray{Float32, 5, CUDA.DeviceMemory}, CuArray{Float32, 5, CUDA.DeviceMemory}})
    @ Base ./reduce.jl:68
[...]

PR #600, opened 2024-07-16
Update to 1.10 (new LTS)

avik-pal:
This is a follow-up of https://github.com/FluxML/NNlib.jl/pull/603 (ignore the diff till that is merged)


ToucheSir:
The main question with this and #599 is when 1.11 comes out. Once it becomes stable, presumably 1.10 will be officially the LTS and listed as such on julialang.org. Are we in a major rush to jump the gun and set compat higher now?

PR #601, opened 2024-07-17
fix: MIOpen only supports activations till dimension 5

avik-pal:
See failure without the patch https://buildkite.com/julialang/nnlib-dot-jl/builds/1394#0190be92-4c29-4fd6-8bf0-5c02065f9304/277-655


pxl-th (approved):
Look good!
We probably should benchmark MIOpen vs Julia implementations (in the future) to see if using MIOpen for them provides any substantial benefit.

Issue #203, opened 2020-06-01
NNPACK convolution issue


Alexander-Barth:
The convolution with NNPACK do not seem to work reliably. The output of all these convolutions should be zero. Sometimes they are, but sometimes they have a strange value (maybe uninitialized memory)? Is this a known issue?

Thanks

```
julia> @time extrema(NNlib.conv(zeros(Float32,100,100,10,32),ones(Float32,4,4,10,10)))
  0.044346 seconds (44 allocations: 23.707 MiB)
(0.0f0, 0.0f0)

julia> @time extrema(NNlib.conv(zeros(Float32,100,100,10,32),ones(Float32,4,4,10,10)))
  0.043753 seconds (44 allocations: 23.707 MiB)
(0.0f0, 0.0f0)

julia> @time extrema(NNlib.conv(zeros(Float32,100,100,10,32),ones(Float32,4,4,10,20)))
  0.055470 seconds (44 allocations: 35.205 MiB)
(NaN32, NaN32)

julia> @time extrema(NNlib.conv(zeros(Float32,100,100,10,32),ones(Float32,4,4,10,30)))
  0.197568 seconds (144.16 k allocations: 54.368 MiB)
(-0.00033701234f0, 2.953125f0)

julia> NNlib.is_nnpack_available()
true

julia> NNlib.NNPACK_CPU_THREADS
0x0000000000000008

julia> NNlib.libnnpack
"/home/abarth/.julia/dev/NNlib/deps/usr/lib/libnnpack.so"


```

ljluestc:
Ensure that NNPACK is installed correctly by checking that the library path is correct (/home/abarth/.julia/dev/NNlib/deps/usr/lib/libnnpack.so). Make sure you have the latest version of NNPACK installed, as newer versions may contain bug fixes and improvements.

 NNlib.NNPACK_CPU_THREADS appears to be 0x0000000000000008. This value corresponds to 8 CPU threads. You might want to verify that it matches your system configuration.

Alexander-Barth:
Thank you for your input!
For some reasons the loading now fails on my system:

```
julia> using NNPACK_jll

julia> using NNlib
┌ Warning: Error requiring `NNPACK_jll` from `NNlib`
│   exception =
│    LoadError: ArgumentError: Package NNlib does not have NNPACK_jll in its dependencies:
│    - You may have a partially installed environment. Try `Pkg.instantiate()`
│      to ensure all packages in the environment are installed.
[...]

PR #603, opened 2024-08-29
remove `NNPACK` and move `ForwardDiff` to an extension

avik-pal:
- `NNPACK` currently cannot be used (has been so since the last minor release). See https://github.com/FluxML/NNlib.jl/issues/203#issuecomment-1607385126. I don't see why they need to be left here
- Moves ForwardDiff to an extension (any reason why it was left to be loaded via Requires?)
- As a consequence of the above 2, we can remove the dep on Requires
- Also Pkg.jl was imported for some reason :cold_sweat:


CarloLucibello (approved):

ToucheSir:
Thanks. I think what happened is that the 1.9 + extensions + Enzyme upgrades took up a lot of energy, leaving none left for cleanup. NNPACK was also a problem because the test coverage was never fixed after it broke a while ago. Either way, we can probably get away with this in a minor release.

PR #606, opened 2024-09-18
Bump patch version to release new Enzyme support

wsmoses:
Now with the new Enzyme released, we should release a patch here so folks aren't broken


avik-pal (approved):

avik-pal:
I pushed a commit. That needs to be reverted before merge

avik-pal (approved):

PR #612, opened 2024-11-08
Add a switch to turn slow paths into errors

### LINKED ISSUE #478: warn on `batched_mul_generic!`?
The `batched_mul_generic!` is really slow on both cpu and gpu, but somehow it isn't something easy to find. It might worth displaying a one time warning like the cuda scalar indexing warning.

mcabbott:
Closes #478

Ideas for a better name are welcome!

### PR Checklist

- [x] Tests are added

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -164,3 +164,10 @@ NNlib.glu
 NNlib.within_gradient
 bias_act!
 ```
+
+Finally, this switch changes warnings on various fallback paths into errors.
+It's a bit like `CUDA.allowscalar(false)`.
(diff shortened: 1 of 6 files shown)


ToucheSir:
My first impulse was `allowfallbacks`, but that might be less intuitive than `allowslow`. All this to say, no objections to the current name.

CarloLucibello (approved):

ToucheSir:
Is this a good time to think about making switches like these contextually configurable via `ScopedValue`s? Those are finally in a stable Julia release.

PR #617, opened 2024-12-21
Fix `NaN` gradient in spectrogram

pxl-th:
With `power > 0` gradient will become `NaN` where spectrogram is `~0` (because of singularity at `abs(0)`).
Adding small epsilon fixes this issue (and the results match pytorch).

```julia
x = gpu(ones(Float32, 1024, 1, 1))
window = gpu(hann_window(1024)))
l, g = Zygote.withgradient(x) do x
    sum(spectrogram(x; n_fft=1024, hop_length=256, power=2.0, window)
end
```

### PR Checklist

- [x] Tests are added


mcabbott (approved):
Sounds fine, a test would be great.

PR #618, opened 2025-01-04
Use `return_type` not `_return_type` to fix an error on nightly

mcabbott:
This change should be harmless on Julia 1.10 and 1.11, but fix an error on 1.12. It's an internal function so we did sign up to fix things as needed.

```
julia> VERSION
v"1.10.7"

julia> Core.Compiler.return_type(sin, Tuple{Int})
Float64

julia> Core.Compiler._return_type(sin, Tuple{Int})
Float64
```

```
julia> VERSION
v"1.11.2"

julia> Core.Compiler.return_type(sin, Tuple{Int})
Float64

julia> Core.Compiler._return_type(sin, Tuple{Int})
Float64
```

Master:
```
julia> VERSION
v"1.12.0-DEV.1731"

julia> Core.Compiler.return_type(sin, Tuple{Int})
Float64

julia> Core.Compiler._return_type(sin, Tuple{Int})
ERROR: MethodError: no method matching _return_type(::typeof(sin), ::Type{Tuple{Int64}})
You may have intended to import Base._return_type
The function `_return_type` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _return_type(::Compiler.AbstractInterpreter, ::DataType)
   @ Base ../usr/share/julia/Compiler/src/typeinfer.jl:1244
```

### DIFF
--- a/src/bias_act.jl
+++ b/src/bias_act.jl
@@ -8,7 +8,7 @@ const RCR = RuleConfig{>:HasReverseMode}
 @inline only_derivative(y,f::F,x) where F = only(only(ChainRulesCore.derivatives_given_output(y, f, x)))
 
 # This has no methods, used for testing whether `derivatives_given_output(Ω, f, x)`
-# is independent of `x`, as `_return_type` says `Union{}` when calling is an error.
+# is independent of `x`, as `return_type` says `Union{}` when calling is an error.
 struct NotaNumber <: Real end
 
 """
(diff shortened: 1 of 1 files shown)


mcabbott:
Note aside that this package tests Zygote but does not constrain the version, hence CI is testing Zygote v0.7.0 here, with thunks from https://github.com/FluxML/Zygote.jl/pull/966

ToucheSir (approved):
I don't recall why we opted to not constrain test dependencies in `test/Project.toml`, does anyone remember?

Issue #628, opened 2025-02-05
"Full" gelu without approximation


se-schmitt:
### Motivation and description

Currently, only the tanh approximation of the gelu activation function is implemented. But the full gelu (`0.5x(1 + erf(x/sqrt(2)))`) is the default in PyTorch and Tensorflow and used in many pretrained models (like BERT). When using pretrained models from HuggingFace, e.g. through Transformers.jl, this causes slight but significant differences. Though, it would be nice to have the option to use the "full" gelu (see also this [PR](https://github.com/chengchingwen/Transformers.jl/pull/209) in Transformers.jl).

### Possible Implementation

I would make a PR to add the full gelu. Options:
- new function `gelu_full` -> least impact
- replace `gelu` and rename existing gelu to e.g. `gelu_tanh` -> impact existing code
- add kwarg like `gelu(x; approximate=false)=...` -> not sure if this this is easily possible as the kw needs to be propagated through the code (?)
What would be the best option?

ToucheSir:
One idea would be to rename the current `gelu` to `gelu_fast`, matching the pattern we've established with other activation functions. However, I don't understand https://github.com/chengchingwen/Transformers.jl/pull/209#issuecomment-2635577038 well enough to know if this would help with that.

se-schmitt:
I added a PR and named the functions as suggested. I think the comment is meant in a way that the configs of Transformers.jl should match the HuggingFace nomenclature (otherwise they would'nt be loaded automatically). With the submitted PR here, that would be possible as it enables the correct assignment of the HuggingFace `hidden_act` options `gelu` and `gelu_python` to the new `gelu` function implemented in the PR. Is this correct, @chengchingwen ?

chengchingwen:
Yes, https://github.com/chengchingwen/Transformers.jl/pull/209#issuecomment-2635577038 is simply saying we need separate functions for gelu w/ and w/o tanh approx. and it would be nice if NNlib.jl has both.

avik-pal:
Changing the implementation of a function as common as `gelu` should be a breaking change

chengchingwen:
This kind of implementation change wasn't considered a breaking change before (#371 is released with `v0.7.31`). I think it won't be that breaking on the accuracy/value side, though the performance side might.

avik-pal:
> this causes slight but significant differences.

This would be true in the reverse direction as well, making any trained models that use the approx gelu impl have differences once this change is made.

[...]

PR #629, opened 2025-02-06
Adds full gelu without approximation

se-schmitt:
Adds the full gelu without approximation as `gelu(x)` and moves the tanh approximation used before to `gelu_fast`. See #628 for details.

### PR Checklist

- [x] Tests are added


ToucheSir:
https://github.com/FluxML/NNlib.jl/actions/runs/13177183802/job/36779142351?pr=629#step:7:842 is a real test failure. I think Flux's `Nil` + `outputsize` machinery needs to be adjusted to understand `SpecialFunctions.erf`. The question is how, so I've opened to track this.

mcabbott:
Quick look at how different these functions are:
```julia
julia> using SpecialFunctions, NNlib

julia> oftf(x, y) = oftype(float(x), y);

julia> new_gelu(x) = x/2*(1 + erf(x/sqrt(oftf(x,2))));

julia> rel(x) = (new_gelu(x) - gelu(x)) / new_gelu(x);

julia> rel.(-3:0.2f0:1)
21-element Vector{Float32}:
   0.101809666
   0.06506126
   0.038386323
   0.02019246
   0.008700018
   0.0021531228
  -0.0010194147
  -0.0021005166
  -0.0020575877
  -0.001547056
  -0.0009626968
  -0.00049560843
  -0.0002001288
  -5.4488235f-5
  -6.02081f-6
 NaN
   4.4374747f-6
   2.876006f-5
   7.55584f-5
   0.00013319723
   0.00018150361

julia> rel_eps(x) = (new_gelu(x) - gelu(x)) / eps(new_gelu(x));

julia> Int.(rel_eps.(-3:0.2f0:1))
21-element Vector{Int64}:
 -885402
 -999594
 -499514
 -213282
 -142868
  -26298
    8849
   24719
   31223
   14336
   10250
    5637
    2210
     504
      68
       0
      69
     253
    1104
    1409
    2562
```

se-schmitt:
I modified it such that `gelu` remains the same and added the full gelu as `gelu_full` as discussed in #628 . This avoids breaking changes and the test failure from above, however, `gelu_full` is still not compatible with Flux' outputsize function.

se-schmitt:
@ToucheSir Can this be merged? If not, what is required to make it mergeable?

mcabbott:
It's a little sad that NNlib must depend on SpecialFunctions... maybe not so expensive?
```
julia> @time_imports using SpecialFunctions
      8.7 ms  IrrationalConstants
               ┌ 0.0 ms DocStringExtensions.__init__() 
     46.6 ms  DocStringExtensions 97.36% compilation time
      0.6 ms  LogExpFunctions
               ┌ 2.5 ms OpenLibm_jll.__init__() 
      4.2 ms  OpenLibm_jll
      0.4 ms  JLLWrappers
               ┌ 9.2 ms CompilerSupportLibraries_jll.__init__() 
     11.1 ms  CompilerSupportLibraries_jll
               ┌ 6.0 ms OpenSpecFun_jll.__init__() 93.49% compilation time
      6.5 ms  OpenSpecFun_jll 86.17% compilation time
      3.2 ms  SpecialFunctions
```

Re name bike-shedding, some chance we should use neutral names like `gelu_erf` and `gelu_tanh`, with both names available immediately but `const gelu = gelu_tanh` for now to be non-breaking. (I do not think either should be called `gelu_fast`, as the point of `tanh_fast` is that we sometimes automatically replace `tanh` with that, but there is no plan to automatically replace one of these with the other.)

se-schmitt:
@mcabbott I modified the code as suggested:

- Instead of the SpecialFuncitons.jl package, only OpenLibm_jll.jl is used now and the erf function is defined via `ccall` (as in SpecialFunctions.jl). 
- I also renamed the functions to `gelu_tanh` and `gelu_erf` with `const gelu = gelu_tanh`. I made this also transparent in the documentation.

src/activations.jl
```diff
@@ -357,6 +357,39 @@ function deriv_gelu(x)
...
+"""
+gelu_erf(x) = x/2*(1 + _erf(x/sqrt(oftf(x,2))))
+
+function deriv_gelu_erf(x)
+    SQRT2 = sqrt(oftf(x,2))
+    Φ = (1 + _erf(x/SQRT2))/2
+    Φ + x/SQRT2*exp(-(x^2)/2)/sqrt(oftf(x,π))
+end
+
+_erf(x::Number) = _erf(float(x))
```
mcabbott, line 373: This intends to catch integers but isn't a safe pattern, as there are other weird numbers out there:
```
StackOverflowError:
  Stacktrace:
   [1] _erf(x::ForwardDiff.Dual{ForwardDiff.Tag{Zygote.var"#141#142"{var"#170#178"{typeof(gelu_erf)}}, Float64}, Float64, 1}) (repeats 79984 times)
     @ NNlib ~/work/NNlib.jl/NNlib.jl/src/activations.jl:373
```
This particular case could be allowed via NNlibForwardDiffExt

mcabbott (commented):
This looks basically fine to me, thanks.

Maybe avoiding SpecialFunctions is a rabbit-hole, sorry. 

One question is: How well do these variants work on the GPU? Presumably `ccall((:erff, libopenlibm), Float32, (Float32,), x)` won't work... does SpecialFunctions have code to make `erf.(cu(rand(10)))` work by another path?

ToucheSir:
> does SpecialFunctions have code to make `erf.(cu(rand(10)))` work by another path?

Yes, CUDA.jl defines its own overloads at https://github.com/JuliaGPU/CUDA.jl/blob/master/ext/SpecialFunctionsExt.jl.

If we want to talk load times, Flux has a direct dep on SpecialFunctions already. If import latency is a pressing concern, we could define a stub `gelu_erf` in NNlib and the method for that function in a `SpecialFunctionsExt`.

se-schmitt:
The implementation via OpenLibm_jll was naive, sorry... I added the missing rules for AD locally which made it compatible with ForwardDiff, Zygote and Enzyme. However, compatibility with other AD and the GPU packages would need further modifications. 

I would prefer an option that includes SpecialFunctions.jl as this seems much cleaner to me (either as direct dependency or extension (is this a problem for Lux.jl where SpecialFun is not a direct dependency?)).

What would you prefer? @mcabbott @ToucheSir

ToucheSir:
Can you try the stub function + extension approach I suggested above? If that turns out to be a dead end, I'm fine with going back to the original plan and having SpecialFunctions as a direct dep. There's already an outsized chance any user of NNlib will have it in their environment, so we wouldn't lose much by including it.

se-schmitt:
@ToucheSir The extension approach worked out well. I tested `gelu_erf` with Flux and Lux and it worked seamlessly.

ext/NNlibSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,16 @@
+module NNlibSpecialFunctionsExt
+
+using NNlib
+import NNlib: oftf
+using SpecialFunctions
```
ToucheSir, line 5:
```suggestion
using NNlib: NNlib, oftf
using SpecialFunctions: erf
```

src/activations.jl
```diff
@@ -357,6 +357,25 @@ function deriv_gelu(x)
...
+function deriv_gelu_erf end
+
+"""
+    gelu(x) = gelu_tanh(x)
+
+Activation function from ["Gaussian Error Linear Units"](https://arxiv.org/abs/1606.08415). 
+See [`gelu_tanh`](@ref).
+"""
+const gelu = gelu_tanh
+export gelu
```
ToucheSir, line 376:
```suggestion
```
This is already covered by https://github.com/FluxML/NNlib.jl/blob/307d402e2cdad99f25b315fbb6a0d86a5081ef93/src/NNlib.jl#L27.

ToucheSir: On that note, could you add `gelu_tanh` to the list at https://github.com/FluxML/NNlib.jl/blob/307d402e2cdad99f25b315fbb6a0d86a5081ef93/src/activations.jl#L6?

ToucheSir (commented):
Looking good, just a couple of touch-ups and let's get this merged.

src/activations.jl
se-schmitt: `gelu_tanh` and `gelu_erf` already are in `ACTIVATIONS`, but I removed `gelu` as it seemed like a duplicate for me. Thats why I added `export gelu` here.

se-schmitt: Is there a reason that `gelu` should be in the list?

ToucheSir: Yes, removing it would break backwards compatibility.

se-schmitt: Done!

ToucheSir (approved):
Thanks, this was great work :)

PR #633, opened 2025-03-12
Don't spawn if: only one iteration or no threads or threading disabled

IanButterworth:
A single pass of ObjectDetector.jl creates 150 new tasks, all of which were within for loops with length-1 iterators.. i.e. the new task can be entirely avoided.

With this there are 0 spawned tasks.

Avoiding constructing the iterators entirely should be possible too, but I thought I'd open a PR with these changes before looking into that.

### PR Checklist

- [x] Tests are added


src/conv.jl
```diff
@@ -288,11 +304,19 @@ for (front_name, backend, signature) in (
                                 C_in = channels_in(cdims) ÷ groupcount(cdims),
                                 C_out = channels_out(cdims) ÷ groupcount(cdims))
 
-            Threads.@sync for (wc, xc, yc) in zip(dw_cs, x_cs, dy_cs)
+            function do_work(wc, xc, yc)
                 x = @view in1[ntuple(i -> i == 4 ? xc : Colon(), 5)...]
                 dy = @view in2[ntuple(i -> i == 4 ? yc : Colon(), 5)...]
-                dw = @view out[ntuple(i -> i == 5 ? yc : Colon(), 5)...]
-                Threads.@spawn $(Symbol("$(front_name)_$(backend)!"))(dw, x, dy, cdims2; kwargs...)
+                dw = @view out[ntuple(i -> i == 5 ? yc : Colon(), 5)...] # TODO: Is this supposed to use wc?
```
IanButterworth, line 310: I noticed that `yc` probably should be `wc` here?

ToucheSir: I agree. Just to be sure, can you test that `length(dy_cs) >= length(dw_cs)` holds? Maybe via a temporary test assert. I'm pretty sure it does, but the conv code is convoluted enough that it might not.

IanButterworth: Added

ToucheSir: Sorry, I meant checking the length of the partitioned iterators before the loop. i.e. `length(dy_cs) >= length(dw_cs) || error("Inconsistent dimensions") # TODO: remove this check`.

IanButterworth: Ah yes. Fixed

IanButterworth:
I'm also thinking it would be good to be able to opt out of threading.

ToucheSir:
An opt-out would be good, but IMO new ones should be done with `ScopedValue`s instead of global toggles. Is https://github.com/vchuravy/ScopedValues.jl a good enough compat package for those?

IanButterworth:
Sounds good. I think that can be a follow-on PR

IanButterworth:
Just trying something out. Will switch `ALLOW_THREADING[]` to a ScopedValue next.

IanButterworth:
Just noting that I tried simplifying this into a macro
```julia
macro maybe_thread(ex)
    ex.head == :for || throw(ArgumentError("Expected a for loop"))
    it, iter = ex.args[1].args
    loop_body = ex.args[2]
    quote
        if should_use_spawn() && length($(esc(iter))) > 1
            Threads.@sync for $(esc(it)) in $(esc(iter))
                Threads.@spawn $(esc(loop_body))
            end
        else
            $(esc(ex))
        end
    end
end
```
i.e. simply
```julia
@maybe_thread for (xc, wc) in zip(x_cs, w_cs)
    x = @view in1[ntuple(i -> i == 4 ? xc : Colon(), 5)...]
    w = @view in2[ntuple(i -> i == 5 ? wc : Colon(), 5)...]
    y = @view out[ntuple(i -> i == 4 ? wc : Colon(), 5)...]
    $(Symbol("$(front_name)_$(backend)!"))(y, x, w, cdims2; kwargs...)
end
```
but hit some macro hygiene issue with `@sync`
```
ERROR: LoadError: syntax: invalid let syntax around task.jl:497
Stacktrace:
 [1] top-level scope
   @ none:1
 [2] eval(m::Module, e::Any)
   @ Core ./boot.jl:430
 [3] top-level scope
   @ ~/Documents/GitHub/NNlib.jl/src/conv.jl:183
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:557
 [5] include(x::String)
   @ NNlib ~/Documents/GitHub/NNlib.jl/src/NNlib.jl:1
 [6] top-level scope
   @ ~/Documents/GitHub/NNlib.jl/src/NNlib.jl:80
```

src/NNlib.jl
```diff
@@ -13,11 +13,27 @@ using LinearAlgebra
...
 const Numeric = Union{AbstractArray{<:T}, T} where {T<:Number}
 
+const ALLOW_SPAWNS = ScopedValue(true)
+should_use_spawn() = Threads.nthreads(:default) > 1 && ALLOW_SPAWNS[]
+"""
+    @disallow_spawns ex
+
+Disallow NNlib to use `@spawn` on divisible workloads. i.e. within `conv` etc.
+"""
+macro disallow_spawns(ex)
```
ToucheSir, line 29: Would it make sense to have this as a function instead? At first I thought the macro would be better because it doesn't introduce new scope, but on second glance the callback to `with` is.

My worry would be that users run into closure boxing problems because they're not aware of this scope. At least with a function `disallow_spawns() do ... end`, that closure piece is explicit?

src/NNlib.jl
```diff
@@ -13,11 +13,27 @@ using LinearAlgebra
 using LinearAlgebra.BLAS: @blasfunc, BlasInt
 using LinearAlgebra: AdjOrTransAbsMat, Adjoint, BlasFloat, Transpose
 using Random
+using ScopedValues
 using Statistics
 using Statistics: mean
 
 const Numeric = Union{AbstractArray{<:T}, T} where {T<:Number}
 
+const ALLOW_SPAWNS = ScopedValue(true)
```
ToucheSir, line 22: Another thought: would it make sense to have this be an integer param that controls the max degree of spawn parallelism NNlib can use? Setting that param <= 1 (or to `false`) would disable spawning as before.

src/conv.jl
```diff
@@ -202,11 +202,21 @@ for (front_name, backend, signature) in (
                                 C_in = channels_in(cdims) ÷ groupcount(cdims),
                                 C_out = channels_out(cdims) ÷ groupcount(cdims))
 
-            Threads.@sync for (xc, wc) in zip(x_cs, w_cs)
+            function do_work(xc, wc)
```
ToucheSir, line 205:
```suggestion
            function conv_group(xc, wc)
```
For ease of debugging, I think it would help to give each `do_work` function a unique name. Ideally a more informative one too.

src/conv.jl
```diff
@@ -246,11 +256,21 @@ for (front_name, backend, signature) in (
                                 C_in = channels_in(cdims) ÷ groupcount(cdims),
                                 C_out = channels_out(cdims) ÷ groupcount(cdims))
 
-            Threads.@sync for (xc, yc, wc) in zip(dx_cs, dy_cs, w_cs)
+            function do_work(xc, yc, wc)
```
ToucheSir, line 259:
```suggestion
            function ∇conv_data_group(xc, yc, wc)
```
Ditto here.

src/conv.jl
```diff
@@ -288,11 +308,21 @@ for (front_name, backend, signature) in (
                                 C_in = channels_in(cdims) ÷ groupcount(cdims),
                                 C_out = channels_out(cdims) ÷ groupcount(cdims))
 
-            Threads.@sync for (wc, xc, yc) in zip(dw_cs, x_cs, dy_cs)
+            function do_work(wc, xc, yc)
```
ToucheSir, line 311:
```suggestion
            function ∇conv_filter_group(wc, xc, yc)
```

src/gemm.jl
```diff
@@ -104,22 +104,34 @@ for (gemm, elt) in gemm_datatype_mappings
 
                 old_threads = get_num_threads()
                 set_num_threads(1)
-                Threads.@sync for ks in Iterators.partition(1:size(C, 3), cld(size(C, 3), n_threads))
-                    Threads.@spawn for k in ks
+
+                parts = Iterators.partition(1:size(C, 3), cld(size(C, 3), n_threads))
+
+                function do_work(ks)
```
ToucheSir, line 110:
```suggestion
                function gemm!_part(ks)
```

src/impl/conv_im2col.jl
```diff
@@ -47,20 +47,28 @@ function conv_im2col!(
...
-            col_slice = col_slice = view(col, :, :, task_n) # col_slice is a task-local workspace
-            for batch_idx in part
-                im2col!(col_slice, view(x, :, :, :, :, batch_idx), cdims)
-                GC.@preserve col_slice w y begin
-                    col_ptr = pointer(col_slice)
-                    w_ptr = pointer(w)
-                    y_ptr = pointer(y, (batch_idx - 1)*M*N + 1)
-                    gemm!(Val(false), Val(false), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
-                end
+    function do_work(task_n, part)
```
ToucheSir, line 50:
```suggestion
    function conv_part(task_n, part)
```

src/impl/conv_im2col.jl
```diff
@@ -152,18 +160,25 @@ function ∇conv_data_im2col!(
...
-            col_slice = col_slice = view(col, :, :, task_n) # col_slice is a task-local workspace
-            for batch_idx in part
-                GC.@preserve col_slice w dy begin
-                    dy_ptr = pointer(dy, (batch_idx - 1)*M*K + 1)
-                    w_ptr = pointer(w)
-                    col_ptr = pointer(col_slice)
-                    gemm!(Val(false), Val(true), M, N, K, alpha, dy_ptr, w_ptr, T(0), col_ptr)
-                end
-                col2im!(view(dx, :, :, :, :, batch_idx), col_slice, cdims, beta)
+    function do_work(task_n, part)
```
ToucheSir, line 163:
```suggestion
    function ∇conv_data_part(task_n, part)
```

src/impl/depthwiseconv_im2col.jl
```diff
@@ -30,25 +30,32 @@ function depthwiseconv_im2col!(
...
-                # We do a separate convolution for each channel in x, as we must
-                for c_in in 1:channels_in(cdims)
-                    # Walk each pointer forward as we process each input channel
-                    GC.@preserve col_slice w y begin
-                        col_ptr = pointer(col_slice, (c_in-1)*M*K+1)
-                        w_ptr = pointer(w, (c_in-1)*K*N+1)
-                        y_ptr = pointer(y, ((batch_idx - 1)*channels_in(cdims) + c_in - 1)*M*N + 1)
-                        gemm!(Val(false), Val(false), M, N, K, alpha, col_ptr, w_ptr, beta, y_ptr)
-                    end
+    function do_work(task_n, part)
```
ToucheSir, line 33:
```suggestion
    function depthwiseconv_part(task_n, part)
```

src/impl/depthwiseconv_im2col.jl
```diff
@@ -117,22 +124,29 @@ function ∇depthwiseconv_data_im2col!(
...
-                # We do a separate convolution for each channel in x, as we must
-                for cidx in 1:channels_in(cdims)
-                    GC.@preserve col_slice w dy begin
-                        # Walk each pointer forward as we process each input channel
-                        dy_ptr = pointer(dy, (batch_idx - 1)*M*K*channels_in(cdims)+(cidx - 1)*K*M + 1)
-                        w_ptr = pointer(w, (cidx - 1)*K*N + 1)
-                        col_ptr = pointer(col_slice, (cidx - 1)*M*N + 1)
-                        gemm!(Val(false), Val(true), M, N, K, alpha, dy_ptr, w_ptr, T(0), col_ptr)
-                    end
+    function do_work(task_n, part)
```
ToucheSir, line 127:
```suggestion
    function ∇depthwiseconv_data_part(task_n, part)
```

src/NNlib.jl
IanButterworth: That could make sense. I'll comment that this is internal, so it can be changed when that design is figured out.

src/NNlib.jl
IanButterworth: Good point. I move to use `@with`

src/NNlib.jl
ToucheSir: Yes, I should've elaborated on this thought. The idea would be to have a way to control the degree of parallelism via a context manager/macro like `disallow_spawns`, but using a numeric param instead of an on/off flag.

But this design can be fleshed out later. Then there are more QoL things (not for this PR) we can consider around threading that would complement it. For example, deciding the number of tasks to spawn based on # of threads / # of BLAS threads.

IanButterworth:
Hard for me to tell if any of these failures are related?

ToucheSir (approved):
No, I just hadn't the time to make sure CI was all good and to do another read-through.

Can you bump the version, and then we'll merge + release?

PR #613, opened 2024-11-09
Bump deps

pxl-th:
- Use `1.10` in Buildkite CI (and `nightly` for CUDA).
- Use `lts`, `1` and `pre` in GitHub CI.
- Add compat GPUArraysCore for `0.2`.
- Bump ChainRulesCore to `0.25`.
- [x] Need to wait for AMDGPU@1.1 release: https://github.com/JuliaGPU/AMDGPU.jl/commit/4385ed941d5bf1b4818d8cd34ca090b8827d1ca4

### PR Checklist

- [ ] Tests are added
- [ ] Documentation, if applicable

PR #636, opened 2025-03-13
Use the correct `nthreads`

### LINKED ISSUE #634: Wrong number of threads used if interactive pool exists.
My fix in https://github.com/FluxML/NNlib.jl/pull/496 introduced a bug where the number of usable threads is overreported by the size of the interactive pool.

NNlib only uses `@spawn ...` so uses the `:default` pool.

Using `maxthreadid` prevents breaking indexing that uses `threadid` but means too many threads are expected to be available.

Julia 1.12 is being trialed with an interactive thread by default, so it will always be +1 too high on 1.12.

To fix this, I think either:
1) the threadid-indexing needs to be removed
2) go back to `nthreads(:default)` and make sure threadid-based indexing offsets by the size of the interactive pool

IanButterworth:
Fixes #634 

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/dim_helpers/ConvDims.jl
+++ b/src/dim_helpers/ConvDims.jl
@@ -73,7 +73,7 @@ function im2col_dims(c::ConvDims)
         # Size of single dotproduct within convolution
         prod(kernel_size(c))*channels_in(c),
         # One workspace per thread
-        VERSION > v"1.9.0-0" ? Threads.maxthreadid() : Threads.nthreads(),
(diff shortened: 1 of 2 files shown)


IanButterworth:
These changes aren't covered by tests it seems.

ToucheSir:
You know why, it's because CI was bumped to 1.10 in https://github.com/FluxML/NNlib.jl/pull/613 but compat was not. I'd be ok with doing that now and getting rid of the VERSION ternary check.

CarloLucibello:
I'm going to merge this and bump to 1.10 in a follow up PR

PR #639, opened 2025-06-06
Fixed the lpnormpool implementation

IBArbitrary:
The lpnormpool implementation was wrong. The existing implementation is `y = (∑ᵢ xᵢ^p)^(1 / p)`, whereas the true Lp norm shuld be `y = (∑ᵢ |xᵢ|^p)^(1 / p)`, where `|x|` is `abs(x)`. This also affects the gradient, which should be `∂y/∂xᵢ = |xᵢ|^(p-1) × y^(1-p) × sign(xᵢ)` instead of `∂y/∂xᵢ = xᵢ^(p-1) × y^(1-p)`. The necessary changes should go into `src/impl/pooling_direct.jl` which is what this PR achieves.

The existing implementation is exact only for the input array `x` such that `xᵢ >= 0 ∀ xᵢ ∈ x`. This will not generally be the case. For example, when the input to the pooling layer comes from a layer which had leaky ReLU activation, then the problem of the implementation would manifest.

MWE of the problem:
```julia
x = randn(4, 4, 4, 4)
lpnormpool(x, 0.5, (2, 2))
> ERROR: DomainError with -0.21106262688542637:
> Exponentiation yielding a complex result requires a complex argument.
> Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
```

### DIFF
--- a/src/impl/pooling_direct.jl
+++ b/src/impl/pooling_direct.jl
@@ -95,14 +95,14 @@ for name in (:max, :mean, :lpnorm)
                 elseif $(name == :mean)
                     m += x[input_kw, input_kh, input_kd, c, batch_idx]
                 elseif $(name == :lpnorm)
-                    # y = (∑ᵢ xᵢ^p)^(1 / p), here to calculate ∑ᵢ xᵢ^p
+                    # y = (∑ᵢ |xᵢ|^p)^(1 / p), here to calculate ∑ᵢ |xᵢ|^p
                     m += abs(x[input_kw, input_kh, input_kd, c, batch_idx])^p
                 else
                     error("Unimplemented codegen path")
                 end
             end
 
-            # for lpnormpool, y = (∑ᵢ xᵢ^p)^(1 / p)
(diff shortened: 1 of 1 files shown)


IBArbitrary:
> PR needs tests. At present only `p=1,2` are tested, perhaps that's all that this was intended to support?

Forgive me if I'm mistaken, I am not familiar with tests. I found [this line](https://github.com/FluxML/NNlib.jl/blob/1468582c4db5f18149cc8fff6fb4633c5debe5c5/test/pooling.jl#L268C8-L268C30) in `/test/pooling.jl`. The `p` values being tested are half integer. It works there because the arrays being pooled are also positive. But that is not the general case is what I feel. For integer values of `p`, the pooling would work for non positive arrays as well, but why I assumed its supposed to support any values is because `p` was typecasted as `Real`.

I belive my implementation is faithful to the definition of Lp norm. If you can give me resources or tips to construct tests, I'll add that to my fork.

PR #614, opened 2024-11-15
add Metal extension for batched_mul

mcabbott:
Closes #581

### PR Checklist

- [x] Tests are added
- [ ] Documentation, if applicable

PR #644, opened 2025-08-17
Add `batched_mul` support for oneAPI

AntonOresten:
Like #614, this PR adds partial support for a previously unsupported backend. It was just a matter of hooking up the bindings, using that PR as template. I think `NNlib.batched_mul` is particularly useful, and it was one of my only hiccups with trying oneAPI.jl, as this function is used extensively in the ecosystem, so even partial oneAPI support would be useful.

image

### PR Checklist

- [x] Tests are added
- [x] Documentation note about partial oneAPI support


CarloLucibello:
This looks good. Just add the buildkite pipeline for testing. It should have the option:
```
        agents:
          queue: "juliagpu"
          intel: "*"
```

PR #648, opened 2025-11-14
Extending scatter! to work with CUDA sparse arrays

### LINKED ISSUE #647: Specialized scatter! dispatch for CUSPARSE.CuSparseMatrix
### Motivation and description

While trying to call `scatter(+, dst, src, idxs)` where `src` is a CUDA sparse matrix (`CUDA.CUSPARSE.CuSparseMatrixCSC` to be precise) I noticed that the GPU kernel isn't called. This is because `CuSparseMatrixCSC` is not a subtype of `CUDA.AnyCuArray` (not sure if I should report this on CUDA.jl or if it's intended behaviour)

Would it make sense to add a special dispatch for sparse arguments? It could simply call the existing kernels, but that way it wouldn't end up calling the CPU version. Perhaps even rely on cuDNN.

 Alternatively, would it perhaps make sense to make an upstream issue so that `CuSparseMatrixCSC <: AnyCuArray`?

### Possible Implementation

add a new method
```julia

function NNlib.scatter!(op::OP, dst::Union{CuSparseMatrixCSC, AnyCuArray}, src::...)
    # Possibly call cuDNN or some other CUDA library that implements an optimized sparse version
    # or simply call the existing `scatter_kernel!`
```

alonsoC1s:
Aims to fix #647 by extending the signature of `scatter!` to work with `AbstractCuSparseArray`, a CUDA array type notably excluded by the original method. With the proposed patch, calling `scatter!` with sparse arrays from `CUDA.CUSPARSE` will correctly call the CUDA-specialized method instead of calling the generic CPU method, which triggered a scalar indexing error. In my testing the existing CUDA kernels work perfectly fine with `CuSparseArrayCSC`.

The proposed implementation, perhaps inelegantly, just expands the types in the signature with `Union{...}`. I am open to discussing more beautiful ways of implementing this. Ideally, `AbstractCuSparseArray` would be a subtype of `AnyCuArray`.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/ext/NNlibCUDAExt/scatter.jl
+++ b/ext/NNlibCUDAExt/scatter.jl
@@ -1,4 +1,5 @@
 # supported op: +, -, *, /, max, min, &, |, mean
+import CUDA: CUDA.CUSPARSE.AbstractcuSparseArray
 
 function scatter_kernel!(op::OP, dst, src, idx) where OP
     index = threadIdx().x + (blockIdx().x - 1) * blockDim().x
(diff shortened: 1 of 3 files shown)


alonsoC1s:
The integration test with Lux fail because `always_inliner!` is not defined, and what look like Enzyme internal errors. Not sure if this is unrelated

alonsoC1s:
@CarloLucibello ~I'll take a look at the failing tests. I hope I can get easier access to a GPU to test  it on this week~

Turns out I don't need a GPU to test. I'm working on the fix

PR #672, opened 2026-01-27
 Extending scatter! to work with CUDA sparse arrays (retry of #648)

alonsoC1s:
This is a new attempt at PR #648 

It extends the signature of scatter! to work with AbstractCuSparseArray, a CUDA array type notably excluded by the original method. With the proposed patch, calling scatter! with sparse arrays from CUDA.CUSPARSE will correctly call the CUDA-specialized method instead of calling the generic CPU method, which triggered a scalar indexing error.


alonsoC1s:
Quick update. The test failures were significant, and I underestimated what needed to be done. I'll mark the PR as draft while I work on https://github.com/JuliaGPU/GPUArrays.jl/issues/694, which will hopefully make finishing this PR for all of the different sparse formats more feasible

alonsoC1s:
The implementation of `scatter!` and friends should now work with GPU sparse device arrays as a source, though not as a destination, since that would potentially require allocating in device code. I recreated the standard testing suite and specialized it so that it makes sense in the sparse case

This PR is heavily dependant on [a related PR to GPUArrays.jl](https://github.com/JuliaGPU/GPUArrays.jl/pull/699) being merged, and shouldn't make its way into main until the other one is succesfully merged and a new release is made

ext/NNlibCUDAExt/scatter.jl
```diff
@@ -79,8 +78,8 @@ function NNlib.scatter!(op::OP, dst::AnyCuArray,
 end
 
 function NNlib.scatter!(op::typeof(mean), dst::AnyCuArray,
-        src::AnyCuArray,
-        idx::AnyCuArray)
+    src::Union{AnyCuArray,AbstractGPUSparseArray},
```
CarloLucibello: if possible, the type should be restricted to CUDA sparse arrays, we don't want generic gpu sparse arrays

alonsoC1s:
Looking at the CI failures, I can't tell if they are related. They seem to happen only when doing gradients with Enzyme, and some don't even fail for `scatter!`, but other functions. While testing locally, it all worked as expected

Issue #429, opened 2022-07-25
`test_rrule` for `gather!, scatter!`


vpuri3:
https://github.com/FluxML/NNlib.jl/blob/023cd3da63892b754b4197fe7a848093128f2bf9/src/scatter.jl#L206-L212

https://github.com/FluxML/NNlib.jl/blob/023cd3da63892b754b4197fe7a848093128f2bf9/src/gather.jl#L82-L87

CarloLucibello:
we already test gradients with enzyme and zygote

Issue #416, opened 2022-06-02
add bounds checking in `gather`


MilkshakeForReal:
Currently `NNlibCUDA.gather!` would not check bounds 
```julia
using NNlib, CUDA
using NNlibCUDA

src = CUDA.rand(2,3)

NNlib.gather(src,cu[1,4])

2×2 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 0.430532  0.0
 0.474528  0.0
```

CarloLucibello:
closed by #705

PR #689, opened 2026-06-08
gather: densify sparse sources, GPU bounds checking, rrule tests

### LINKED ISSUE #625: `gather` on a sparse source should return a dense array
This is the current behavior:
```julia
julia> s = [1, 2, 3];

julia> t = [2, 3, 1];

julia> A = sparse(s, t, [1.0, 1.0, 1.0], 3, 3)
3×3 SparseMatrixCSC{Float64, Int64} with 3 stored entries:
  ⋅   1.0   ⋅ 
  ⋅    ⋅   1.0
 1.0   ⋅    ⋅ 

julia> w = NNlib.gather(A, s, t)
3-element SparseVector{Float64, Int64} with 3 stored entries:
  [1]  =  1.0
  [2]  =  1.0
  [3]  =  1.0
```
I would expect `w` to be a dense vector instead.

CarloLucibello:
Closes #625, #416, #429.

## #625 — `gather` rejects a sparse source
Previously `gather` allocated its destination with `similar(src, ...)`, which yields a *sparse* array when `src` is sparse. Matching PyTorch (`torch.gather`) and PyTorch Geometric — where the gather source is always dense — `gather` on an `AbstractSparseArray` now throws an informative `ArgumentError` (via a lightweight `SparseArrays` extension) rather than silently returning a sparse result:

```julia
A = sparse([1,2,3],[2,3,1],[1.0,1.0,1.0],3,3)
NNlib.gather(A, [1,3,1])
# ERROR: ArgumentError: `gather` does not support sparse sources; convert
# `src` to a dense array first, e.g. `gather(collect(src), idx)`.
```

## #416 — bounds checking in GPU `gather`
The CPU `gather!` already throws a `BoundsError` (via `view`), but the GPU kernel uses `@inbounds` and would silently read out-of-range memory. Added `checkbounds_gather`, called from the `AnyGPUArray` `gather!`, which validates all indices with a single `mapreduce` (handling scalar, tuple, and `CartesianIndex` indices) and throws `BoundsError` on out-of-range access.

> Note: a previously-silent out-of-bounds gather on GPU now errors.

## #429 — `test_rrule` for `gather!`/`scatter!`
Added CPU-only `test_rrule` testsets covering `gather!` (scalar and tuple indices) and `scatter!`/`scatter` (`+` and `-`).

## Testing
New CPU tests pass locally. The GPU bounds-check path can't be exercised here (no GPU), but `checkbounds_gather` is backend-agnostic and tested on CPU arrays.


CarloLucibello:
`dense_like` could go to MLUtils.jl

CarloLucibello:
tries to do too many things, closing in favor of separate PRs

PR #709, opened 2026-06-10
skip enzyme tests on amdgpu

CarloLucibello:
follow up to #708,  skipping enzyme tests on AMDGPU since they fail (the `gather` testset) or crash julia entirely (unknown tests)


CarloLucibello:
AMDGPU tests are crashing now, probably another enzyme bug. We should skip the whole combination julia 1.12 + enzyme + amdgpu

PR #718, opened 2026-06-10
Refactor softmax and hook cudnn backward

CarloLucibello:
This PR does the following regarding `softmax(x; dims)` and `logsoftmax(x; dims)`:
- gets rid of deprecated methods and awkward reindirections
- clarifies what NNlib exposes, adding it to the documentation
- hooks the cuDNN backward, which was greatly sped up in #717, to the rrule, so making the gain real
- tests second derivatives on CUDA

## Benchmark — Zygote gradient path on GPU

Times the end-to-end Zygote AD path (`benchmark/cuda_softmax/bench_softmax.jl`): forward + the `rrule` pullback, not the `∇softmax` kernels directly. **Before** = a few commits back, pre-#717 (CHANNEL forward; broadcast AD backward). **After** = this PR. So the comparison captures the combined effect of #717 and hooking the cuDNN backward into the rrule.

TITAN RTX, cuDNN 9.2, Float32, times in µs (lower is better). `bwd-*` = Zygote pullback only.

### `dims=1` (leading contiguous axis → cuDNN INSTANCE applies)

The Zygote backward now hits cuDNN INSTANCE instead of the broadcast — **1.5–3.4× faster**:

| size | bwd-sm before | bwd-sm after | speedup | bwd-ls before | bwd-ls after | speedup |
|---|---:|---:|---:|---:|---:|---:|
| (256, 10, 32) | 35.6 | 12.5 | 2.8× | 27.2 | 11.4 | 2.4× |
| (256, 1000)   | 40.4 | 12.0 | 3.4× | 27.1 | 12.4 | 2.2× |
| (1000, 1000)  | 101.5 | 33.3 | 3.0× | 71.4 | 34.8 | 2.1× |
| (100, 10000)  | 103.0 | 30.2 | 3.4× | 71.5 | 30.2 | 2.4× |
| (10000, 100)  | 81.9 | 49.1 | 1.7× | 52.2 | 40.2 | 1.3× |
| (32000, 64)   | 141.6 | 91.9 | 1.5× | 92.5 | 80.2 | 1.2× |
| (1000, 128)   | 33.5 | 11.5 | 2.9× | 26.6 | 11.9 | 2.2× |

Forward `dims=1` also improves (CHANNEL→INSTANCE), most on a short axis: `(32000, 64)` 104.0 → 40.1 µs, `(10000, 100)` 39.1 → 22.0 µs.

### `dims=2` (strided axis → INSTANCE not applicable)

Backward is **unchanged** — broadcast both before (`∇softmax_data`) and after (cuDNN can't help a strided axis):

| size | bwd-sm before | bwd-sm after | bwd-ls before | bwd-ls after |
|---|---:|---:|---:|---:|
| (256, 10, 32) | 42.9 | 40.7 | 30.5 | 31.1 |
| (256, 1000)   | 43.5 | 41.1 | 30.8 | 33.3 |
| (1000, 1000)  | 110.5 | 105.7 | 79.9 | 79.9 |
| (100, 10000)  | 90.7 | 86.1 | 59.1 | 59.7 |
| (10000, 100)  | 103.4 | 100.6 | 74.2 | 74.2 |
| (32000, 64)   | 179.5 | 173.9 | 125.6 | 125.4 |
| (1000, 128)   | 35.9 | 32.6 | 26.8 | 24.0 |

Note: forward `dims=2` could be regressing on a short aggregation axes, although is not seen here since the backward dominates. A cure would be to use in the forward CUDNN_SOFTMAX_MODE_CHANNEL below a certain axis size (e.g. 128) and our implementation above, as was done on master before #717. I think the extra complexity and the arbitrary heuristic is not worth it though.


wsmoses:
@CarloLucibello would it make sense to also attach Enzyme rules to these?

PR #727, opened 2026-06-24
Mooncake rule for `NNlib.batchnorm`.

AstitvaAggarwal:
`NNlib.batchnorm` on `CuArrays` calls `cudnnBNForward!`, which contains a `try/catch` block. `Mooncake` cannot trace through `try/catch` in reverse mode, so AD of `batchnorm` on `GPU` fails without an explicit rule.

The PR adds `ext/NNlibMooncakeCUDAExt.jl`, loaded when both `Mooncake` and `CUDA` are available, which provides an `rrule!!` intercepting `Core.kwcall(..., batchnorm, g, b, x::CuArray, ...)`. The primal runs via `cuDNN` as normal and the pullback uses `NNlib.∇batchnorm`, which is safe to call directly from a pullback.

**Note :** 
The pullback passes a `_BNFwdCache` to both the forward and backward calls. Without it, `cudnnBNBackward!` recomputes batch mean and inverse-variance from `x` via a non-deterministic GPU parallel reduction, causing Mooncake's Reuse test (which checks exact equality across two pullback calls) to fail. The cache causes `cuDNN` to save these values during the forward and reuse them in the backward, making the pullback deterministic.

### PR Checklist

- [x] Tests are added


CarloLucibello:
buildkite CUDA tests are not passing

AstitvaAggarwal:
@CarloLucibello I think this is ready now. Please let me know if anything more is needed here, thanks!

AstitvaAggarwal:
Thank you @CarloLucibello !

CarloLucibello:
Thank you!

PR #750, opened 2026-08-03
Gradient operators for normalization ops (second-order differentiation)

CarloLucibello:
Follow-up to #748. Adds explicit VJP operators for each normalization forward and wires the forward `rrule`s through them, enabling efficient first-order **and** correct second-order differentiation.

## New operators

```julia
∇batchnorm(g, b, x, dy, running_mean=nothing, running_var=nothing, momentum=0.1f0; eps=1f-5, training=true)
∇instancenorm(g, b, x, dy, running_mean=nothing, running_var=nothing, momentum=0.1f0; eps=1f-5, training=true)
∇groupnorm(g, b, x, dy, G::Integer; eps=1f-5)
∇layernorm(g, b, x, dy; dims=1, eps=1f-5)
```

Each returns `(dg, db, dx)` (`dg`/`db` are `nothing` when the matching parameter is).

- **`∇batchnorm` now has a generic fallback** — previously it existed only as a cuDNN method, so a CPU caller hit a `MethodError`. The other three are new.
- Computed analytically from `mean`/`var`/broadcast (standard normalization backward); validated against finite differences (via `test_gradients`) and Zygote-through-body.

## Second-order

The forward `rrule`s route through these operators. Because the operators are themselves built from differentiable primitives, the pullbacks are differentiable too, so nested AD yields correct second-order derivatives. Verified Hessian-vector products (Zygote-over-Zygote) against ForwardDiff-over-ForwardDiff to machine precision for all four ops.

The rrules carry the same positional defaults as the forwards, so the common 3-arg call form (`batchnorm(g, b, x)`) also uses the analytic path.

## Dispatch / correctness notes

- On the GPU the cuDNN `batchnorm` rrule (typed on `DenseCuArray`, 6-positional) stays more specific and wins for the 2D/4D/5D Flux path; 3-arg or non-cuDNN-shape GPU calls use the generic rrule + generic `∇`. The generic `∇` operators run on-device with no scalar indexing (checked in the CUDA suite).
- Half-precision inputs accumulate the gradient statistics in Float32 and cast `dx` back, matching the forward.
- Second-order is **not** nested on the GPU cuDNN batchnorm path (its backward is a non-differentiable kernel); the second-order tests are CPU-gated.

## Tests

- `gradient operators (∇)` — runs on CPU + every GPU backend, exercising the generic fallbacks on-device.
- `second order` (CPU) — Hessian-vector products vs ForwardDiff.

All CPU and CUDA normalization suites pass locally.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -173,6 +173,10 @@ NNlib.batchnorm
 NNlib.instancenorm
 NNlib.groupnorm
 NNlib.layernorm
+NNlib.∇batchnorm
+NNlib.∇instancenorm
+NNlib.∇groupnorm
+NNlib.∇layernorm
 ```
 
 ## Losses
--- a/src/NNlib.jl
+++ b/src/NNlib.jl
@@ -127,7 +127,8 @@ export stft, istft, hann_window, hamming_window, spectrogram, melscale_filterban
 # `public` is only valid syntax on Julia 1.11+, so parse it lazily.
 @static if VERSION >= v"1.11.0-DEV.469"
     eval(Meta.parse("public gather, gather!, scatter, scatter!, fold, unfold, glu, within_gradient, " *
-                    "normalise, batchnorm, instancenorm, groupnorm, layernorm"))
+                    "normalise, batchnorm, instancenorm, groupnorm, layernorm, " *
+                    "∇batchnorm, ∇instancenorm, ∇groupnorm, ∇layernorm"))
 end
 
 end # module NNlib
--- a/src/normalization.jl
+++ b/src/normalization.jl
@@ -241,9 +241,175 @@ function layernorm(g, b, x::AbstractArray{T}; dims=1, eps=1f-5) where {T}
     return Tc === T ? y : T.(y)
 end
 
-# --- Low-level cuDNN batchnorm bridge ----------------------------------------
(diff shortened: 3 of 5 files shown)


CarloLucibello:
Also folds in a fix for the Buildkite CUDA failure from #748.

**Cause:** the functional `batchnorm` from #748 added a generic `batchnorm(g, b, x::AbstractArray, …)` method that — unlike the cuDNN methods, which only accept `(array, array)` or `(nothing, nothing)` — also matched a *mixed* `(array, nothing)` and silently defaulted the missing parameter. `test/gpu/cuda/batchnorm.jl` asserts mixed affine params are rejected, so it began failing on the GPU runner (the new generic `∇batchnorm` in this PR would break it the same way).

**Fix:** enforce the both-or-neither contract explicitly via `_check_affine` in every forward and gradient operator, throwing a descriptive `ArgumentError` (matching the existing mixed running-statistics check). The GPU test now expects `ArgumentError` instead of the previously-incidental `MethodError`, and the shared suite gained device-agnostic coverage.

Verified locally: `test/gpu/cuda/batchnorm.jl` 46/46, plus the CPU and CUDA normalization suites.

(The separate "Julia nightly" red on #748 is unrelated — an upstream Enzyme/GPUCompiler precompile break on Julia 1.14-dev: `UndefVarError: WorldView not defined in Compiler`.)

